PR #2, opened 2015-08-23
Use `State::default()` directly, instead of per-member.

Byron:
The example is sightly simplified that way, even though showing the
`..` notation for struct field initialization is probably a good thing
to do as well just for the learning experience.

### DIFF
--- a/examples/test_window.rs
+++ b/examples/test_window.rs
@@ -87,9 +87,7 @@ impl Default for AutoResizeState {
 }
 
 fn main() {
-    let mut state = State {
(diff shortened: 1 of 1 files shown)


Byron:
You can watch the development stream [on youtube](https://youtu.be/n8idYWCwl5s).

*`Looksies #10 [Let's look at the imgui-rs]`*

[\thumb](https://youtu.be/n8idYWCwl5s)

Gekkio:
Thanks, looks good! The .. was an unnecessary leftover of older code.

PR #9, opened 2015-10-09
Support for Text Input

Twinklebear:
I've got rough support for text input working but at the moment we do a kind of nasty cast of the char to a u16. We really should be using [encode_utf8](http://doc.rust-lang.org/std/primitive.char.html#method.encode_utf8) or whatever stabilizes in its place to send ImGUI an array of utf8 characters.

Along with this I've been working on keyboard support in the examples but seem to have some issues with glutin not giving me virtual keycodes for keys like ctrl and shift and such. I'm also looking for a way to get the scancode for a virtual key so we can tell ImGUI where the arrows, A, Z, tab, ctrl and so on are located. If you've had good luck with SDL2 I may just use this in my project but it's still needed for the demo to work cleanly. I'll have to ask in the IRC about this.

Let me know what you think!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ default = ["glium"]
 libc = "0.1"
 
 [dependencies.glium]
-version = "0.9.3"
+version = "0.10"
 default-features = false
 optional = true
 
@@ -30,6 +30,6 @@ gcc = "0.3"
 time = "0.1"
 
 [dev-dependencies.glium]
-version = "0.9.3"
+version = "0.10"
 features = ["glutin"]
 default-features = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ default = ["glium"]
 libc = "0.1"
 
 [dependencies.glium]
-version = "0.9"
+version = "0.9.3"
 default-features = false
 optional = true
 
@@ -30,6 +30,6 @@ gcc = "0.3"
 time = "0.1"
 
 [dev-dependencies.glium]
-version = "0.9"
+version = "0.9.3"
 features = ["glutin"]
 default-features = false
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -16,7 +16,8 @@ fn main() {
     let mut support = Support::init();
 
     loop {
-        let active = support.render(CLEAR_COLOR, hello_world);
+        support.render(CLEAR_COLOR, hello_world);
+        let active = support.update_events();
         if !active { break }
     }
 }
--- a/examples/support/mod.rs
+++ b/examples/support/mod.rs
@@ -25,26 +25,25 @@ impl Support {
         let mut imgui = ImGui::init();
         let renderer = Renderer::init(&mut imgui, &display).unwrap();
 
-        // TODO: How can we get the virtual key -> scancode mapping from glium?
-        imgui.set_imgui_key(ImGuiKey::Tab, 15);
-        imgui.set_imgui_key(ImGuiKey::LeftArrow, 75);
-        imgui.set_imgui_key(ImGuiKey::RightArrow, 77);
-        imgui.set_imgui_key(ImGuiKey::UpArrow, 72);
-        imgui.set_imgui_key(ImGuiKey::DownArrow, 80);
-        imgui.set_imgui_key(ImGuiKey::PageUp, 73);
-        imgui.set_imgui_key(ImGuiKey::PageDown, 81);
-        imgui.set_imgui_key(ImGuiKey::Home, 71);
-        imgui.set_imgui_key(ImGuiKey::End, 79);
-        imgui.set_imgui_key(ImGuiKey::Delete, 83);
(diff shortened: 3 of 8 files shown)


Gekkio:
Sorry, I totally forgot about this. Looks promising!

Some random comments:

* The version code issue is a stupid mistake by me, and I'll fix it in master. It's okay to declare 0.0.4-pre versions and depend on 0.0.4, but I missed one place which is causing the problem.
* Changes in imgui-sys look good. I'm also considering exposing io.keys_down in the API instead of just having a setter method for a single key. I don't know which one is better
* Good points about semantics of add_input_character. I'll need to take a closer look at how imgui works internally to figure out if there is another way of doing it without using nightly APIs. I recall seeing something in the implementation that was a bit suspicious if all valid UTF-8 is supposed to be accepted.

Now, about the scancode/virtual key mapping in the examples. Couldn't we just use the virtual keys? If I understand correctly, when we get a key event, in practice we should get a virtual key for all the important keys.

If we want to support raw scancodes as a fallback, we can also exploit the fact that keys_down is an array of 512 elements. There's less than 256 virtual keys and scancodes are u8, so we can map both into a 9-bit value space using one bit as a discriminator. For example, the hex value `0x1F` could refer to virtual key number 31, while the hex value `0x11F` could refer to scancode 31. An event with both virtual key code and scancode would be mapped by the virtual key, while an event with just a scancode would use the scancode.

Twinklebear:
- Exposing io.keys_down might be better, it would reduce the function calls to set_key, but the user would be required to do the cast from their scancode value to a usize. So I'm not sure which would be best either.
- For the Unicode we might be able to use [EscapeDefault](http://doc.rust-lang.org/std/char/struct.EscapeDefault.html) which looks like the current stable version of encode_utf8/16. It sounds like maybe a performance thing to expose encode_utf8/16 but we'll be dealing with very few characters at a time so it may not be an issue. Switching when/if it stabilizes might be good though.

Just using the virtual keys should work, we could set the keys ImGui needs to be scan codes 0-19 or something and then use the virtual key to set the appropriate 'scan code'. This would be better actually since scan codes will be different on different machines, I'm not sure if glutin has an API to get the mapping. I think it's reasonable to expect getting a virtual key code for all important keys but I've noticed on my laptop and desktop (Win10) I don't get ones for control and shift, though on my linux machine I do get these virtual keys. I'll have to take a look at glutin since this seems like an issue related to that.

Twinklebear:
After some more digging into why we aren't getting the virtual keycodes on Win10 it looks to be glium/glutin's fault. The issue is fixed in glutin 0.4 and it looks like there's a PR to update glium's dependencies tomaka/glium#1304. Testing the glium triangle example on master (glutin 0.3) I don't get virtual keycodes for ctrl/shift/alt on Windows but on the update-deps branch (glutin 0.4) I do. So it looks like this should be fixed by just waiting a bit :smile:

Twinklebear:
glium 0.10 was released with the update to glutin 0.4 so ctrl/shift/alt/etc now get the right virtual keycodes on Win10. I've updated the glium dependencies which required some minor changes in glium_renderer.rs and it looks like things are now working on Win10 as well.

I also switched to using the encode_default API to encode the char to a String which we then convert to raw bytes and to a c-string to send to ImGui. This is definitely not as nice as the encode_utf8 API which would let us go straight to a Vec<u8> but should work ok for now on stable Rust. I'll track the issue for merging in encode_utf8 and update when it stabilizes (hopefully it does).

Gekkio:
Looks good!
There's some minor things that could use a bit more refining (panics are unnecessary, and reading keys_down is not considered in the API), but I'll gladly merge this now.

Could you please rebase and remove the changes that touch imgui-rs version numbers? I fixed the build in the master branch, so those changes should not be necessary and the -pre versions should remain. Thanks!

Twinklebear:
Sounds good! I wasn't sure how you'd want to change the handling of shaders that fail to compile, if you have something in mind I could add it or just merge and you can change it later.

Gekkio:
We should be able to wrap and propagate all errors, but you don't need to worry about that now unless you want to. I haven't yet looked at Glium 0.10 API changes so I don't know exactly what is needed, but the basic idea is to always use try! and implement From to convert 3rd party errors to a common library error type.

Twinklebear:
Ah I see, yes when I initially ported it over I was running into some compile errors about From not supporting the new error type. Glium changed the error type returned by the `program!` macro to [ProgramChooserCreationError](http://tomaka.github.io/glium/glium/program/enum.ProgramChooserCreationError.html) so switching to handle that should let us go back to using From.

Gekkio:
Alright, that makes sense. It looks like ProgramChooserCreationError wraps ProgramCreationError, so I think we probably can just replace ProgramCreationError everywhere with ProgramChooserCreationError (so basically we should change: Program variant in RendererError, From implementation, return type of compile_default_program).

Twinklebear:
Cool, I've added a commit to switch to wrapping ProgramChooserCreationError instead.

Gekkio:
Looks good! I'll merge this now.

PR #10, opened 2015-10-17
add function to get framerate

matklad:
As far as I undestand, there is no way to get framerate information, provided by imgui yet. This PR adds such ability. 

There is no `igGetFrameRate` function in C-api, so the direct acces to `io`'s field is used.


Gekkio:
Looks good. Thanks!

PR #15, opened 2016-01-16
Exposed more input functionality

bitshifter:
Hi, I've been using your wrapper for a little project I'm working on and adding missing functionality as I go. I've exposed float and int array inputs and color inputs so far. This is still a work in progress, I'm continuing to add stuff as I need it.

I've added macros to reduce some of the code duplication, but due to small differences between all of these functions there's still a bit of duplication.

At the moment the demo code I've added reuses the same members for different inputs, I'll probably change this at some point so there's a dedicated struct for each section of the demo.

Naming wise I named functions like `input_float2`, `input_int3` before I noticed the single value versions were `input_f32`, `input_i32` etc. I've stuck with the original imgui names though, because `input_f32_2` looks weird and is painful to type, also I figure staying closer to the original API is probably better where possible even though it's less Rusty.

In doing all this I have some observations about the API:
 * Having to `im_str!` everything is a pain, I guess it's unavoidable, that `_` just makes it twice as annoying to type though.
* Because of the builder pattern being used to handle optional parameters, there is a lot of boiler plate required to expose new functions. This doesn't affect people using the library so much, but I do feel having to call `build()` after some functions is a bit unintuitive and could be easy to forget.

On the latter point I'm temped to experiment with changing the API to have a default function call (e.g. `input_float`) and a function call which can take a parameters struct (e.g. `input_float_with` or `input_float_ex`), the struct would be similar to the existing builder structs. The advantage of this is it's less boilerplate to add, you don't have to call build and you can reuse the same parameters between different calls (often you want the same parameters anyway). Of course all of this will be unnecessary if Rust gets default parameters one day.


Gekkio:
Thanks! I'll take a look during the weekend.

>  that _ just makes it twice as annoying to type though. 

We could rename it to `imstr!`. It's not consistent naming, but could be a reasonable compromise.

> having to call build() after some functions is a bit unintuitive and could be easy to forget

`#[must_use]` guarantees a compiler warning if that happens. When I added it, I found out I had forgotten to call `build()` sometimes in the `test_window_impl` example ;)

bitshifter:
> #[must_use] guarantees a compiler warning if that happens. When I added it, I found out I had forgotten to call build() sometimes in the test_window_impl example ;)

Ah, I was wondering what that did :)

bjz:
> Having to im_str! everything is a pain

My PR #19 changes everything to use `Into` :)

bitshifter:
@Gekkio any feedback on these changes?

bitshifter:
Removed implementation of combo as master got it's own implementation.

imgui-sys/src/lib.rs
```diff
@@ -785,7 +785,7 @@ extern "C" {
     pub fn igRadioButton(label: *const c_char, v: *mut c_int,
                          v_button: c_int) -> bool;
     pub fn igCombo(label: *const c_char, current_item: *mut c_int,
-                   items: *mut *const c_char, items_count: c_int, height_in_items: c_int) -> bool;
+                   items: *const *const c_char, items_count: c_int, height_in_items: c_int) -> bool;
```
Gekkio: I'm a bit pedantic here, but this change is not correct.
A `const char** items` has only one level of const. It is true that ImGui will not modify items so a `*mut` pointer is unnecessary, but I'd prefer wrapping stuff exactly as-is to match the level of const correctness in the original library

imgui-sys/src/lib.rs
```diff
@@ -794,8 +794,8 @@ extern "C" {
                     data: *mut c_void, items_count: c_int,
                     height_in_items: c_int) -> bool;
     pub fn igColorButton(col: ImVec4, small_height: bool, outline_border: bool) -> bool;
-    pub fn igColorEdit3(label: *const c_char, col: [c_float; 3]) -> bool;
-    pub fn igColorEdit4(label: *const c_char, col: [c_float; 4], show_alpha: bool) -> bool;
+    pub fn igColorEdit3(label: *const c_char, col: *mut f32) -> bool;
```
Gekkio: Wow, nice catch! :scream:
I don't know what I was thinking when I wrote this.

src/lib.rs
```diff
@@ -507,6 +550,13 @@ impl<'ui> Ui<'ui> {
     }
 }
 
+// Widgets: Selectable / Lists
+impl<'ui> Ui<'ui> {
+    pub fn selectable<'p>(&self, label: ImStr<'p>) -> bool {
```
Gekkio: This seems to be work in progress? Most of the arguments are hard-coded to be defaults...

Gekkio:
Sorry, I totally forgot about this pull request!
There's still stuff missing, so I thought you were still working on it.

I feel a bit conflicted about the macros...It does reduce code duplication, but I can't shake the feeling that there could be a better way to do that with traits and type parameters. However, I can't yet see in my mind how the API would look like, so let's keep the macros for now.

src/lib.rs
bitshifter: I think I didn't want to use the parameters at the time so I didn't expose them. I've added them now. In C++ ImGui these have default values, it will be nice when Rust adds support for default parameters!

bitshifter:
> There's still stuff missing, so I thought you were still working on it.

Ah right, I just added some things I needed at the time. Sorry I wasn't clearer.

> I feel a bit conflicted about the macros...It does reduce code duplication, but I can't shake the feeling that there could be a better way to do that with traits and type parameters. However, I can't yet see in my mind how the API would look like, so let's keep the macros for now.

Possibly the flags could be done with generic traits rather than macros, although that would make the trait part of the interface which I think is a bit confusing since really it's an implementation detail. I don't think the impl_input_intn and impl_input_floatn macros could be done with generics in Rust.

In any case, I'll leave the macros as is for now.

Gekkio:
Alrighty! Just one last thing: I think we should remove the input_f32 and input_i32 functions, because they are just duplicates. I agree with your reasoning about f32_4 vs float4, and I'd much rather change the existing ones (e.g. slider_i32) to match these new ones.

bitshifter:
Done. It's unfortunate that f32_4 etc. are not nice to type, otherwise it would have been good to stick with the Rust type names.

Gekkio:
Thanks! Merging now

PR #16, opened 2016-01-17
update cimgui to 1.47

not-fl3:
Updated cimgui especially for password edit fields.


Gekkio:
Thanks! I'll implement the rest of 1.47 changes soon

Issue #7, opened 2015-08-25
API design: string arguments (static)


Gekkio:
One design decision that is often discussed is how string arguments are handled. This issue is for discussing static string arguments (= string literals). Also, the discussion considers safe Rust code. As usual, in unsafe code, all bets are off.

First, let's look at the facts:

* ImGui expects UTF-8 strings with a null terminator. Interior nul characters are therefore *not supported*, and will truncate the string
* Invalid UTF-8 will not crash ImGui, and is comparable to a typo in text. A crash is a bug, and should be fixed upstream
* Rust &str slices are guaranteed to be UTF-8
* Rust CString values are guaranteed to have a null terminator and no interior nuls
* Rust supports both string literals with type &'static str (e.g "hello") and byte string literals &'static [u8] (e.g. b"Hello")
* Rust string literals are UTF-8 with support for interior nuls
* Rust byte string literals are ASCII + optional arbitrary bytes with escapes
* Rust source code files are UTF-8

See the [Rust reference](https://doc.rust-lang.org/stable/reference.html#characters-and-strings) for more details about string and byte string literals.

# Syntax choices

Let's look at some different syntax choices (not considering possible conversion using Into):

1. `ui.text(im_str!("Hello"));` (macro)
2. `ui.text(ImStr::new("Hello"));` (ImStr)
3. `ui.text(CString::new("Hello").unwrap());` (CString)
4. `ui.text("Hello");` (string)
5. `ui.text("Hello\0");` (string with explicit null terminator)
6. `ui.text(b"Hello");` (byte string)
7. `ui.text(b"Hello\0");` (byte string with explicit null terminator)

Most people would agree with me that number 4 is the optimal syntax. Number 6 doesn't add any value and in fact makes it impossible to use non-ASCII characters in the literal. Numbers 5 and 7 require the library user to remember the null terminator. Number 1 uses a macro, which is a very custom way of doing things. Numbers 2 are 3 are quite similar from a syntax point of view, but by using a custom ImStr type we can guarantee some extra safety, and can avoid the extra error check by handling interior nuls in a different way.

# Dynamic allocation

One important question is whether passing a string requires dynamic allocation and copying. Dynamic allocation is required in these situations:

* The string requires modifications (e.g. adding null terminator, skipping interior nuls)
* The used type wants to take ownership of the string instead of borrowing it (e.g. CString always makes a copy)

[...]

PR #19, opened 2016-03-20
Improve string handling and add more widgets

bjz:
Replaces `im_str!` in favour of using `Into` trait, and adds support for `Ui::{button, invisible_button, progress_bar}`.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -8,14 +8,14 @@
 ![Hello world](hello_world.png)
 
 ```rust
-ui.window(im_str!("Hello world"))
+ui.window("Hello world")
     .size((300.0, 100.0), ImGuiSetCond_FirstUseEver)
     .build(|| {
-        ui.text(im_str!("Hello world!"));
-        ui.text(im_str!("This...is...imgui-rs!"));
+        ui.text("Hello world!");
+        ui.text("This...is...imgui-rs!");
         ui.separator();
         let mouse_pos = ui.imgui().mouse_pos();
-        ui.text(im_str!("Mouse Position: ({:.1},{:.1})", mouse_pos.0, mouse_pos.1));
+        ui.text(format!("Mouse Position: ({:.1},{:.1})", mouse_pos.0, mouse_pos.1));
     })
 ```
 
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -1,6 +1,5 @@
 #[macro_use]
 extern crate glium;
-#[macro_use]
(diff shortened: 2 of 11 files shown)


bjz:
Weird... for nightly and beta: https://travis-ci.org/Gekkio/imgui-rs/jobs/117204435#L189

```
no matching package named `imgui-sys` found (required by `imgui`)
```

Gekkio:
Thanks!

Unfortunately, I'm not fond of changing everything to use Into in that way. Please check out #7 for thorough discussion about string handling. While Into still retains the possibility of using the im_str! macro and other zero-overhead methods, in practice everyone will just use normal strings if the into() call is moved from user code to the library code. This makes the macro pointless, and encourages bad practice. There's two solutions that are from my point of view acceptable:

* adding the null terminator at compile-time using constexpr or similar feature once Rust has it in the future
* changing dear imgui itself. A such fork exists, but I haven't made a decision yet to use it. The decision is complicated, because a fork without a proper test suite would significantly increase the risk of having subtle but deadly bugs, and would require active maintaining. So far it looks good, but I'm not ready to make the leap yet.

The new widgets look good! I noticed just one thing: progress bar should be usable without an overlay.
In the original API you can pass a null pointer, so we should either use Option or use the builder pattern as used in other places in this library.

The build failure in nightly and beta seems to be caused by the combination of using a version and a path in Cargo.toml. I'll see if I can figure out how to fix it properly, so don't worry about it.

bjz:
Dang, that's a shame. I guess seeing as complex UIs could be using text heavily, performance is a concern. Hopefully string views become part of imgui in the future!

bjz:
Perhaps it would be good to add the reasoning behind `im_str!` in the readme?

bitshifter:
@bjz FYI my fork which removes the need for im_str! while avoiding string copies is here https://github.com/bitshifter/imgui-rs/tree/imstr

bjz:
Nice! :)

Gekkio:
Sorry, I should've linked the fork here since I mentioned it.

> Perhaps it would be good to add the reasoning behind im_str! in the readme?

That's a good idea! It's a significant source for confusion, so I'll mention it in the readme.

In any case, since compile-time functions in Rust seem to be a bit far off, I think I'll probably switch to bitshifter's imgui/cimgui forks soon once I manage to create a basic test suite in this project.

bjz:
That would be super nice from a UX point of view! <3

PR #20, opened 2016-04-13
Updated to glium 0.14

shockham:
Updating dependencies so works with glium 0.14 which had a couple of breaking changes.


Gekkio:
Thanks! Looks good

PR #21, opened 2016-05-06
Make ImStr#as_ptr public

mneumann:
This allows using ```ImStr``` in combination with functions (widgets) exported by ```imgui-sys```.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -86,7 +86,7 @@ impl<'a> ImStr<'a> {
             bytes: Cow::Borrowed(bytes)
         }
     }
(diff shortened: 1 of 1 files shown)


Gekkio:
Looks good! Thanks!

As a side note: is there a widget that is missing from the main crate that you would like to use?

mneumann:
Thanks! Yes, I am using `igCombo2` (https://github.com/mneumann/hypernsga/blob/master/examples/graph_opt_gui.rs#L553) and `igPlotLines2` (https://github.com/mneumann/hypernsga/blob/master/examples/graph_opt_gui.rs#L582). That's why I needed the public `as_ptr()` method.

PR #22, opened 2016-05-25
WIP: Column support

michelson:
Hi , this PR implement the following methods to work with columns:

+ igColumns
+ igGetColumnIndex()
+ igGetColumnOffset()
+ igSetColumnOffset( )
+ igGetColumnWidth()
+ igGetColumnsCount()

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -400,8 +400,8 @@ impl<'ui> Ui<'ui> {
     pub fn spacing(&self) { unsafe { imgui_sys::igSpacing() }; }
 
 
-    pub fn columns(&self, i: i32, b: *const i8, b2: bool){
-        unsafe { imgui_sys::igColumns(i, b, b2 ) };
+    pub fn columns<'p>(&self, i: i32, b:  ImStr<'p>, b2: bool){
+        unsafe { imgui_sys::igColumns(i, b.as_ptr(), b2 ) };
     }
 
     pub fn next_column(&self) {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,7 +38,7 @@ pub use imgui_sys::{
     ImGuiWindowFlags_HorizontalScrollbar, ImGuiWindowFlags_NoFocusOnAppearing,
     ImGuiWindowFlags_NoBringToFrontOnFocus,
     ImVec2, ImVec4,
-    ImGuiKey
+    ImGuiKey 
 };
 pub use input::{InputText};
 pub use menus::{Menu, MenuItem};
@@ -398,7 +398,40 @@ impl<'ui> Ui<'ui> {
         }
     }
     pub fn spacing(&self) { unsafe { imgui_sys::igSpacing() }; }
+
+
+    pub fn columns(&self, i: i32, b: *const i8, b2: bool){
+        unsafe { imgui_sys::igColumns(i, b, b2 ) };
+    }
+
+    pub fn next_column(&self) {
+        unsafe { imgui_sys::igNextColumn() };
+    }
+
+    pub fn get_column_index(&self) {
+        unsafe { imgui_sys::igGetColumnIndex() };
+    }
+
+    pub fn get_column_offset(&self, i: i32) {
+        unsafe { imgui_sys::igGetColumnOffset(i) };
+    }
+
+    pub fn set_column_offset(&self, column_index: i32, offset_x: f32) {
+        unsafe { imgui_sys::igSetColumnOffset(column_index, offset_x ) };
+    }
+
+    pub fn get_column_width(&self, i: i32) {
+        unsafe { imgui_sys::igGetColumnWidth(i) };
+    }
+
+    pub fn get_columns_count(&self) {
+        unsafe { imgui_sys::igGetColumnsCount() };
+    }
+
+
 }
+
+
 
 // Widgets
 impl<'ui> Ui<'ui> {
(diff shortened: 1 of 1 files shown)


Gekkio:
Thanks!

However, I'm sorry but this needs more work.

* Please use meaningful variable names from the upstream imgui API instead of stuff like `b2: bool`
* None of the getters return anything, so they don't make any sense. Have you even tried using this code?
* Are absolutely 100% sure the `<'p>` lifetime is correct? The code is claiming that the `ImStr<'p>` is only used during the function call. Since `igColumns` doesn't print the string, I'm pretty sure it's stored somewhere.
* There's some minor formatting errors. I can fix these myself, but a pull request with correct formatting is always a bit nicer than one without

michelson:
Hi @Gekkio , thanks for the code review.

+ Return types for getters added.
+ Args name changed with the corresponding names of ig methods.
+ I'm not sure about the `<'p>` there was the only way I found for the compiler to no complain, Rust newbie here.
+ If there are any other formatting error I can fix let me know.
+ Also I've added `tree_pop()` method but it raise and error on runtime!, So I've removed that from this PR.


image

## example

```rust
                ui.columns(4, im_str!("mycolumns"), true);
                ui.separator();
                ui.next_column();

                ui.text(im_str!("id"));
                ui.next_column();

                ui.text(im_str!("path"));
                ui.next_column();

                ui.text(im_str!("flags"));
                ui.next_column();

                ui.separator();
```

src/lib.rs
```diff
@@ -398,6 +398,36 @@ impl<'ui> Ui<'ui> {
...
+
+    pub fn next_column(&self) {
+        unsafe { imgui_sys::igNextColumn() }
+    }
+
+    pub fn get_column_index(&self) -> i32 {
+        unsafe { imgui_sys::igGetColumnIndex() }
+    }
+
+    pub fn get_column_offset(&self, i: i32) -> f32 {
```
bjz: `column_index` to be consistent?

src/lib.rs
```diff
@@ -398,6 +398,36 @@ impl<'ui> Ui<'ui> {
...
+        unsafe { imgui_sys::igGetColumnOffset(i) }
+    }
+
+    pub fn set_column_offset(&self, column_index: i32, offset_x: f32) {
+        unsafe { imgui_sys::igSetColumnOffset(column_index, offset_x ) }
+    }
+
+    pub fn get_column_width(&self, i: i32) -> f32 {
```
bjz: ditto.

bjz:
Might be cool to put an example in the err... example.

Gekkio:
Very good, looks better!

    Also I've added tree_pop() method but it raise and error on runtime!, So I've removed that from this PR.

Trees are already supported, so you don't need to add any tree functions:

```rust
ui.tree_node(im_str!("A")).build(|| {
  ui.text(im_str!("Hello"));
  ui.tree_node(im_str!("B")).build(|| {
  });
});
```

    If there are any other formatting error I can fix let me know.

Looks good otherwise, but there's some extra spaces in the code. At least `offset_x )`, `id:  ImStr`, `border )`

    I'm not sure about the <'p> there was the only way I found for the compiler to no complain, Rust newbie here.

This one is actually very tricky. Think about how this would work in C if dynamic strings were used:

```c
  const char *id = malloc(...); // needs a free() call later!
  // fill *id with a proper utf-8 string
  igColumns(1, id, false);
  // 10 000 lines of UI code
  free(id); // <- at which point in the code can this be done safely?
```

The current code claims that this is safe (but I don't think it is):

```c
   igColumns(1, id, false);
   free(id);  
   // 10 000 lines of UI code. If ImGui internally uses id anywhere -> crash
```

In order to answer this question, we need to read the ImGui source code and understand when and how the id string is used.

michelson:
Hi @Gekkio , 
I've tweaked the formatting errors you've point out.
Also I've check the code for the base implementation of ImGui::Column and it seems that the `id` is used to set an unique id to differentiate from others.

```cpp
    // Differentiate column ID with an arbitrary prefix for cases where users name their columns set the same as another widget. 
    // In addition, when an identifier isn't explicitly provided we include the number of columns in the hash to make it uniquer.
    ImGui::PushID(0x11223347 + (id ? 0 : columns_count)); 
    window->DC.ColumnsSetID = window->GetID(id ? id : "columns");
    ImGui::PopID();
```

But I couldn't find where the id is freed, maybe could be free right away with the `ImGui::PopID();` call ?
As far I can see, the `id` is only used to set ColumnSetID , so the `id` var is not used anymore.

Gekkio:
>  I've tweaked the formatting errors you've point out.

Thanks! Looks good

> But I couldn't find where the id is freed

You won't find any `free(id)` calls in ImGui itself, because it is the library users' responsibility to free any dynamic strings that were not allocated by ImGui.

> As far I can see, the id is only used to set ColumnSetID , so the id var is not used anymore.

Good observation! However, we still need to know what `window->GetID` does with the id.

This is the function that gets called:

```cpp
ImGuiID ImGuiWindow::GetID(const void* ptr)
{
    ImGuiID seed = IDStack.back();
    ImGuiID id = ImHash(&ptr, sizeof(void*), seed);
    ImGui::KeepAliveID(id);
    return id;
}
```

Notice how the parameter is of type `const void* ptr` so this `GetID` function works on any pointer. It also means that it doesn't care at all whether the parameter is a string or not. The pointer value (a memory address) is just used for hashing :scream: 

So, to summarize, this means that the original `id: ImStr` is never stored anywhere, and the lifetime `<'p>` is correct. I'll merge the pull request shortly :smile:

Gekkio:
Merged :smile:

Sorry about the close/reopen mess...I didn't look at which button I was clicking

michelson:
Great! , thanks for your time on this PR.
Best

PR #25, opened 2016-05-27
combo and list_box api wrapper

orhanbalci:


Gekkio:
Thanks!

There are a few issues, and I'll add line comments about them.

The formatting is also wrong, and does not follow this project's style or standard Rust style.
In most projects you would just run rustfmt, but I don't have a configuration file for this project yet. Sorry! But you can probably fix most formatting problems with standard Rust style defaults and rustfmt.

src/lib.rs
```diff
@@ -496,3 +496,29 @@ impl<'ui> Ui<'ui> {
     pub fn menu<'p>(&self, label: ImStr<'p>) -> Menu<'ui, 'p> { Menu::new(label) }
     pub fn menu_item<'p>(&self, label: ImStr<'p>) -> MenuItem<'ui, 'p> { MenuItem::new(label) }
 }
+
+//Widgets: Combos 
+impl<'ui> Ui<'ui> {
+    pub fn combo<'p>(&self, label : ImStr<'p>, current_item : &mut i32, items : &mut Vec<ImStr<'p>>, items_count : i32, height_in_items : i32) -> bool{
```
Gekkio: Do we really need:

* a *mutable* container? How is the container modified during this function call?
* to only support *vector*?
* size separately?

This is not very Rust-like

src/lib.rs
```diff
@@ -496,3 +496,29 @@ impl<'ui> Ui<'ui> {
...
+        let items_inner : Vec<*const c_char> = items.into_iter().map(|item| item.as_ptr()).collect();
+        unsafe {
+            imgui_sys::igCombo(label.as_ptr(), current_item, items_inner.as_ptr() as *mut *const c_char, items_count, height_in_items)
+        }
+    }
+
+    pub fn combo2<'p>(&self, label : ImStr<'p>, current_item :&mut i32, items : ImStr<'p>, height_in_items : i32) -> bool {
```
Gekkio: If I've understood correctly, the raw C `igCombo2` functions gets items in a single C string separated with interior NULs. However, using ImStr here is a violation of its semantics, because interior NULs are not allowed in ImStr. This is not actually checked anywhere at the moment, but once Rust gets support for compile-time functions, this rule will be enforced. So, something else than ImStr needs to be used as the function parameter.

src/lib.rs
```diff
@@ -496,3 +496,29 @@ impl<'ui> Ui<'ui> {
...
+        unsafe{
+            imgui_sys::igCombo2(label.as_ptr(), current_item,  items.as_ptr(),  height_in_items)
+        }
+    }
+}
+
+//Widgets: ListBox
+impl<'ui> Ui<'ui> {
+    pub fn list_box<'p>(&self, label : ImStr<'p>, current_item : &mut i32, items : &mut Vec<ImStr<'p>>, items_count : i32, height_in_items : i32)-> bool{
```
Gekkio: Please see my comments about `combo`, because they apply here as well.

src/lib.rs
orhanbalci: - I'll update &mut -> &
- I'll add a new function with slice parameter. how shall I name the function
- I,ll take size from container parameter.

src/lib.rs
orhanbalci: You are right with the interior NULs in parameter. But i could not get why interior NULs should not be allowed in ImStr. What's the problem with current ImStr implementation?

Gekkio: C-style strings do not support interior NULs, because there is no way to distinquish between them and the terminator NUL. In the Rust standard library, the [CString](https://doc.rust-lang.org/std/ffi/struct.CString.html) type has a guarantee that no interior NULs are used.

ImStr is basically CString with two differences:

* Guaranteed UTF-8
* Can be constructed compile-time without runtime cost using the im_str! macro

The interior NULs are not currently checked, but the semantics still apply. Like I said earlier, I might add checks for the semantics at some point, so I don't want code that intentionally breaks them.

Also, I've thought about the `combo2` function, and I don't see what the value of that function is. If we already have `combo`, why would we need another function? Just because the underlying library has it, doesn't mean the safe wrapper has to provide it.

src/lib.rs
Gekkio: Can't we just have one function that takes a slice? I don't see any value in having more than one function:

```rust
pub fn combo<'p>(&self, label: ImStr<'p>, current_item: &mut i32, items: &'p [ImStr<'p>], height_in_items: i32) -> bool { }
```

src/lib.rs
orhanbalci: Thanks for the explanation. For the `code2` function I thought that all underlying functions must be supported. I even tried to implement he one with callback namely 'combo3' using closures but I could not get it working.

orhanbalci:
@Gekkio formatted the code and removed unnecessary functions.

Gekkio:
Looks good! I think I'll merge this now.

There's still some very tiny formatting errors (`let items_inner : Vec`, `-> bool{`), but I'll let those slide, because I'm going to fix all the formatting anyway with rustfmt in the near future.

PR #26, opened 2016-06-03
plotlines api added

orhanbalci:
implemented plotlines api wrapper

### DIFF
--- a/src/plotlines.rs
+++ b/src/plotlines.rs
@@ -7,12 +7,11 @@ use libc::c_float;
 pub struct PlotLines<'p> {
     label: ImStr<'p>,
     values: &'p [f32],
-    values_ofset: i32,
+    values_offset: usize,
     overlay_text: Option<ImStr<'p>>,
     scale_min: f32,
     scale_max: f32,
     graph_size: ImVec2,
-    stride: i32,
 }
 
 impl<'p> PlotLines<'p> {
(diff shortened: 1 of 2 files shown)


src/plotlines.rs
```diff
@@ -0,0 +1,75 @@
+use imgui_sys;
+use super::{ImStr};
+use imgui_sys::ImVec2;
+use std::{f32, mem, ptr};
+use libc::c_float;
+#[must_use]
+pub struct PlotLines<'p> {
+    label: ImStr<'p>,
+    values: &'p [f32],
+    values_ofset: i32,
```
bjz: Am I missing something, or should all these `ofset`s be spelled `offset`?

Gekkio: Yeah, this is a clear typo that should be fixed

Gekkio:
Thanks!

Looks mostly good, but this API is not safe. The problem is the fact that ImGui internally uses unchecked pointer arithmetic to access the data:

```cpp
static float Plot_ArrayGetter(void* data, int idx)
{
    ImGuiPlotArrayGetterData* plot_data = (ImGuiPlotArrayGetterData*)data;
    const float v = *(float*)(void*)((unsigned char*)plot_data->Values + (size_t)idx * plot_data->Stride);
    return v;
}
```

Mistakes with the parameters can lead to:

* misaligned reads
* buffer over-reads
* buffer under-reads

orhanbalci:
What if we remove stride parameter from our wrapper and use default value? Idx parameter is internally managed by the library and should be between [0,size). With this assumptions can not we say that the API is safe?

Gekkio:
> What if we remove stride parameter from our wrapper and use default value? Idx parameter is internally managed by the library and should be between [0,size). With this assumptions can not we say that the API is safe?

Yes, that's a reasonable compromise. We can add the other API with the getter function later, and it will provide a more flexible version for those who need it.

Could you also change the type of offset to usize? Even though we have to cast to i32 in the end, usize makes much more sense in the Rust side of the API, because the offset cannot be negative, and usize is the right type for stuff involving memory, such as array indexes. This also means that values larger than `i32::MAX` will fail, but casts between usize/u32/c_uint and isize/i32/c_int already exist in this library. So it's ok to have that edge case here, and I should think later about how to handle those error cases.

orhanbalci:
Fixed the typo and removed stride parameter. Also updated offset type to usize

Gekkio:
Thanks! Merging now

orhanbalci:
Thanks for all the comments

PR #27, opened 2016-06-09
PlotHistogram API

orhanbalci:
PlotHistogram api is the same as plotlines API just uses different functions. I copied the work done for plotlines. In terms of code reuse this solution is not good but it saves the day.


Gekkio:
Thanks!

orhanbalci:
Thanks for merging

PR #28, opened 2016-06-12
ImGui ImStr fork support

bitshifter:
I thought I'd make a PR to track if this fork is merging and building cleaning. I don't expect it to be merged although if you have any comments that would be useful. If you'd rather not have a PR hanging around feel free to close it.


sanbox-irl:
Hello!

I am going to close this PR, since we now have an `AsRef<str>` implementation in this library.

This approach was completely different, and now that `string-view` has gotten some work in the upstream Dear ImGui library, I think it's time to close this PR. I understand it was primarily to document the effort, and for that work, thank you!

PR #29, opened 2016-06-13
Add hidpi support

### LINKED ISSUE #18: Examples should show how to take into account high-dpi screen resolutions
Glutin allows you to query the high-dpi factor by calling `window.hidpi_factor()`. It would be good if the examples showed how to scale imgui by that factor so that it is not so tiny!

screen shot 2016-03-16 at 9 22 57 pm

bjz:
Closes #18.

screen shot 2016-06-13 at 2 51 11 pm

Not sure about the API design for this one. I initially had the user handle the hidpi factor themselves (see 
https://github.com/bjz/imgui-rs/commit/734eda2d5752e2f5c152ae9c61708aa961dab7c2), but I decided to move it into the `ImGui` struct and move it behind a setter. This has the least impact downstream, but requires users to remember to set the hidpi factor before setting mouse values, and means we have to do conversions when accessing/setting the mouse position and mouse wheel values.

Thoughts?

cc. ocornut/imgui#287, ocornut/imgui#426

### DIFF
--- a/examples/support/mod.rs
+++ b/examples/support/mod.rs
@@ -56,10 +56,10 @@ impl Support {
         }
     }
 
-    pub fn update_mouse(&mut self) {
-        self.imgui.set_mouse_pos(self.mouse_pos.0 as f32, self.mouse_pos.1 as f32);
+    pub fn update_mouse(&mut self, hidpi_factor: f32) {
+        self.imgui.set_mouse_pos(self.mouse_pos.0 as f32 / hidpi_factor, self.mouse_pos.1 as f32 / hidpi_factor);
         self.imgui.set_mouse_down(&[self.mouse_pressed.0, self.mouse_pressed.1, self.mouse_pressed.2, false, false]);
-        self.imgui.set_mouse_wheel(self.mouse_wheel);
+        self.imgui.set_mouse_wheel(self.mouse_wheel / hidpi_factor);
     }
 
     pub fn render<'ui, 'a: 'ui , F: FnMut(&Ui<'ui>)>(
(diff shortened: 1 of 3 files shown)


Gekkio:
Thanks!

Hmm, what about `io.display_frame_buffer_scale`?
I have absolutely no clue how the API should look, but it looks like there's some differences compared to the hidpi support code in the examples:

* ImDrawData.ScaleClipRects vs manual scaling
* I don't see any mouse coordinate scaling in the upstream examples...? If this is specific requirement for glutin, the hidpi_factor parameter should not be required in `update_mouse` in the main imgui-rs API

Also, could we pass a &Ui to the draw list rendering closure instead of hidpi_factor? In my opinion it could be cleaner since the renderer could access other information as well.

examples/support/mod.rs
```diff
@@ -57,28 +57,34 @@ impl Support {
...
         target.clear_color(clear_color.0, clear_color.1,
                            clear_color.2, clear_color.3);
 
-        let (width, height) = target.get_dimensions();
-        let ui = self.imgui.frame(width, height, delta_f);
-        f(&ui);
+        // Early return if the window closes before we get the sizes
+        let window = match self.display.get_window() { Some(x) => x, None => return };
+        let size_points = match window.get_inner_size_points() { Some(x) => x, None => return };
+        let size_pixels = match window.get_inner_size_pixels() { Some(x) => x, None => return };
```
bjz: Not sure about this... :/

src/glium_renderer.rs
```diff
@@ -67,21 +68,33 @@ impl Renderer {
...
+        let (scale_width, scale_height) = ui.imgui().display_framebuffer_scale();
+
+        if width == 0.0 || height == 0.0 {
+            return Ok(());
+        }
+
+        let matrix = [[2.0 / width as f32, 0.0, 0.0, 0.0],
                       [0.0, 2.0 / -(height as f32), 0.0, 0.0],
                       [0.0, 0.0, -1.0, 0.0],
                       [-1.0, 1.0, 0.0, 1.0]];
```
bjz: Tried to keep this in line with: https://github.com/ocornut/imgui/commit/e58f99179aa3e8fa78e783f772c31ee23d40863d#diff-47471da84bd4194a17a350e728b638b1R45

bjz:
bump

Gekkio:
Thanks! Merging now

bjz:
Woo 🎊

PR #35, opened 2016-11-08
Use character input instead of string input.

bitshifter:
Fixes input on Linux at least.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -260,14 +260,8 @@ impl ImGui {
         io.key_map[key as usize] = mapping as i32;
     }
     pub fn add_input_character(&mut self, character: char) {
-        // TODO: This is slightly better. We should use char::encode_utf8 when it stabilizes
-        // to allow us to skip the string intermediate since we can then go directly
-        // to bytes
-        let mut string = String::new();
-        string.push(character);
-        string.push('\0');
         unsafe {
-            imgui_sys::ImGuiIO_AddInputCharactersUTF8(string.as_ptr() as *const i8);
+            imgui_sys::ImGuiIO_AddInputCharacter(character as u16);
         }
     }
     pub fn get_time(&self) -> f32 { unsafe { imgui_sys::igGetTime() } }


Gekkio:
Thanks!

I'm not familiar with the semantics of `AddInputCharacters` and I would be very surprised if imgui itself had good Unicode support, but this change adds assumptions to our code so here are some questions:

* Why does this fix input? What's wrong with passing UTF-8? I'm running Linux and haven't had any problems
* What's the meaning of a 16-bit character in this case? In the old version we were passing UTF-8 data, which makes perfect sense, but now we are truncating a 32-bit unicode scalar value into a 16-bit value. What's the meaning of this value? 16-bit unicode scalar? Something to do with UTF-16?
* If it's UTF-16, what happens if we pass a surrogate code point? Can we call this function again to form a valid surrogate pair?
* What happens if I pass `0x10000`, which gets truncated to `0x0000` and happens to be the null terminator?
* What happens if I pass valid 32-bit scalar values which become surrogate code points when truncated to 16 bits?

bitshifter:
Sorry, I did this really quickly at a Rust meetup/hack night. It does appear to be working fine so, the problem must have been caused by something else.

Partially I changed it as the existing code is allocating a string every time a character is typed, which seems counter to the intention of imgui, as it's very careful not to allocate memory.

To answer your other questions, the input.rs `add_input_character` function takes a `char` type as input, so I don't think it is possible to pass a surrogate pair, `0x10000` or a 32 bit scalar to this function.

Looking at how `AddInputCharater` and `AddInputCharactersUTF8` are used in the imgui examples, it appears to depend on platform. DirectX/Windows examples casts a `WPARAM` to `unsigned short` and passes it to `AddInputCharacter`, the SDL examples call `AddInputCharactersUTF8`. So perhaps imgui-rs should actually be mimicking the imgui interface, providing both `u16` and `&str` input methods?

The comment for `AddInputCharater` says it expects ASCII, even though it's taking an `unsigned short`.

Anyway, I'll close this since it's not a bug and this solution probably isn't the right way to go.

Gekkio:
Thanks for the research!
You're absolutely right about the string allocation...it's a very significant wart that I'd like to fix at some point.

bitshifter:
Ah, I was mistaken about the `char`, it's 4 bytes in Rust, not 1 like C.

PR #38, opened 2016-12-30
Added ImInput to make it easier to handle cstr input.

bitshifter:
I was using text input to input a file name in a project I was working on and was finding it a bit unwieldy to use, both in the creation of a `String` suitable to pass to `input_text` and then getting something usable on the Rust side again once it's been edited. In particular, the `String` is fulled with `\0`s which means the length as far as Rust is concerned is effectively the capacity.

For example `println!("\"{}\"", state.text)` would result in `"hello world!  ... (to capacity) ..."`. Passing it to the PathBuf functions would cause a panic due to internal nuls, etc.

To get a string of the correct length for Rust I was passing this `String` to `ffi::CStr`, calling `to_str` on that and then getting the length (of the null terminated string). I thought it would be nicer to try and encapsulate all of this complexity and added the `ImInput` type for this purpose.

What do you think of this approach? Perhaps this functionality would be better off being added to `ImStr`?

### DIFF
--- a/examples/test_window_impl.rs
+++ b/examples/test_window_impl.rs
@@ -4,7 +4,6 @@ extern crate glium;
 extern crate imgui;
 
 use imgui::*;
-use std::iter::repeat;
 
 use self::support::Support;
 
@@ -31,10 +30,10 @@ struct State {
     no_menu: bool,
     bg_alpha: f32,
     wrap_width: f32,
-    buf: String,
+    buf: ImInput,
     item: i32,
     item2: i32,
-    text: String,
+    text: ImInput,
     i0: i32,
     f0: f32,
     vec2f: [f32;2],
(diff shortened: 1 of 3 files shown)


Gekkio:
I've got a WIP branch that implements separate owned (ImString) and borrowed (&ImStr) types that would probably fix this use case. The basic idea is something like this:

```rust
   let mut my_input = ImString::new("Initial value").expect("Input had interior nuls");

   // pass &my_input or &mut my_input to imgui-rs functions
   some_imgui_function(&my_input);
   some_imgui_input_function(&mut my_input);

   let works_as_str: &str = &my_input; // Deref coercions: ImString -> &ImStr -> &str
```

The deref coercions enable you to read and process the ImString as if it was a normal Rust &str. Note that this can only give you an immutable &str, because a mutable &mut str could be used to insert interior nuls and break the rules.

Gekkio:
#46 was merged recently and included the string overhaul I mentioned earlier. Any comments or do you notice any necessary things that are missing?

In the end the owned type looks pretty much the same as yours in this pull request but with just some minor differences. I'm not a huge fan of how capacity works, but it's pretty much unavoidable since imgui can only work with a preallocated buffer.

Gekkio:
I'll close this since the overhaul last year seems to have fixed the use cases. Please open a new pull request and/or report issues if there's problems or something is missing. Thanks!

PR #41, opened 2017-02-06
glium 0.16

erethedaybreak:
Just updating the glium version. Without this I was getting errors for `GlutinFacade` not implementing `Facade`.


Gekkio:
Thanks!

PR #43, opened 2017-02-18
Added progress bar

ShadowIce:


ShadowIce:
I don't know why travis fails. rustup.sh fails on macOS, but that should not have anything to do with my changes....?

Gekkio:
Thanks!

Don't worry about the travis fail. I'll restart the build and will merge even if it fails :sunglasses:

PR #45, opened 2017-03-14
Add push and pop item width

andystanton:


Gekkio:
Thanks!

Did you check what happens if you call these functions in an unbalanced way (e.g. one push, two pops)?
Probably aborts the process like many other push/pop function pairs. If so, I'll still accept this pull request, but I'd like to know first if there's anything we could do to make the API safer.

andystanton:
## Too many pops:

Error on compile:

```sh
Assertion failed: (Size > 0), function pop_back, file third-party/cimgui/imgui/imgui.h, line 895.
```

## Too many pushes:

No error on compile.

andystanton:
As for making the API safer, how about this instead:

```rust
fn push_item_width(&self, width: f32) { unsafe { imgui_sys::igPushItemWidth(width) } }
fn pop_item_width(&self) { unsafe { imgui_sys::igPopItemWidth() } }

pub fn with_item_width<F>(&self, width: f32, f: F) where F: FnOnce() {
    self.push_item_width(width);
    f();
    self.pop_item_width();
}
```

Usage:

```rust
ui.with_item_width(-1.0, || {
    ui.slider_float(ImStr::from(format!("##v{}y", i)), &mut vertex.position[1], -1.0, 1.0).build();
});
```

Gekkio:
Are you sure you meant "on compile" and not "at runtime"? I don't think it's possible to detect mismatched pops and pushes in general, although maybe in a simple case this time the C compiler managed to do it...? In the general case an assertion failure will abort the process at runtime, but this is not memory unsafe in Rust terms, so the functions don't have to be unsafe and the developer using this API will just have to be careful.

I like the closure-based API and would probably use it myself if I had the choice. How about supporting both ways? That way the API user can choose between something that might abort, and something that is safe but a bit annoying to use if there's multiple levels of nesting or a complicated control flow. Another option could be to use a "guard value" that has `impl Drop`, but in my opinion the closure-based API is much clearer and you can easily see how long the effect of the function is.

So, could you please add the closure-based `with_item_width` function while keeping the push and pop functions unchanged? :smiley: I'll merge after that change.

andystanton:
Thanks for the feedback @Gekkio! Yes, you're correct - the failure was at runtime.

I've made the change to include the closure-based function as well as the original push/pull functions and squashed the commit.

Gekkio:
Thanks! Merging :+1:

PR #48, opened 2017-04-25
Rustfmt and clippy-lints fixes

mulimoen:
The transmutes have been replaced acording to [this](https://github.com/Manishearth/rust-clippy/wiki#transmute_ptr_to_ref) lint. Clippy also revealed a [mut_from_ref](https://github.com/Manishearth/rust-clippy/wiki#mut_from_ref) error.

### DIFF
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -27,10 +27,10 @@ fn hello_world<'a>(ui: &Ui<'a>) {
     ui.window(im_str!("Hello world"))
         .size((300.0, 100.0), ImGuiSetCond_FirstUseEver)
         .build(|| {
-            ui.text(im_str!("Hello world!"));
-            ui.text(im_str!("This...is...imgui-rs!"));
-            ui.separator();
-            let mouse_pos = ui.imgui().mouse_pos();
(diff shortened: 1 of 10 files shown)


Gekkio:
Thanks!

Those lints are spot on...I really should run clippy more often :stuck_out_tongue: 

I'm a bit torn about rustfmt, because the defaults keep changing in new versions and now we get a lot of formatting changes that an earlier version didn't suggest at all.
I'll accept this PR anyway, but I should probably make some decisions about formatting and override all the defaults in the config file to have more consistent formatting with different rustfmt versions.

For example, I really don't think this

```rust
.build(|| {
           ui.text(im_str!("Hello world!"));
           ui.text(im_str!("This...is...imgui-rs!"));
           ui.separator();
           let mouse_pos = ui.imgui().mouse_pos();
           ui.text(im_str!("Mouse Position: ({:.1},{:.1})", mouse_pos.0, mouse_pos.1));
       })
```

is better than this

```rust
.build(|| {
    ui.text(im_str!("Hello world!"));
    ui.text(im_str!("This...is...imgui-rs!"));
    ui.separator();
    let mouse_pos = ui.imgui().mouse_pos();
    ui.text(im_str!("Mouse Position: ({:.1},{:.1})", mouse_pos.0, mouse_pos.1));
})
```

PR #50, opened 2017-05-07
Changed string pointer casts to be inferred for cross-compiling to ARM.

ostrosco:
I can into an issue while cross-compiling to arm-unknown-linux-gnueabihf where the pointer types for strings weren't matching as ARM expects u8 instead of i8.


Gekkio:
Thanks!
ARM hasn't been tested at all, so please report/fix any other issues you might find too 😃

PR #52, opened 2017-05-28
Add radiobuttons

mulimoen:


Gekkio:
Thanks for the pull request! :smiley: 

In this case, a simple function call is probably better. The builders don't have any functions other than `new` and `build`, so they are not very useful.

PR #56, opened 2017-07-01
Expose more functionality from imgui.

bjadamson:
Implement new_line(), calc_text_size(), and with_style_var_pushed_*()
methods I needed while making a demo.

I'm pretty new to Rust, but ran into this issue and thought I could
maybe solve it while I was learning the library.
https://github.com/Gekkio/imgui-rs/issues/39

I'm curious on your thoughts on these additions. Please LMK if you would
prefer a different direction, I'm pretty new to both Rust and this
library so please let me know if you have any suggestions!

I'm hoping to contribute to this library moving forward, as I work on
my game's UI.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,7 +18,7 @@ pub use imgui_sys::{ImDrawIdx, ImDrawVert, ImGuiInputTextFlags, ImGuiInputTextFl
                     ImGuiInputTextFlags_ReadOnly, ImGuiKey, ImGuiSelectableFlags,
                     ImGuiSelectableFlags_DontClosePopups, ImGuiSelectableFlags_SpanAllColumns,
                     ImGuiSetCond, ImGuiSetCond_Always, ImGuiSetCond_Appearing,
-                    ImGuiSetCond_FirstUseEver, ImGuiSetCond_Once, ImGuiStyle, ImGuiTreeNodeFlags,
+                    ImGuiSetCond_FirstUseEver, ImGuiSetCond_Once, ImGuiStyle, ImGuiStyleVar, ImGuiTreeNodeFlags,
                     ImGuiTreeNodeFlags_AllowOverlapMode, ImGuiTreeNodeFlags_Bullet,
                     ImGuiTreeNodeFlags_CollapsingHeader, ImGuiTreeNodeFlags_DefaultOpen,
                     ImGuiTreeNodeFlags_Framed, ImGuiTreeNodeFlags_Leaf,
@@ -44,6 +44,7 @@ pub use progressbar::ProgressBar;
 pub use sliders::{SliderFloat, SliderFloat2, SliderFloat3, SliderFloat4, SliderInt, SliderInt2,
                   SliderInt3, SliderInt4};
 pub use string::{ImStr, ImString};
+pub use style::StyleVar;
 pub use trees::{CollapsingHeader, TreeNode};
 pub use window::Window;
 
@@ -54,6 +55,7 @@ mod plotlines;
 mod progressbar;
 mod sliders;
 mod string;
+mod style;
 mod trees;
 mod window;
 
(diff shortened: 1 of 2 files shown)


bjadamson:
Oops, looks like I had a compilation error in my documentation code. Cool that rust caught that, I pushed a forced update to fix this error, hopefully the above error will pass this time.

bjadamson:
Ok, looks like that was the problem. Yay!

Gekkio:
Thanks!

Here are some small random thoughts:

* `new_line` is fine :smiley: 
* I think there's no need for `calc_text_size_with_buffer`. A simple value like ImVec2 can be just returned as a value (= what `calc_text_size` does, so just keep that function and remove the other)
* You could drop the unnecessary `<'p>` lifetimes since they are not used anywhere

Now, about the style var stuff: I'd like to see a more "Rusty" API instead of a straightforward wrapping of the raw imgui-sys API. The main two problems here are:

* There's no way to tell which style var is which type, and the compiler won't help at all. Is `GrabMinSize` a `f32` or a `ImVec2`? What happens if I call the wrong function? The underlying imgui library will abort the process -> not cool
* Multiple levels of nested closures are needed to set multiple style vars. The problem is not about the use of closures in the API, but the inability to configure multiple variables at once.

One alternative way to design the API could be to add a proper Rust enum with data (not just C-style "choice of N options") instead of exporting the imgui-sys raw enum. You could then include the value type information in the enum itself and could easily support multiple style vars per function call.

```rust
#[derive(Copy, Clone, Debug, PartialEq)]
pub enum StyleVar {
  // Each enum variant carries the value. This is type-safe and impossible to misuse
  Alpha(f32),
  WindowPadding(ImVec2),
  // etc...
}

// You'll need to match each style var variant with the right function call and raw enum value
fn push_style_var(style_var: StyleVar) {
  use StyleVar::*;
  use imgui_sys::{igPushStyleVar, igPushStyleVarVec};
  match style_var {
    Alpha(v) => unsafe { igPushStyleVar(ImGuiStyleVar::Alpha, v) },
    WindowPadding(v) => unsafe { igPushStyleVarVec(ImGuiStyleVar::WindowPadding, v) },
    // etc...
  }
}

// Now you only need one function to push any of the supported style vars
pub fn with_style_var<F: FnOnce()>(&self, style_var: StyleVar, f: F) {
  push_style_var(style_var);
  f();
  unsafe { imgui_sys::igPopStyleVar(1) }
}

// Supporting multiple is also easy since you can freely mix and match them in a safe manner
pub fn with_style_vars<F: FnOnce()>(&self, style_vars: &[StyleVar], f: F) {
  for style_var in style_vars {
    push_style_var(style_var);
  }
  f();
  unsafe { imgui_sys::igPopStyleVar(style_vars.len()) };
}
```

I didn't actually compile this so it's just a guideline, but something like this could be one alternative.

bjadamson:
Thanks a bunch for your feedback, as a newcomer to rust I extra-appreciate it! I'm excited to try your suggestions out, keep your eye out for an update to this PR.

;P

bjadamson:
The only thing that makes me unhappy with doing things this way is now all all return values will be the size of a ImVec2, even when a single f32 is all that is required. This will take twice as much space on the CPU cache line, instead if the user didn't have this enum abstraction in front of them. Game programmers seem to care about this kind of stuff.

Do you have any thoughts regarding this tradeoff here?

bjadamson:
Meh, I guess I'm pretty happy with this after all. If the user finds this to be their bottleneck, they can work around the enum I suppose. Would you like me to make any further changes here?

Thanks again for your help!!

Gekkio:
Yeah, I would be very surprised if StyleVar things would end up being a bottleneck. If you really wanted to fiddle with things, you could mark `push_style_var` with `#[inline]`. In a simple naive test this made the compiler optimize the entire `with_style_var function` into a small amount of code if a hard-coded StyleVar was used (this is the typical case). But there's no point unless we have some real-world case where this is an issue.

I'm ready to merge, but now this pull request has a lot of other commits. I'd prefer keeping things separate, because I still have feedback about other things.

bjadamson:
Whoops, sorry about that. I pushed my merged branch as master accidentally. It's fixed now, should only contain the intended fixes. I also marked the functions inline as requested, excited to get it merged!

Gekkio:
Almost there 😃 
Just one thing about inlining: I meant the private `push_style_var` function, not the public functions (`with_style_var` / `with_style_vars`). Will merge after that tiny fix

bjadamson:
Ooh. Okay, sure I can add no prob. Do you prefer them not to be on the public functions?

Gekkio:
Thanks! 😄

PR #57, opened 2017-07-06
Initial window support, looking for ideas/help

bjadamson:
Hello!  I needed to expose more functionality from imgui. Here's my initial idea for supporting child frame's within a window.

    Imgui allows the user to create child frame's, and render into them.
    This isn't exposed currently, this is my first idea at how rust can
    support child frame's.

    Presently it's not the easiest to use, to have a window with a child
    frame supported, internally the imgui library must have called
    beginWindow() for the parent window, before beginChild() is ever called
    for the child frame. Doing this without causing unneeded
    allocations/complexity, and making the API ergonomic is something I hope
    to work on next / get some feedback on.

Functionally, I believe this code to be correct. I have an example of how client code looks using this, and how it renders on my machine here:
imgui_child_window

I'm pretty unhappy that I ended up with a second self-consuming build() method on the window builder. The fact that I need to pass in two closure's forced me to do this for now, at least until I figure out how to store a closure without a dynamic allocation.

Things I'd like to make better / get some help with:

1. Allow the user to pass in a closure to a member function of the window builder, instead of an additional parameter to the build_with_child_frame() method, add another method to the Window type, that takes the user's function and store's it (so it can be called during build()).
    *  I did some work on this, I added a new field to the Window struct which had the type Option<fn() -> ()>, but the compiler wouldn't let me pass in a closure and store it as a plain function pointer. Do you have any ideas on how we might store a function pointer / closure without an additional allocation? I would think an additional data member on the Window struct would be fine, as long as it is not dynamically allocated.

2. Consolidate back to a single build() method taking a single closure, like the original design, still supporting child frame's though. Right now it's awkward to pass in both closure's like this, the original design made more sense. I get this for free by completing #1 above. If the user can pass in a closure for constructing the child frame, there's no need for the second build_with_child_frame() method.

3. Support arbitrarily nested child-frames. Imgui supports nesting arbitrary frame's, right now this design only supports a maximum of one child frame.

### DIFF
--- a/src/child_frame.rs
+++ b/src/child_frame.rs
@@ -1,0 +1,126 @@
+use std::mem::transmute;
+
+use ImVec2;
+use ImGuiWindowFlags;
+
+use super::{ImGuiWindowFlags_NoTitleBar, ImGuiWindowFlags_NoResize, ImGuiWindowFlags_NoMove,
+            ImGuiWindowFlags_NoScrollbar, ImGuiWindowFlags_NoScrollWithMouse,
+            ImGuiWindowFlags_NoCollapse, ImGuiWindowFlags_AlwaysAutoResize,
+            ImGuiWindowFlags_ShowBorders, ImGuiWindowFlags_NoSavedSettings,
+            ImGuiWindowFlags_NoInputs, ImGuiWindowFlags_MenuBar,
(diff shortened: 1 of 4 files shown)


bjadamson:
opened against wrong base branch, closing in favor of:
https://github.com/Gekkio/imgui-rs/pull/58

Issue #59, opened 2017-07-07
Implement std::ops::Add for ImVec2 (maybe ImVec4 too?)


bjadamson:
Ran into this paper cut, I can't think of a reason not to do this. It's not hiding any allocations, I think it's strictly an improvement if this trait were implemented for ImVec2 and  ImVec4.

What do you think Gekkio, if I figured out how to implement this would you support it?

bjadamson:
Closed as discussed in issue #61

PR #60, opened 2017-07-07
Implement (atleast partial) fix for issue #59.

bjadamson:
Implement fix for issue #59.

This is my first crack at issue 59, which I admit I opened myself. This commit contains the first macro's I've ever written, so I apologize ahead of time for the duplication across the macros. I'm a beginner at macros, so these are beginner macros :package: 

If you see a way to reduce the duplication across the different functions, please do let met know!

### DIFF
--- a/imgui-sys/src/lib.rs
+++ b/imgui-sys/src/lib.rs
@@ -12,6 +12,8 @@ extern crate glium;
 
 use std::convert::From;
 use std::mem;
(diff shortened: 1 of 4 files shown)


bjadamson:
Opened against wrong base, closing in favor of:
https://github.com/Gekkio/imgui-rs/pull/61

PR #61, opened 2017-07-07
Implement fix for issue #59.

bjadamson:
Implement fix for issue #59.

This is my first crack at issue 59, which I admit I opened myself. This commit contains the first macro's I've ever written, so I apologize ahead of time for the duplication across the macros. I'm a beginner at macros, so these are beginner macros :package: 

If you see a way to reduce the duplication across the different functions, please do let met know!

### DIFF
--- a/imgui-sys/src/lib.rs
+++ b/imgui-sys/src/lib.rs
@@ -12,6 +12,8 @@ extern crate glium;
 
 use std::convert::From;
 use std::mem;
+use std::ops::{Add, Sub, Mul, Div, Rem};
+use std::ops::{AddAssign, SubAssign, MulAssign, DivAssign, RemAssign};
 use std::os::raw::{c_char, c_float, c_int, c_short, c_uchar, c_uint, c_ushort, c_void};
 use std::slice;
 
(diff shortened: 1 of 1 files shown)


Gekkio:
Hmm, I'm not sure how I feel about this. Regardless of how they are named, ImVec2/4 types are just tuples, and I don't think tuples should have these operators in general. Also, these types are mostly meant for temporary conversion to support passing values to imgui, and use of normal tuples (e.g. `(1.0, 2.0)` and `(1.0, 2.0, 3.0, 4.0)` is encouraged instead. If you intend to do some actual linear algebra, then you should use a proper library that implements the operations you need (e.g. nalgebra, cgmath).

If we also think about what the ImVec2/4 represent in the library, many operations are nonsensical. ImVec4 values are in practice either colours or (X, Y, W, H) rectangles, and for example element-wise addition makes no sense for any combinations of these values.

Do you have a practical example where you needed this functionality? It would help me understand the value vs cost vs confusion trade-offs here a bit more.

bjadamson:
Yeah, after reading your comments, I agree. My use case actually came from being new to rust it turns out. 
 The From trait actually covers my use-case and I no-longer need this. We can close this from my point of view.

PR #62, opened 2017-07-08
Expose color variable stack to user.

bjadamson:
In order to have colored and wrapped text, I needed to use this
functionality in my game. This PR implements a version similar to
with_style_var().

From my testing it is safe to use a Vec4 for all the variants of the
enumeration, so I didn't bother putting an enumeration in imgui-rs, I
just reused the enumeration defined in imgui-sys directly.

Let me know if you have any feedback on this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,7 +18,7 @@ pub use imgui_sys::{ImDrawIdx, ImDrawVert, ImGuiInputTextFlags, ImGuiInputTextFl
                     ImGuiInputTextFlags_ReadOnly, ImGuiKey, ImGuiSelectableFlags,
                     ImGuiSelectableFlags_DontClosePopups, ImGuiSelectableFlags_SpanAllColumns,
                     ImGuiSetCond, ImGuiSetCond_Always, ImGuiSetCond_Appearing,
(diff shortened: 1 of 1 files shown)


bjadamson:
Hey, I created this branch (I'm working with locally) that has all the changes from my various PR's merged into one branch. It would be easier to merge I believe.

https://github.com/Gekkio/imgui-rs/compare/master...bjadamson:bjadamson-changes-merged

Gekkio:
Looks pretty good!

Yes, you're right that an enumeration is not needed here.
Two comments:

* Use `Into<ImVec4>>` mechanism in `with_color_var`
* I don't think the typical use case for `with_color_vars` would be to set many color vars to the same value. Maybe accepting `&[(ImGuiCol, ImVec4)]` would be better?

bjadamson:
Agreed on both points! I agree, I think I was pretty tired when I came up with the first interface for with_color_vars. Thanks for suggesting the improvement. I was able to use the Into<ImVec4> mechanism again in the with_color_vars() function as well. This is what you intend, yes?

bjadamson:
I can resolve the conflicts this one after the other merge gets accepted, if you like. I just happened to do that one first, and I don't think resolving this conflict now will help us after that one is merged.

PR #64, opened 2017-07-22
Update glutin and dependencies that rely on it

brendanzab:
I tried to also make the gfx and glutin support modules more consistent in the examples.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ rust:
   - stable
   - beta
   - nightly
-  - 1.16.0
 os:
   - linux
   - osx
--- a/README.markdown
+++ b/README.markdown
@@ -2,7 +2,7 @@
 
 **Still fairly experimental!**
 
-Minimum Rust version: 1.16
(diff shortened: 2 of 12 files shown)


brendanzab:
Hmm, seems like `wayland-client` doesn't like 1.16. 🤔

Gekkio:
Thanks!

I suppose we'll have to bump the minimum version in `travis.yml` and `README.markdown` :disappointed: 
I'd prefer offering better backwards compatibility, but there's not much we can do if dependencies aggressively use features available only in recent Rust versions. It's not even the first time I've had to bump the minimum Rust version because of `wayland-client`.

brendanzab:
done :(

Gekkio:
Thanks! :) :( :)

brendanzab:
Probably would be a good idea to release this as soon as you're able because it's blocking our work on mac OS.

Gekkio:
0.0.15 is here :tada:

PR #67, opened 2017-07-27
Add shader switching

andystanton:
Add dependency on gfx_app


andystanton:
I'm closing this as it's a bad solution and lots of things are changing in the next version of gfx.

PR #70, opened 2017-08-15
Switch to using virtual cargo workspace.

bjadamson:
Are you open to something like this?


bjadamson:
NOTE: In the proposed module rework, it seems they assume that projects will be arranged with a src/ directory. 

https://github.com/withoutboats/rfcs/blob/modules/0000-modules/overview.md

Gekkio:
Hmm, could you explain a bit more? It seems to be mostly renames and slight changes to the directory structure since a Cargo workspace is already used, so I don't understand the reason for these changes.

bjadamson:
Sorry, this switches cargo to use a [virtual manifest](http://doc.crates.io/manifest.html#virtual-manifest). I tried to normalize the project structure to be forward compatible with what seems to be the expected project structure. Quite simply I think it simplifies the project structure, or atleast makes it [more intuitive](https://github.com/bjadamson/imgui-rs/tree/cargo-virtual-workspace/src) to someone new.

thomcc:
This is very out of date and while I'll switch to a virtual workspace soon, this won't be the pattern I use.

PR #74, opened 2017-09-02
Added functions in lib for radios

Palmr:
I know someone else has an open pull request with a radio button implementation but it's been sat open for a while and seems a little overly-complex, so I made the same change but without the builder.


Gekkio:
Thanks!

PR #75, opened 2017-09-04
expose a method to set font_global_scale

onelson:
I noted that you hoped PRs would come with a `cargo fmt` included, but I saw there was _a lot_ of reformatting done by this, so I thought I'd offer to do this as a follow-up rev if required.


Gekkio:
Thanks!

It's ok, I should actually remove `cargo fmt` from the instructions because they keep changing the defaults and the formatting results change all the time :sleepy:

PR #79, opened 2017-09-18
Work in progress - update to latest gfx

andystanton:
This is a work in progress branch bringing compatibility with the bleeding edge changes in gfx, using their examples as a reference.

It isn't intended to be considered for merge yet, but might be useful for the next major release of gfx.

The gfx example can be run from the imgui-examples directory with this command:

```
cargo run --example hello_gfx
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 target
 Cargo.lock
 imgui.ini
+.idea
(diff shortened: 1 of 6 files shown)


andystanton:
Closing this - gfx is a very fast moving target at the moment.

Issue #83, opened 2017-10-13
Custom font support.


aleksander:
It would be greate to have custom TTF font support.

PR #84, opened 2017-10-13
Custom font support

aleksander:
Draft for the #83. Not sure about API.


anderejd:
Looks like a good start.

You have to handle the null pointer case though since null can be returned on failure:
https://github.com/ocornut/imgui/blob/06a32a2b00b8d974c773d3b5f7abcd811de5d0b5/imgui_draw.cpp#L1526

aleksander:
> You have to handle the null pointer case

Done. But I'm not sure I did error handling right. `failure` may be used.

anderejd:
Looks good to me :) but @Gekkio needs to review and merge this.

Regarding failure, I suppose you are referring to the crate https://crates.io/crates/failure ? That looks really good imho and the author withoutboats have made a good impression on me in the past, he is also since recently working for mozilla, on the Rust team (I think).

Gekkio:
Thanks for the pull request! :smile: 

And sorry, I haven't had time to think about this much yet. I'll try to look at it during the weekend. Some quick points:

* `set_font` is a misleading name, because if you call this function multiple times, fonts are added and not replaced. It *can* replace the default font, but only if you call it only once and at the right time. I'm not sure what we should call this :frowning_face:. Maybe `add_font` would be better since that matches the upstream name...?
* How do you use this if you want multiple fonts?
* What happens if you call this after a renderer is initialized?
* What happens if you load invalid TTF data?
* Could you add a realistic example to imgui-examples that uses this function?
* There's actually no null pointer case so this new error handling is unnecessary. If you look at the upstream code, you'll see that there's `IM_ASSERT(0)` which aborts the process with an assertion failure, so the Rust code is never executed in the error case. It's in theory possible to disable the asserts in dear imgui with `NDEBUG` but `imgui-rs` doesn't use this so they are always active.
* `&str` is a strange type for file names (&Path is more appropriate). But I think in this case it's correct because dear imgui seems to assume the file name is UTF-8...?
* Could we use `AddFontFromMemoryTTF` instead? This would let us keep the file loading logic in Rust code and handle errors better than with a random process abortion. With that function we're not even handling paths so NulError and filename encoding complexities disappear.

anderejd:
> There's actually no null pointer case so this new error handling is unnecessary. If you look at the upstream code, you'll see that there's IM_ASSERT(0) which aborts the process with an assertion failure, so the Rust code is never executed in the error case. It's in theory possible to disable the asserts in dear imgui with NDEBUG but imgui-rs doesn't use this so they are always active.

I think I did look in the wrong imgui header, as you say it is defined as a normal assert:
https://github.com/ocornut/imgui/blob/3f5b2a3fe353ff29b1b15334964aaec4e7dff576/imgui.h#L27

Must have found this and missed that it is commented out:
https://github.com/ocornut/imgui/blob/8fd5620277f3d89605ff8e2c286d744194f89e25/imconfig.h#L10

I agree that following the imgui default and let imgui assert/abort on programmer error is a good choice for now.

PR #85, opened 2017-10-23
Blurry fonts on glium renderer

luteberget:
Hey, thanks for making this library.
I tried loading a custom TTF font using imgui-sys, but the imgui-glium-renderer then renders blurry text, like this:
screenshot_2017-10-23_22-01-02

The cause seems to be that glium sets `minify_filter` to `MinifySamplerFilter::LinearMipmapLinear` by default. Text renders correctly if this property is set to `Linear` or `Nearest` instead (imgui's bundled examples use `Linear`).

The text seems to render fine on the gfx renderer.


Gekkio:
Thanks!

PR #95, opened 2017-12-13
Use $crate instead of ::imgui.

shawnscode:
This provides a simple workaround when name collision happens.


Gekkio:
Thanks! :smile: Merging

PR #102, opened 2018-02-04
Custom drawing

JoNil:
Hi!

This is an initial attempt at making a safe version of the imgui custom drawing api.

I submit this mostly to get some early feedback on the design. It should probably go through some iterations and review before you pull it in so we can make sure it is safe!

I'm not certain if i actually need the PhantomData in the struct, since i only allow a refernece to it in the callback that should ensure that we can only have a reference when it is actually safe to have one.

And the *mut should ensure the struct is not Send as well :)

I may also attempt to add the path drawing api depending on if i need it for my application. 

Cheers!

### DIFF
--- a/imgui-sys/src/lib.rs
+++ b/imgui-sys/src/lib.rs
@@ -1920,7 +1920,7 @@ extern "C" {
         uv_d: ImVec2,
         col: ImU32,
     );
-    pub fn ImDrawList_AddPolyLine(
+    pub fn ImDrawList_AddPolyline(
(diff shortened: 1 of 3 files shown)


Gekkio:
Thanks for doing this! 😃 

I'm fairly busy at the moment, but I'll try to find time to look at this later this week.

Gekkio:
Custom drawing is now available in master, so I'll close this. Please open an issue or a pull request if something is missing or broken. Thanks! :smile:

PR #109, opened 2018-03-26
Allow mouse cursor to be updated

malikolivier:
This PR updates the mouse cursor in the examples.
The necessary helper functions are added into the lib.

Here is the result after the merge request. This is an example when a window is resized:
Cursor is update while resizing a window

Prior to this PR, the mouse cursor would have been the default Arrow.

### DIFF
--- a/imgui-examples/examples/support/mod.rs
+++ b/imgui-examples/examples/support/mod.rs
@@ -1,4 +1,4 @@
-use imgui::{ImGui, Ui};
+use imgui::{ImGui, ImGuiMouseCursor, Ui};
 use std::time::Instant;
 
 #[derive(Copy, Clone, PartialEq, Debug, Default)]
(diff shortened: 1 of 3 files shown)


malikolivier:
As a side note, maybe it would be judicious to make a separate crate linking the window handling lib (`winit` here) and Dear ImGui to make the `support` module windowing-lib-agnostic. For `winit`, I would call it something similar to `imgui-winit-binder`. We could then handle all the events/keystrokes/mouse in the same crate.

Gekkio:
Thanks! :smile:

imgui-examples/examples/support/mod.rs
```diff
@@ -105,6 +105,26 @@ pub fn run<F: FnMut(&Ui) -> bool>(title: String, clear_color: [f32; 4], mut run_
...
+
+        let mouse_cursor = imgui.mouse_cursor();
+        if imgui.mouse_draw_cursor() || mouse_cursor == ImGuiMouseCursor::None {
+            // Hide OS cursor
+            gl_window.set_cursor_state(glutin::CursorState::Hide).unwrap();
+        } else {
+            // Set OS cursor
+            gl_window.set_cursor_state(glutin::CursorState::Normal).unwrap();
+            gl_window.set_cursor(match mouse_cursor {
+                ImGuiMouseCursor::None => unreachable!(),
```
Gekkio: Could you add a message here as an argument to `unreachable!`? Something like "Mouse cursor was none"

Rust generates normal panic code for `unreachable!` and doesn't optimize the case away, so if we ever make a mistake here and end up in the unreachable case, it would be nice to have a good error message.

malikolivier:
I added a message to the `unreachable!` macro. Thanks for pointing that out.

PR #110, opened 2018-03-26
Add some helper functions: mouse_delta, get_window_size, is_item_active

malikolivier:
Include 3 helper functions to `imgui` crate:

- `mouse_delta`: Get mouse delta since previous frame
- `get_window_size`: Wraps `sys::igGetWindowSize` to get latest window's size.
- `is_item_active`: Wraps `sys:igIsItemActive`

I would just like to know your opinion regarding the naming of the functions. I tried to use the same convention as the code around each function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -392,6 +392,18 @@ impl<'ui> Ui<'ui> {
         }
         (out.x, out.y)
     }
+    /// Get current window's width in pixels
+    pub fn get_window_width(&self) -> f32 {
+        unsafe {
+            sys::igGetWindowWidth()
+        }
+    }
+    /// Get current window's height in pixels
+    pub fn get_window_height(&self) -> f32 {
+        unsafe {
+            sys::igGetWindowHeight()
+        }
+    }
 }
 
 // Layout
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -190,6 +190,11 @@ impl ImGui {
         io.mouse_pos.x = x;
         io.mouse_pos.y = y;
     }
+    /// Get mouse's position's delta between the current and the last frame.
+    pub fn mouse_delta(&self) -> (f32, f32) {
+        let io = self.io();
+        (io.mouse_delta.x, io.mouse_delta.y)
+    }
     pub fn set_mouse_down(&mut self, states: &[bool; 5]) {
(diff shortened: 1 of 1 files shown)


Gekkio:
Thanks! :smile:

> I would just like to know your opinion regarding the naming of the functions. I tried to use the same convention as the code around each function.

Looks good! It's always OK to follow the naming of the wrapped `imgui-sys` functions, and we can rename things later if some we think of better names.

src/lib.rs
```diff
@@ -248,6 +253,14 @@ impl ImGui {
     pub fn get_time(&self) -> f32 { unsafe { sys::igGetTime() } }
     pub fn get_frame_count(&self) -> i32 { unsafe { sys::igGetFrameCount() } }
     pub fn get_frame_rate(&self) -> f32 { self.io().framerate }
+    /// Get current window's size in pixels
+    pub fn get_window_size(&self) -> (f32, f32) {
```
Gekkio: This function should be part of `Ui` instead `ImGui`, because it should only be called while an imgui frame is being prepared. If it's called at the wrong time, the whole application will immediately abort (not even a panic).

malikolivier: Ok! Now I understand better why methods in Ui and UmGui are separated. Will fix it shortly (and add a few other wrapper functions in the mean time).

Gekkio:
> add a few other wrapper functions in the mean time

Are you still planning to add something, or is this ready for merging? :smiley: 

One minor thing: I don't think `get_window_width` and `get_window_height` are needed, since `get_window_size` already exists and it's so easy to get just one dimension if the other one is not needed.

malikolivier:
> One minor thing: I don't think get_window_width and get_window_height are needed, since get_window_size already exists and it's so easy to get just one dimension if the other one is not needed.

I agree with your reasoning. The only reason I included them was because `get_window_width` and `get_window_height` were already there in imgui.
I will delete them if you wish so.

> Are you still planning to add something, or is this ready for merging?

I have more helper functions in my working branch, I will add them within the day. Thanks!

malikolivier:
Assuming the build passes, this is ready to be merged!

Gekkio:
Merging :smile:

PR #112, opened 2018-04-08
Implement custom fonts.

Lymia:
This implements custom fonts as suggested in #83. It avoids many of the concerns brought up in #84, since that pull request seems to be dead.


Gekkio:
Thanks! Looks very good :smile: 

Three minor things:

* `c_float` is basically always `f32`, so it would make more sense to have `f32` instead of `f64` as the argument type for all float arguments in functions (e.g. `ImFontConfig::size_pixels`)
* Use `Into` and generics to accept `ImVec2` arguments *or* use tuples all the way until the final `imgui-sys` call where `ImVec2` is needed. Using normal tuples (e.g. `(0.0, 0.0)`) is preferable in user code and with `Into` we can convert those to `ImVec2` ergonomically. In the underlying library `ImVec2` seems to exist only because there's no canonical "tuple" type, but in Rust we have tuples so we should prefer using them whenever possible.
* Convert to `(f32, f32)` when returning an `ImVec2` from the underlying library (`ImFont::get_display_offset` in this case). Same reasoning as above.

Lymia:
Done.

Gekkio:
Merging :smile:

PR #114, opened 2018-04-14
Update glium to 0.21

### LINKED ISSUE #98: Examples don't work on OSX
I'm trying to integrate this into a Glium project and am running into issues on OSX. I thought the issue could be with the way I'm integrating it so I tried running the examples and am having issues.
The hello_world example just gives a blank white screen.
The test_window example renders a clickable menu but the mouse cursor doesnt line up with where Imgui thinks the cursor is (you have to click about half a window above where the item is).
It also looks like the alignment is off and part of the menu gets chopped off.
screen shot 2018-01-12 at 3 31 01 pm

I'm on OSX 10.13.2 with rust nightly.

ebarnard:
Closes #98

### DIFF
--- a/imgui-examples/Cargo.toml
+++ b/imgui-examples/Cargo.toml
@@ -11,7 +11,7 @@ publish = false
 [dev-dependencies]
 gfx = "0.16"
 gfx_window_glutin = "0.19"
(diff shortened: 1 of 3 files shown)


Gekkio:
Thanks! :smile:

PR #116, opened 2018-04-24
Add drag widgets

michaelfairley:


Gekkio:
Thanks! :smile: 

Looks pretty straightforward...I'm a bit allergic to using macros to copy-paste code, but in this case it seems fairly reasonable, and we can abstract things later in a different way if that seems appropriate.

PR #122, opened 2018-05-03
Custom texture support for glium back-end

### LINKED ISSUE #65: Using custom textures
I was attempting to load a texture from my application into an imgui window using something similar to the following code:

```
ui.window(im_str!("PPU"))
    .size((150.0, 130.0), ImGuiSetCond_FirstUseEver)
    .resizable(true)
    .build(|| {
        unsafe {
            imgui_sys::igImage(ppu.get_tex_id(), ImVec2::new(160.0, 144.0), ImVec2::new(0.0, 0.0), ImVec2::new(1.0, 1.0), ImVec4::new(0.0, 0.0, 0.0, 1.0), ImVec4::new(1.0, 0.0, 0.0, 1.0));
        }
    });
```

But it seems custom textures are disallowed by the imgui-rs glium renderer: https://github.com/Gekkio/imgui-rs/blob/master/imgui-glium-renderer/src/lib.rs#L108

I'm wondering what needs to be done for this to get supported?

malikolivier:
Fixes #65.

Implement custom texture support. Each commit message thoroughly explain the content of the commit.

Before merging, I want to know the maintainer(s) opinion about the design choices of this PR.
It is designed with memory safety and back-end agnosticity in mind.
Suggestions and improvements are welcome.

Shortcomings of the design choices are explained in the very first commit of the PR: 3013e03.
Please see the new example to get an idea how the custom texture API is used.

- Some rework on the `support` module of the example is necessary, as creating custom textures in `glium` requires access to the gl context. In this PR, I created a `support_custom_textures` module that allows to get the gl context (c58eb9c). I would like to include this `support_custom_textures` module and its AppContext/AppConfig types into `imgui_glium_renderer` and rewrite all the examples. Any objection?
- I would like to add safe wrappers to `igImageButton` and `ImDrawList_AddImage` once the design is validated. 
- My current knowledge of `gfx` is abysmal. I need to study a bit more `gfx` before I can write the part of the gfx backend for custom texture support. Moreover, it seems that the `gfx` crate is about to be deprecated as `gfx` is currently being rewritten from scratch [1], so I would rather directly implement a back-end using the master branch of `gfx` as base. What do you think we should do? Can we merge with gfx support first?

[1] https://github.com/gfx-rs/gfx/#pre-ll

EDIT: Font textutre is no longer treated as a separate case in 1de593a

### DIFF
--- a/imgui-glium-renderer/src/lib.rs
+++ b/imgui-glium-renderer/src/lib.rs
@@ -11,9 +11,10 @@ use glium::program;
 use glium::index::{self, PrimitiveType};
 use glium::texture;
 use glium::vertex;
-use imgui::{DrawList, ImDrawIdx, ImDrawVert, ImGui, Ui};
+use imgui::{DrawList, FromImTexture, ImDrawIdx, ImDrawVert, ImGui, Ui};
 use std::borrow::Cow;
 use std::fmt;
+use std::ops::Deref;
 use std::rc::Rc;
 
 pub type RendererResult<T> = Result<T, RendererError>;
@@ -118,10 +119,13 @@ impl Renderer {
 
         let mut idx_start = 0;
         for cmd in draw_list.cmd_buffer {
-            // We don't support custom textures...yet!
-            assert!(cmd.texture_id as usize == font_texture_id);
-
             let idx_end = idx_start + cmd.elem_count as usize;
+
+            let texture = if cmd.texture_id as usize == font_texture_id {
+                &self.device_objects.texture
+            } else {
+                <Texture as FromImTexture>::from_id(cmd.texture_id).deref()
+            };
 
             try!(
                 surface.draw(
@@ -133,7 +137,7 @@ impl Renderer {
                     &self.device_objects.program,
                     &uniform! {
                           matrix: matrix,
-                          tex: self.device_objects.texture.sampled()
+                          tex: texture.sampled()
                               .magnify_filter(MagnifySamplerFilter::Nearest)
                               .minify_filter(MinifySamplerFilter::Nearest),
                       },
--- a/imgui-glium-renderer/src/lib.rs
+++ b/imgui-glium-renderer/src/lib.rs
@@ -5,7 +5,7 @@ extern crate imgui;
 mod im_texture;
 pub use im_texture::Texture;
 
-use glium::{DrawError, GlObject, IndexBuffer, Program, Surface, Texture2d, VertexBuffer};
+use glium::{DrawError, IndexBuffer, Program, Surface, Texture2d, VertexBuffer};
 use glium::backend::{Context, Facade};
 use glium::program;
 use glium::index::{self, PrimitiveType};
@@ -14,7 +14,6 @@ use glium::vertex;
 use imgui::{DrawList, FromImTexture, ImDrawIdx, ImDrawVert, ImGui, Ui};
 use std::borrow::Cow;
 use std::fmt;
-use std::ops::Deref;
 use std::rc::Rc;
 
 pub type RendererResult<T> = Result<T, RendererError>;
@@ -115,17 +114,12 @@ impl Renderer {
             [0.0, 0.0, -1.0, 0.0],
             [-1.0, 1.0, 0.0, 1.0],
         ];
-        let font_texture_id = self.device_objects.texture.get_id() as usize;
 
         let mut idx_start = 0;
         for cmd in draw_list.cmd_buffer {
             let idx_end = idx_start + cmd.elem_count as usize;
 
-            let texture = if cmd.texture_id as usize == font_texture_id {
-                &self.device_objects.texture
-            } else {
-                <Texture as FromImTexture>::from_id(cmd.texture_id).deref()
-            };
+            let texture = <Texture as FromImTexture>::from_id(cmd.texture_id);
 
             try!(
                 surface.draw(
@@ -165,7 +159,6 @@ pub struct DeviceObjects {
     vertex_buffer: VertexBuffer<ImDrawVert>,
     index_buffer: IndexBuffer<ImDrawIdx>,
     program: Program,
-    texture: Texture2d,
 }
 
 fn compile_default_program<F: Facade>(
@@ -213,7 +206,7 @@ impl DeviceObjects {
         ));
 
         let program = try!(compile_default_program(ctx));
-        let texture = try!(im_gui.prepare_texture(|handle| {
+        let texture = try!(im_gui.register_font_texture(|handle| {
             let data = RawImage2d {
                 data: Cow::Borrowed(handle.pixels),
                 width: handle.width,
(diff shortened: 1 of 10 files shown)


malikolivier:
I hacked support for gfx in a working branch (see link below), however, performance is very, very bad.
With the current design, `FromImTexture::from_id` returns a reference.
However, the pipe object cannot use a reference to the `TextureSampler`. It internally uses a `ShaderResourceView`, which holds an `Arc` to the underlying texture object.
I need to clone the reference to `ShaderResourceView` for each `draw_list` with the current design.
This is SLOW (for each rendering loop, you use 2x or 3x more CPU with this patch).

I want to find another way. But I don't have much idea for now.

https://github.com/malikolivier/imgui-rs/blob/custom-textures-gfx-draft/imgui-gfx-renderer/src/lib.rs
https://github.com/malikolivier/imgui-rs/blob/custom-textures-gfx-draft/imgui-gfx-renderer/src/im_texture.rs

src/texture.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn new<T: 'static + ImTexture>(texture: T) -> Self { AnyTexture(Rc::new(Box::new(texture))) }
+}
+
+impl ImTexture for AnyTexture {
+    fn get_id(&self) -> ImTextureID { self.deref().get_id() }
+    fn get_size(&self) -> (u32, u32) { self.deref().get_size() }
+}
+
+/// Allows to directly use the methods implemnted on [`AnyTexture`]
+impl Deref for AnyTexture {
```
Gekkio: Are you sure this is necessary? You've implemented `ImTexture` above anyway for `AnyTexture`, so the deref doesn't seem to add any value here

src/texture.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// Query texture size, in pixels (for convenience)
+    fn get_size(&self) -> (u32, u32);
+}
+
+/// We implement [`ImTexture`] for [`ImTextureID`] one only to be able to use
+/// it on the ID of the font texture (as is).
+impl ImTexture for ImTextureID {
+    fn get_id(&self) -> ImTextureID { *self }
+    fn get_size(&self) -> (u32, u32) {
+        panic!("Cannot get size of ImTextureID (it is a raw pointer)!")
```
Gekkio: A convenience function that panics sounds like a bad idea. Do we really need `get_size` in the trait if we can't implement it for the font texture?

src/texture.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/// Trait defining how an object implementing [`ImTexture`] should be converted
+/// back to the native texture type used by the back-end.
+pub trait FromImTexture {
+    fn from_im_texture<T: ImTexture>(texture: &T) -> &Self {
+        let texture = texture.get_id();
+        Self::from_id(texture)
+    }
+    fn from_id<'a>(texture_id: ImTextureID) -> &'a Self;
```
Gekkio: This is very dangerous, because the signature tells that we can conjure from any `*const c_void` a valid reference with any lifetime to anything that has a `FromImTexture` implementation.

src/texture.rs
```diff
@@ -0,0 +1,109 @@
...
+pub struct TextureCache(RefCell<BTreeMap<u64, AnyTexture>>);
+
+impl TextureCache {
+    pub fn new() -> Self { TextureCache(RefCell::new(BTreeMap::new())) }
+
+    /// Register the texture inside the cache.
+    ///
+    /// Swap and return some texture if another texture with the same name was
+    /// already registered.
+    pub fn register_texture<T>(&self, name: &ImStr, texture: T) -> Option<AnyTexture>
```
Gekkio: Use `ImStr` only when it needs to be passed to the underlying C library. `&str` is always more ergonomic if we're only doing stuff in Rust code.

src/lib.rs
```diff
@@ -1359,3 +1433,90 @@ impl<'ui> Ui<'ui> {
...
+        T: IntoImTexture<U>,
+        U: 'static + ImTexture,
+    {
+        let imgui = self.imgui();
+        if let Some(texture) = imgui.textures.get_texture(name) {
+            texture
+        } else {
+            let texture = f().into_texture();
+            imgui.textures.register_texture(name, texture);
+            imgui.textures.get_texture(name).unwrap()
```
Gekkio: `unwrap` :cry:

src/lib.rs
```diff
@@ -1359,3 +1433,90 @@ impl<'ui> Ui<'ui> {
...
+    /// ```
+    pub fn replace_texture<T, U>(&self, name: &ImStr, t: T) -> AnyTexture
+    where
+        T: IntoImTexture<U>,
+        U: 'static + ImTexture,
+    {
+        let imgui = self.imgui();
+        let texture = t.into_texture();
+        imgui.textures.register_texture(name, texture);
+        imgui.textures.get_texture(name).unwrap()
```
Gekkio: `unwrap` :cry:

imgui-glium-renderer/src/lib.rs
```diff
@@ -111,15 +114,13 @@ impl Renderer {
...
-        let font_texture_id = self.device_objects.texture.get_id() as usize;
 
         let mut idx_start = 0;
         for cmd in draw_list.cmd_buffer {
-            // We don't support custom textures...yet!
-            assert!(cmd.texture_id as usize == font_texture_id);
-
             let idx_end = idx_start + cmd.elem_count as usize;
 
+            let texture = <Texture as FromImTexture>::from_id(cmd.texture_id);
```
Gekkio: This is very unsafe, because we're assuming the texture_id always corresponds to a glium texture, but there's nothing in the main imgui crate functions that guarantees it. This is the main difficulty in custom textures: they are used in renderers but added/declared/registered somewhere else.

malikolivier:
Thanks for your feedback!

> Also, texture registration takes ownership of the texture, so if you want to have a changing texture, it needs to be allocated on every frame. This doesn't seem right to me.

Not necessarily, at least with the glium backend you can do this:

```rust
let draw_texture = ui.make_texture(im_str!("#Draw"), || {
                    Texture2d::empty(&gl_ctx, 100, 100).unwrap()
                });
                // Get the texture as a surface. It must first be converted to a
                // `glium::Texture2d` object by using `Texture::from`.
let mut surface = Texture::from_im_texture(&draw_texture).as_surface();
// You can now draw on the surface.
```

From example: https://github.com/malikolivier/imgui-rs/blob/c58eb9c9dce6d72ca401c17e9d7f20a9dd0f8c13/imgui-examples/examples/custom_textures.rs#L77

I don't know how to replicate that with the gfx backend.

> Right now there's quite a lot of memory safety issues, so we need to think about the design a bit more.

Yes, I agree. I will write some idea in an inline comment.

src/texture.rs
malikolivier: You are right. This is useless! Will get rid of it in the next iteration.

src/texture.rs
malikolivier: I we ever were to accept this kind of design, should the trait be marked `unsafe`?

imgui-glium-renderer/src/lib.rs
malikolivier: Yes. It is also unsafe if we move the textures within `imgui`. The the textures are moved, we will get a dangling pointer.

To ensure that texture_id always corresponds to the same type of texture, an idea of mine was to use generics on the ImGui instance as below. But that would break all existing code. That's why I decided to go with a type `AnyTexture` encapsulating a trait object.

```rust
/// `T` is the texture type expected by the backend.
pub struct ImGui<T> {
    ini_filename: Option<ImString>,
    log_filename: Option<ImString>,
    textures: TextureCache<T>,
}
```

PR #123, opened 2018-05-13
Add support for other PushID overloads

michaelfairley:
Or, a different approach: https://github.com/michaelfairley/imgui-rs/tree/push_id_overloads2

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,7 @@ pub extern crate imgui_sys as sys;
 
 use std::ffi::CStr;
 use std::mem;
-use std::os::raw::{c_char, c_float, c_int, c_uchar};
+use std::os::raw::{c_char, c_float, c_int, c_uchar, c_void};
 use std::ptr;
 use std::slice;
 use std::str;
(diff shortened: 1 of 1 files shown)


Gekkio:
Thanks!

I think I prefer the trait version in theory, but there's some problems with it:

* The only way to implement it correctly is to use unsafe code -> pretty much makes the trait pointless from an implementation point of view
* It violates all lifetime and initialization guarantees of `imgui-rs`. I can call `"lol".push()` at any time with very bad consequences

What if we made an enum instead and supported converting to it? Something like

```
pub enum ImId<'a> {
  Int(i32), Str(&'a str), ImStr(&'a ImStr), Ptr(*const c_void)
}
```

Then you could:

* implement `Into<ImId<'static>> for i32` and others
* accept `<I: Into<ImId<'a>>` in `push_id` and `with_id`
* match based on the enum and call the right push function in push_id -> dangerous pushing is limited to one place where we already do it and have certain guarantees

So, in other words, right now the polymorphism in the API is all about pushing (= a side-effect), but it could be better if it was just about knowing which of the supported id types is being used.

Also, there are some conflicts that need to be resolved.

michaelfairley:
Great points. This resulted in a complete rewrite, so I've opened #133 to have cleaner history.

PR #125, opened 2018-05-23
window_draw_list: Add support for drawing text

malikolivier:


philipaconrad:
This would be a stupendously useful feature to have for a custom plotting widget I'm writing right now.

src/window_draw_list.rs
```diff
@@ -209,6 +209,19 @@ impl<'ui> WindowDrawList<'ui> {
...
     }
 
+    /// Draw a text whose upper-left corner is at point `pos`.
+    pub fn add_text<P, C>(&self, pos: P, col: C, text: &str)
+    where
+        P: Into<ImVec2>,
+        C: Into<ImColor>,
+    {
+        unsafe {
+            let start = text.as_ptr() as *const i8;
```
Gekkio: Please use `*const c_char`. `c_char` is not always `i8` (on ARM for example it's `u8` so this won't compile)

malikolivier: Thx! Pushed the fix on 9b0804f

malikolivier:
I added another commit, 485c8a2, to allow the use of any string-like type with `AsRef<str>`.

PR #126, opened 2018-05-23
window_draw_list: Add #[must_use] to relevant types

malikolivier:
It is easy to forget to call .build() on these temporary types.

To avoid this kind of error, we emit a compiler warning when .build() is
forgotten.


Gekkio:
Thanks!

PR #132, opened 2018-05-31
Add want_capture_mouse and want_capture_keyboard to ImGui (fixes #128)

### LINKED ISSUE #128: Correct way to access want_capture_mouse and friends
I really like this library and imgui. Thanks for writing this wrapper!

I read the FAQ in dear imgui and it mentions using want_capture_mouse etc.. for determining if input should be sent to imgui or the game, however those are only members of Ui which only has an immutable reference to ImGui.

What this means (I think) is you can't send input to ImGui then take that same event and use want_capture_mouse to try to decide if you are going to pass it on to the games code on the fly.

I can see a few workarounds here like buffering events or using the last frame's values for want_* but I am curious if there is a "right" way to do it.

Maybe we could update the examples to have this in them?

thomcc:
These should be on the ImGui object so that you can update state separately from drawing the Ui (there are also lifetime annoyances that are made much worse by them only being on Ui).

I left the versions on Ui (both for backwards compatibility and because I don't see why they shouldn't be there), but they just call the versions on ImGui now.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -314,6 +314,8 @@ impl ImGui {
     pub fn get_time(&self) -> f32 { unsafe { sys::igGetTime() } }
     pub fn get_frame_count(&self) -> i32 { unsafe { sys::igGetFrameCount() } }
     pub fn get_frame_rate(&self) -> f32 { self.io().framerate }
+    pub fn want_capture_mouse(&self) -> bool { self.io().want_capture_mouse }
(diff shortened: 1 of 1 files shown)


Gekkio:
Closing because 0.1 exposes the Io struct and these fields, so getters are no longer needed and the original use case is now possible (updating state before starting a frame and getting an `Ui` instance)

PR #138, opened 2018-06-08
Allow setting flags on tree nodes

michaelfairley:
[Relevant imgui.h enum](https://github.com/ocornut/imgui/blob/8e48ab6b19d153c2ff9b56379ff1fb60034ca133/imgui.h#L638)


malikolivier:
Thanks!

Issue #115, opened 2018-04-16
Regression on updating glium to 0.21


malikolivier:
I get a panic when I run an example, to be specific the program panics when I close the window.
Here is the backtrace:

```
imgui-rs/imgui-examples$  cargo run --example test_window_impl
    Updating registry `https://github.com/rust-lang/crates.io-index`
   Compiling imgui-sys v0.0.19-pre (file:///home/malik/workspace/lab/aflak-backend/node_editor/imgui-rs/imgui-sys)
warning: third-party/cimgui/imgui/imgui.cpp: In function ‘void ImGui::RenderTriangle(ImVec2, ImGuiDir, float)’:
warning: third-party/cimgui/imgui/imgui.cpp:3136:11: warning: this statement may fall through [-Wimplicit-fallthrough=]
warning:          r = -r; // ...fall through, no break!
warning:          ~~^~~~
warning: third-party/cimgui/imgui/imgui.cpp:3137:5: note: here
warning:      case ImGuiDir_Down:
warning:      ^~~~
warning: third-party/cimgui/imgui/imgui.cpp:3144:11: warning: this statement may fall through [-Wimplicit-fallthrough=]
warning:          r = -r; // ...fall through, no break!
warning:          ~~^~~~
warning: third-party/cimgui/imgui/imgui.cpp:3145:5: note: here
warning:      case ImGuiDir_Right:
warning:      ^~~~
   Compiling imgui v0.0.19-pre (file:///home/malik/workspace/lab/aflak-backend/node_editor/imgui-rs)
   Compiling imgui-glium-renderer v0.0.19-pre (file:///home/malik/workspace/lab/aflak-backend/node_editor/imgui-rs/imgui-glium-renderer)
   Compiling imgui-gfx-renderer v0.0.19-pre (file:///home/malik/workspace/lab/aflak-backend/node_editor/imgui-rs/imgui-gfx-renderer)
   Compiling imgui-examples v0.0.19-pre (file:///home/malik/workspace/lab/aflak-backend/node_editor/imgui-rs/imgui-examples)
    Finished dev [unoptimized + debuginfo] target(s) in 7.97 secs
     Running `/home/malik/workspace/lab/aflak-backend/node_editor/imgui-rs/target/debug/examples/test_window_impl`
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', libcore/option.rs:335:21
note: Run with `RUST_BACKTRACE=1` for a backtrace.
[...]

PR #139, opened 2018-07-19
Move to ImGui instead of cimgui, use bindgen

semtexzv:
I needed features introduced in 1.61. So I decided to rewrite imgui-sys using bindgen.
Everything seems to work. 

One drawback is that oldest supported rust version will need to be bumped.


malikolivier:
Thanks. That's a very good idea. I will be testing to see if nothing breaks.

First, this merge request contains at least five different contributions:
1. Updating of glium/glutin. That's great as it fixed #115 
2. Feature flag for toggling glutin to glium renderer
   - Could you explain the use cases behind these flags?
3. Fix the scaling in the examples
   - Could you be more specific regarding what was the expected behaviour and what was the issue, and how it was fixed?
4. Replace cimgui with imgui using bindgen
   - That's awesome. I will need to check that nothing breaks on all platform before merging.
5. Numerous aesthetic changes.
   - These changes are good. But they introduce a lot of noise in the changelogs, making it very hard to review. I would take them as a separate merge request.

Among those changes, a new dependency `freetype` was included. Could you explain why it was included? (Sorry if I am too dumb to understand by myself xD)

> One drawback is that oldest supported rust version will need to be bumped.

What would be the oldest supported stable version after your MR?
Could it be [1.21.0](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1210-2017-10-12). It seems the only test failures come from the implementation of Clone.

Overall, could you split each of the contributions I mentioned above? We can merge the more straightforward ones first, then review the most complicated ones.

I think we need @Gekkio's opinion as well regarding migration to `bindgen`.

Gekkio:
Yeah this is quite a big pull request with a lot of changes, and it would be easier to merge the main changes individually. Here are some comments:

1. Experimental freetype rendering

This is very interesting, but I think it's way too experimental for imgui-rs master, especially if it's enabled by default. Even in upstream it's labeled as "Currently not optimal and probably has some limitations or bugs"

2. bindgen

I've thought in the past about using bindgen (with cimgui), but there's a lot of pros and cons to consider and for me the pros don't feel bigger than the cons. The main question is: why use bindgen?

I can quickly come up with these points:

Pros:

* a bit easier to maintain (imgui version upgrades still need manual work though! especially if we ever want to go 1.x.x)
* minimizes human errors

Cons:

* breaks pretty much all existing code (because many imgui-sys types are used in safe code too)
* user code becomes a mix of idiomatic and non-idiomatic code (because imgui-sys is used in safe code too)
* using bindgen with build.rs makes compiling much more complicated (e.g. LLVM is required on Windows. Do the bindings work with both msvc and gnu ABIs?)
* LLVM 3.9 requirement means imgui-rs won't work out-of-the-box on Debian stable or old Ubuntu 16.04 LTS
* a lot of magic in build.rs
* bindgen has some limitations with C++. Are we sure the bindings work on all platforms and continue to work without causing a maintenance nightmare? For example, inline functions are lost since bindgen can't handle them. There's not a lot of those in imgui, but this pull request loses some functions from the imgui-rs API because of this

photex:
Just wanted to drop a note of support of using bindgen if it generates a suitably useful sys module.
What is especially interesting to me here is that this could open the door to an extension such as ImGuizmo as well.

Gekkio:
Closing because bindgen is used since 0.1 for cimgui/imgui upgrades (but not at build time so LLVM is not a build-time dependency) and this pull request is no longer valid.

PR #141, opened 2018-07-30
Update glium to 0.22

shockham:


shockham:
Seems the following won't compile using rust 1.20.0:
- on linux x11-dl v2.18.1 due to _destructuring a reference_
- on macOS winit v0.16.2 due to _const fns being an unstable feature_
Both of them not being direct dependencies but for glutin 0.17, so not sure on the best way to sort this.

Gekkio:
It's unfortunate, but I think we'll just have to bump the minimum Rust version a bit. It's not the first time :wink: 

It looks like we'll have to try 1.24...? Could you bump the version in README and travis.yml in this pull request and we'll see what Travis has to say after that?

Makes me sad though...I like new language features as much as anybody else, but it's a bit weird that common Rust libraries require a very fresh compiler (<6 months since release).

shockham:
Looks good with 1.24.

Think the x11-dl error was fixed in v2.18.2 (erlepereira/x11-rs@74f3eadb), so it is all for one line in winit using a const fn on a single platform too.

I haven't looked but if there aren't any feels like there should be guidelines for lib ecosystem to support a min version.

Gekkio:
Thanks! :smile: 

It looks like winit doesn't crash on exit anymore, so I think that release blocker is finally removed

PR #143, opened 2018-08-10
[examples] support: Fix scale bug

malikolivier:
The new version of glutin apparently changed the behaviour regarding
HDIPI.
This PR is partly inspired from the patch submitted by @semtexzv #139.

However, the scaling issue is not completely fixed yet, everything appears way enlarged on my system.
(below is a screenshot taken from a 3200-pixel-wide screen)

image

I want the examples in imgui-rs to have the same behaviour as before. @Gekkio  Any thought?

EDIT: After reading glutin's documentation [about dpi](https://docs.rs/glutin/0.18.0/glutin/dpi/index.html), I was able to get the behaviour we had before updating glutin by setting the environment variable WINIT_HIDPI_FACTOR to 1 (my system uses x11).

```sh
WINIT_HIDPI_FACTOR=1 cargo run --example test_window_impl
```

I am not sure what should be the correct thing to do for imgui-rs...

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -41,6 +41,9 @@
   Keep redirection.
   - Rename `ImGuiCol::ChildWindowBg` to `ImGuiCol::ChildBg`.
   Keep redirection.
+- Upgrade glium to 0.22.0. This updates winit to 0.16. This changes the way
+HIDPI are calculated. Depending on your needs, you may want to set HIDPI to 1
+by setting the environment variable `WINIT_HIDPI_FACTOR=1` if you use X11.
 
 ### Deprecated
 
--- a/imgui-examples/examples/support/mod.rs
+++ b/imgui-examples/examples/support/mod.rs
@@ -129,14 +129,11 @@ pub fn run<F: FnMut(&Ui) -> bool>(title: String, clear_color: [f32; 4], mut run_
             });
         }
 
-        let size_pixels = gl_window.get_inner_size().unwrap();
+        let size_points = gl_window.get_inner_size().unwrap();
         let hdipi = gl_window.get_hidpi_factor();
-        let size_points = (
-            (size_pixels.width as f64 / hdipi) as u32,
-            (size_pixels.height as f64 / hdipi) as u32,
-        );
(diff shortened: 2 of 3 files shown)


malikolivier:
> The window might feel way too big, but I think it's actually working as intended.

That's what I feared. That's actually the expected behaviour. The app I am building is completely broken when the DPI is not 1.0 ;(

I will include the `WINIT_HIDPI_FACTOR=1` hack in the README, and I think we can merge this. I am a bit sorry for people downstream. I guess it was my fault for ignoring DPI until now.

By the way, any way to fiddle with the scaling from imgui then?
I really looks ugly on my screen. The fonts are pixelated and enormous.

Gekkio:
> By the way, any way to fiddle with the scaling from imgui then?

Yeah there's some workarounds reported in upstream github issues. IIRC the application has to fiddle with font oversampling and global scaling settings to get nice results.

malikolivier:
Indeed, there seems to be some work in progress upstream.

I pushed some notes in the CHANGELOG.

Please let me know if it's OK as is to me merged.

For future reverence, the HIDPI factor can be work around on X11 by setting WINIT_HIDPI_FACTOR to 1.

This can be done as follows:

```sh
WINIT_HIDPI_FACTOR=1 cargo run --example test_window_impl
```

You could even set the environment variable at runtime:

```rust
use std::env;

fn main() {
    env::set_var("WINIT_HIDPI_FACTOR", "1");
    /* ... Continue your program as usual ... */
}
```

Gekkio:
Yeah looks good to me!
I'll do a release this weekend if no other regressions appear

Gekkio:
@malikolivier , could you test the latest master version on your machine? I've added several HiDPI fixes and basic font scaling, which should improve things a bit and it's probably the best we can do right now.

Unfortunately scaling with non-integer factors will always be a bit blurry...It seems that on X11 winit can give non-integer scaling factors (I get 1.1666666666666667 on my 1440p screen).

malikolivier:
@Gekkio Thanks! That's way better (especially for the Japanese font).

Before: 
image

After:
image

PR #149, opened 2018-08-25
Alternate Implementation of custom texture support for glium back-end

Gopiandcode:
The other thread  #122  on Custom texture support #65 

hasn't been active in a while, and I really needed custom texture support, so I've provided an alternate implementation based on the discussion on that thread. Many thanks to @malikolivier for his implementation which provided a handy starting point.

Currently the custom texture support is only implemented for `glium`, however, I believe the implementation is sufficiently generalized such that it would be easy to extend it to `gfx` - if the implementation looks good, I can get to work on extending it to `gfx`.

### Core Logic
The core logic of the implementation is encapsulated within the `TextureCache<T>` struct, located within `src/texture_cache.rs`.

This struct provides a mapping from `String -> usize -> T`, with the intention being that once a  texture type `T` is bound to a `String`, subsequent calls will pass the `usize` to `imgui` to manage, and then at the point of rendering, the `usize`'s can be mapped back into `T`. This avoids the issue found in the alternate implementation where an unsafe mapping had to be made from `ImTextureID` to `T`.

### Integration with existing api
Next, from looking at both the `glium` and `gfx` Texture structs, it seems that they are too different on a conceptual level, to use a single trait to represent them - this means that trait objects won't be idiomatic to implement.

Now, due to the fact that a copy of `Ui<'a>` is stored globally within `CURRENT_UI`, it is not possible to parameterize the `ImGui` or `Ui` structs with the type of the texture T as const generics aren't supported.

So to work around these limitations, in the implementation of the `imgui_glium_textured_renderer::Renderer`, I provide a wrapper around `Ui<'a>`,  `TexturedUi<'a,'b,T>`, which holds a reference to the the Renderer's texture cache, and thereby imbuing the `Ui<'a>` struct with capabilities to make and remove textures.

###  Example
I have also provided an example (once again based off the ones provided by @malikolivier) - the API for using textures is also based off his implementation:

As a quick reference, using api would look like:
```
if !init {
    ui.make_texture("#textureA".into(), /* Make a texture */);
}

ui.window().build(|| {
   ui.image("#textureA").build(); // will return an error if textureA is not defined.
});

``` 

textures

### DIFF
--- a/imgui-gfx-renderer/src/lib.rs
+++ b/imgui-gfx-renderer/src/lib.rs
@@ -120,7 +120,7 @@ impl<R: Resources> Renderer<R> {
             gfx::memory::Usage::Dynamic,
             Bind::empty(),
         )?;
-        let (_, texture) = imgui.prepare_texture(|handle| {
+        let (_, texture) = imgui.prepare_font_texture(|handle| {
             factory.create_texture_immutable_u8::<gfx::format::Rgba8>(
                 gfx::texture::Kind::D2(
                     handle.width as u16,
@@ -215,6 +215,8 @@ impl<R: Resources> Renderer<R> {
         self.bundle.slice.start = 0;
         for cmd in draw_list.cmd_buffer {
             // TODO: check cmd.texture_id
+            // TODOK: here we can change the texture based on
+            // self.bundle.data.texture = imgui.retrieve_texture(cmd.texture_id)?;
 
             self.bundle.slice.end = self.bundle.slice.start + cmd.elem_count;
             self.bundle.data.scissor = Rect {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,4 +16,4 @@ travis-ci = { repository = "Gekkio/imgui-rs" }
 imgui-sys = { version = "0.0.21-pre", path = "imgui-sys" }
 
 [workspace]
-members = ["imgui-examples", "imgui-sys", "imgui-gfx-renderer", "imgui-glium-renderer", "imgui-glutin-support"]
+members = ["imgui-examples", "imgui-sys", "imgui-gfx-renderer", "imgui-glium-renderer", "imgui-glium-textured-renderer", "imgui-glutin-support"]
--- a/imgui-examples/Cargo.toml
+++ b/imgui-examples/Cargo.toml
@@ -16,3 +16,4 @@ glutin = "0.18"
 imgui = { version = "0.0.21-pre", path = "../" }
 imgui-gfx-renderer = { version = "0.0.21-pre", path = "../imgui-gfx-renderer" }
 imgui-glium-renderer = { version = "0.0.21-pre", path = "../imgui-glium-renderer" }
+imgui-glium-textured-renderer = { version = "0.0.21-pre", path = "../imgui-glium-textured-renderer" }
--- a/imgui-examples/examples/custom_textures.rs
+++ b/imgui-examples/examples/custom_textures.rs
@@ -1,0 +1,93 @@
+#[macro_use]
+extern crate glium;
+#[macro_use]
+extern crate imgui;
+extern crate imgui_glium_textured_renderer;
+
+use glium::backend::Context;
+use glium::Texture2d;
+use imgui::*;
+use std::rc::Rc;
+mod support_texture;
+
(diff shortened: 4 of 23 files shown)


imgui-gfx-renderer/src/lib.rs
```diff
@@ -215,6 +215,8 @@ impl<R: Resources> Renderer<R> {
         self.bundle.slice.start = 0;
         for cmd in draw_list.cmd_buffer {
             // TODO: check cmd.texture_id
+            // TODOK: here we can change the texture based on
+            // self.bundle.data.texture = imgui.retrieve_texture(cmd.texture_id)?;
```
malikolivier: Last time I try, you could do this and it worked:

```rust
self.bundle.data.tex.0 = imgui.retrieve_texture(cmd.texture_id)?
                .clone();
```

But it would really slow down the program, so I would not recommend it.

malikolivier:
Sorry for the long wait before the review!

Thank you very much for submitting this PR.

I get how it works, and I'm willing to accept this PR with @Gekkio's approval.

Pros:
- No `unsafe` code
- Complete separation between code to manage custom texture (`TexturedUi` and `imgui-glium-textured-renderer`) and the previous code. People who wants to migrate to the new API using custom textures will be able to do so, without breaking the code of people who do not care about custome textures. However, it may confuse newcomers in the long run.

Drawbacks
- On replacing texture, the texture needs to be re-allocated on each frame (my PR had the same issue)
- font_id being used as a special case
- weak gfx support

Nitpicks:
- Could you fix all compiler warnings when running `cargo build --all`?
- Could you fix formatting with `cargo fmt` (using `rustfmt 0.8.2-stable`)?
- Typo: colissions -> colisions

SpaceManiac:
I'm still working on implementing it, but I want to propose another possible API, which may be simpler to implement and more `gfx`-friendly (I am currently using the gfx backend and need this feature):

* No string-keyed `TextureCache` or separate `TexteredUi`
* The texture ID -> underlying texture mapping is maintained by the renderer, which accepts a renderer-specific texture object and produces a `TextureID`
  * Alternatively, it may produce something like an `ImTexture` struct (not a trait) which consists of a `TextureID` and size
* `Ui::image` and the `Image` struct work as in this PR, but accept a `TextureID` (or `ImTexture`) instead of a `&str`

I don't agree that complete separation between previous API and "with textures" API is a pro; it seems like it doubles the amount of copy-paste code (the glium-textured renderer is mostly just the glium renderer, and the textured support is mostly just the support).

Basically, whatever function is responsible for preparing the textures which will appear in the UI should accept an `&mut` to the renderer it knows how to produce textures for.

Gekkio:
This pull request doesn't need unsafe code, so it's a huge improvement, but the huge amount of copy-paste and the extra steps needed to have texturing shouldn't really be necessary. After seeing two alternative pull requests for this feature, I'm starting to imagine a third approach which sounds pretty much like what SpaceManiac is suggesting.

I can imagine something like this:

```
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
pub struct TextureId(pub usize)
```

Image accepts TextureId and doesn't need to be parameterized (e.g. no Image<E>)

Renderer implementations (note that T is not a generic parameter, but just a placeholder which is a concrete type in the actual implementations):

```
pub fn register_texture(&mut self, texture: T) -> TextureId
pub fn deregister_texture(&mut self, id: TextureId)
pub fn clear_textures(&mut self)

fn lookup_texture(&self, id: TextureId) -> Result<T, RendererError>
```

with glium, T is probably `Rc<Texture2D>`
with gfx, T is probably something like `SharedResourceView<R, <[f32; 4] as Formatted>::View>`
with new master version of gfx, I don't really know it that well, but t with a quick glance it looks like this approach can work there too since we can accept an `Rc<"allocated memory handle">` or something like that.

@SpaceManiac, does this sound similar to what you're working on?

SpaceManiac:
Yes, very similar indeed. I found time for it tonight and opened the first version (it still needs a little work) at #154.

Gekkio:
Sorry, Github closed this pull request automatically so I forgot to comment earlier...In the end, the other pull request implemented this feature in a better way so it ended up being chosen, but I'd just like to thank you for taking the time to write this implementation! For me, `TextureCache` in this pull request was the missing piece that gave me an "a-ha" moment and allowed me to imagine what the final solution could roughly look like.

PR #152, opened 2018-09-01
Add Clone to ImGuiStyle

SpaceManiac:
It should probably be `Debug` and `PartialEq` too, but the `[ImVec4; 44]` gets in the way.

Useful in the situation that you need to save/restore styles in any way.


malikolivier:
Thanks! That's helpful. The CI is failing probably because minimum rust version should be bumped to 1.26.0. This was fixed in c5c9d6ff303dfcf61b8dd758d69689e1ef570855.

Could you rebase this PR so that we can be sure? Thanks!

SpaceManiac:
Rebasing seems to have done the job.

PR #154, opened 2018-09-10
Add custom texture support (both renderers)

### LINKED ISSUE #65: Using custom textures

SpaceManiac:
Alternative to #122 and #149. The `Image` struct is based originally on that from #149.

Closes #65. Closes #122. Closes #149.

* `ImTexture` is a typesafe wrapper around a `usize`.
* `Ui::image(&self, ImTexture, ImVec2) -> Image` allows placing an image in the UI.
* `Textures<T>` can be re-used by renderers to keep `ImTexture` -> texture mappings.
  * Renderers can expose `textures(&mut self) -> Textures<_>` for a common interface.

Pros:
* API surface is small (one `Ui` method, three support structs)
* No need for separate `imgui-*-textured-renderer` crates.
* `ImTexture` is typesafe (not confusable with non-texture `usize`s).
* No need to assign/deal with string IDs for textures (just pass around `ImTexture`s).
* Works with both renderers.
* No `unsafe` in the texture logic.
* The font is not a special case: it's typically `ImTexture(0)` due to the renderers registering it first, but is otherwise just another texture as far as the renderers are concerned.

Cons:
* Requires applications wanting to use custom textures to know about their renderer. I think this is inevitable because they need to know what format textures are expected to be in.
* ~~`gfx` renderer involves cloning Arcs~~ Fixed
  * Number of clones could be reduced by only cloning when texture ID changes
  * Investigation says a true fix exists, but I will have to actually write the code later
* Texture registration takes ownership of the texture
  * The texture can be retrieved again with `Textures::get` (a `get_mut` could be added if needed)
  * Not a problem for `gfx` where textures are just handles

Action shot:

image

TODO:
* [x] Reduce cloning in gfx renderer
* [ ] Write an example (port from #149?)
* [ ] Rustfmt

### DIFF
--- a/imgui-gfx-renderer/src/lib.rs
+++ b/imgui-gfx-renderer/src/lib.rs
@@ -8,7 +8,7 @@ use gfx::texture::{FilterMethod, SamplerInfo, WrapMode};
 use gfx::traits::FactoryExt;
 use gfx::{CommandBuffer, Encoder, Factory, IntoIndexBuffer, Rect, Resources, Slice};
 use gfx::pso::{PipelineData, PipelineState};
-use imgui::{DrawList, FrameSize, ImDrawIdx, ImDrawVert, ImGui, Ui, Textures};
+use imgui::{DrawList, FrameSize, ImDrawIdx, ImDrawVert, ImGui, Ui, Textures, ImTexture};
 
 pub type RendererResult<T> = Result<T, RendererError>;
 
@@ -18,6 +18,7 @@ pub enum RendererError {
     Buffer(gfx::buffer::CreationError),
     Pipeline(gfx::PipelineStateError<String>),
     Combined(gfx::CombinedError),
+    BadTexture(ImTexture),
 }
 
 impl From<gfx::UpdateError<usize>> for RendererError {
@@ -136,7 +137,6 @@ pub struct Renderer<R: Resources> {
     bundle: Bundle<R, pipe::Data<R>>,
     index_buffer: Buffer<R, u16>,
     textures: Textures<Texture<R>>,
-    backup_texture: Texture<R>,
 }
 
 impl<R: Resources> Renderer<R> {
@@ -175,7 +175,7 @@ impl<R: Resources> Renderer<R> {
             factory.create_sampler(SamplerInfo::new(FilterMethod::Trilinear, WrapMode::Clamp));
         let pair = (texture, sampler);
         let mut textures = Textures::new();
-        imgui.set_font_texture_id(textures.insert(pair.clone()));
+        imgui.set_font_texture_id(textures.insert(pair));
 
         let slice = Slice {
             start: 0,
@@ -193,7 +193,6 @@ impl<R: Resources> Renderer<R> {
             },
             index_buffer: index_buffer,
             textures: textures,
-            backup_texture: pair,
         })
     }
 
@@ -254,7 +253,9 @@ impl<R: Resources> Renderer<R> {
 
         self.bundle.slice.start = 0;
         for cmd in draw_list.cmd_buffer {
-            let tex = self.textures.get(cmd.texture_id.into()).unwrap_or(&self.backup_texture);
+            let texture_id = cmd.texture_id.into();
+            let tex = self.textures.get(texture_id)
+                .ok_or_else(|| RendererError::BadTexture(texture_id))?;
 
             self.bundle.slice.end = self.bundle.slice.start + cmd.elem_count;
             let scissor = Rect {
--- a/imgui-gfx-renderer/src/lib.rs
+++ b/imgui-gfx-renderer/src/lib.rs
@@ -227,7 +227,7 @@ impl<R: Resources> Renderer<R> {
 
         self.bundle.slice.start = 0;
         for cmd in draw_list.cmd_buffer {
-            if let Some(tex) = self.textures.get(cmd.texture_id as usize) {
+            if let Some(tex) = self.textures.get(cmd.texture_id.into()) {
                 // cloning handles is okay, since they're Arcs internally
                 self.bundle.data.tex = tex.clone();
             }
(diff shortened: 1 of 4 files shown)


malikolivier:
Thanks! I have a question, how would you register a new texture when the application is running?

If my understanding is correct, to register a new texture, tou would need to get a hold of a mutable reference to the `Textures` object in the UI loop. Is it possible to do that without much tedium?

An example would be very welcome!

imgui-glium-renderer/src/lib.rs
```diff
@@ -128,12 +132,17 @@ impl Renderer {
...
-        let font_texture_id = self.device_objects.texture.get_id() as usize;
-
         let mut idx_start = 0;
         for cmd in draw_list.cmd_buffer {
-            // We don't support custom textures...yet!
-            assert!(cmd.texture_id as usize == font_texture_id);
+            let texture = match self.device_objects.textures.get(cmd.texture_id.into()) {
+                Some(tex) => tex,
+                // if an invalid ID is supplied, fall back to the font
+                None => match self.device_objects.textures.get(0.into()) {
```
Gekkio: I think it would be nicer to return an error if an invalid ID is seen, because it's a clear programming error that should be fixed. We already have `RendererError` so it should be easy to just add another variant there

src/image.rs
```diff
@@ -0,0 +1,146 @@
+use super::{ImVec2, ImVec4, Ui};
+use std::marker::PhantomData;
+use std::collections::HashMap;
+use std::os::raw::c_void;
+use sys;
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
```
Gekkio: What's the semantic meaning of Ord here? I know it's easy to derive since the content is just usize, but is there any meaningful reason to order texture IDs? If not, let's just drop Ord and PartialOrd

src/image.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+/// Represent an image about to be drawn.
+/// See [`Ui::image`].
+///
+/// Create your image using the builder pattern then [`Image::build`] it.
+#[must_use]
+pub struct Image<'ui> {
+    /// we use Result to allow postponing any construction errors to the build call
```
Gekkio: This comment is a leftover and should be removed

src/image.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    pub fn insert(&mut self, texture: T) -> ImTexture {
+        let id = self.next;
+        self.textures.insert(id, texture);
+        self.next += 1;
+        ImTexture(id)
+    }
+
+    pub fn replace(&mut self, id: ImTexture, texture: T) {
+        assert!(self.textures.contains_key(&id.0));
```
Gekkio: Assert seems a bit unnecessary here since it's not the end of the world if the previous texture wasn't there.
What if we renamed the function a bit and return the previous texture if it existed?

```
pub fn insert_with_id(&mut self, id: ImTexture, texture: T) -> Option<T> {
```

This is basically just `HashMap::insert` so it's an insertion without automatic key generation. And the caller can decide what to do if the previous texture was or wasn't there.

src/image.rs
```diff
@@ -0,0 +1,146 @@
...
+        self.textures.insert(id.0, texture);
+    }
+
+    pub fn remove(&mut self, id: ImTexture) {
```
Gekkio: Let's return the removed texture here (return `Option<T>`)

src/lib.rs
```diff
@@ -170,8 +172,8 @@ impl ImGui {
             })
         }
     }
-    pub fn set_texture_id(&mut self, value: usize) {
-        self.fonts().set_texture_id(value);
+    pub fn set_texture_id(&mut self, value: ImTexture) {
```
Gekkio: I think we should now rename this function to `set_font_texture_id`, since the font texture is not going to be the only texture anymore.

imgui-gfx-renderer/src/lib.rs
```diff
@@ -214,10 +254,10 @@ impl<R: Resources> Renderer<R> {
 
         self.bundle.slice.start = 0;
         for cmd in draw_list.cmd_buffer {
-            // TODO: check cmd.texture_id
+            let tex = self.textures.get(cmd.texture_id.into()).unwrap_or(&self.backup_texture);
```
Gekkio: I think it would be better to return an error if an invalid id is seen. See my other comment in the glium renderer

src/image.rs
SpaceManiac: It's essentially "which texture was added first". It's not *that* useful, but outside of my UI I use ordering by texture IDs as a cheap hack to collate objects with the same texture adjacent in the draw list, to reduce draw calls.

Gekkio: Sequential ordering is a property of the generator, not the IDs themselves, so Ord seems slightly out of place. But I suppose there's no harm since IDs are newtype numbers anyway.

Gekkio:
I think the implementation looks very good now...are you still planning to write an example? It's not strictly necessary but would be nice. I'd like to get this merged very soon :smiley:

SpaceManiac:
My time has been somewhat consumed. I'd like to include an example, but it may be wiser to defer it to a future PR.

Gekkio:
That's alright, I'll merge this now :smile: Thanks for the great work!

PR #162, opened 2018-10-03
[WIP] Vulkano renderer 

### LINKED ISSUE #91: vulkan support
hi @Gekkio 
Do you have a plan to add vulkan support?

jaynus:
This is a WIP pull request for vulkano support. I wanted to create a PR early, for visibility and to get feedback/input on what needs to be included for what is considered a full renderer backend. I've implemented the bare basics from what I could see from the glium examples, but welcome further input on what all needs to be implemented. Would close #91

TODO: 

- [x] Working imgui rendering
- [x] winit key/mouse events (for cross support with vulkan-win)
- [ ] imgui image support (this doesn't seem to be supported in any of the other renderers either?)
- [x] Code conformance, lock vulkano+etc versions
- [x] clippy and rustfmt pass
- [ ] renderer should not use unwrap, and we should return RendererError instead in all error cases
- [ ] the vulkano renderer converts all ImDrawVert values to Vertex values. Is there no way to avoid this? It feels like a fairly big overhead, because in every frame we need to both 1) allocate a new vector (for collect()), and 2) go through all vertexes

Please let me know if I'm missing anything.


jaynus:
This code is now functionally complete, supporting imgui in the same capacity as the other backends. Image support is not implemented, but it doesn't seem that is implemented in any of the other examples either.

Gekkio:
Thanks for the hard work! There's been a lot of changes in master, so could you rebase the branch?
I don't know vulkan at all, but I'll try to give the pull request a proper look after it's up to date.

Here's some notes after a very quick glance:

* vulkano feature `imgui-sys` doesn't seem to have any effect? It should be removed if it's not needed
* custom texture support has landed in master. It shouldn't be very tricky to add, but needs to be supported before we can merge vulkan support to master
* renderer should not use unwrap, and we should return `RendererError` instead in all error cases
* the vulkano renderer converts all ImDrawVert values to Vertex values. Is there no way to avoid this? It feels like a fairly big overhead, because in every frame we need to both 1) allocate a new vector (for `collect()`), and 2) go through all vertexes

jaynus:
Sorry for the long delay. I have picked this up again, and have so far fixed the majority of your items. I've updated the root checklist to reflect the rest needed.

Gekkio:
No worries! :smile: 
I've got two new review comments based on a very quick look:

* The new renderer crate should use Rust 2018 and its conventions (e.g. no extern crates, import macros instead of using macro_use, etc...). Clippy + Rustfmt can probably help a lot here
* Dependency versions should be explicit (two digits, e.g. 1.0) and wildcard versioning (*) should not be used

PR #163, opened 2018-10-09
Use imgui-glutin-support in examples

malikolivier:
1. Update imgui-glutin-support to work with user-set HIDPI values. 97506a4 @Gekkio Are you for such feature? Is there any reason why it was not included to begin with?
2. Use imgui-glutin-support in glium-based examples. 1b48202

I used `imgui-glutin-support` on the gfx examples as well, however, as current latest glium still requires glutin 0.17 while gfx_window_glutin requires glutin 0.18, we get nasty version conflicts. I suggest to wait until glium is updated before using `imgui-glutin-support` for gfx examples.

### DIFF
--- a/imgui-examples/Cargo.toml
+++ b/imgui-examples/Cargo.toml
@@ -16,3 +16,4 @@ glutin = "0.18"
 imgui = { version = "0.0.21-pre", path = "../" }
 imgui-gfx-renderer = { version = "0.0.21-pre", path = "../imgui-gfx-renderer" }
 imgui-glium-renderer = { version = "0.0.21-pre", path = "../imgui-glium-renderer" }
+imgui-glutin-support = { version = "0.0.21-pre", path = "../imgui-glutin-support" }
--- a/imgui-examples/examples/color_button.rs
+++ b/imgui-examples/examples/color_button.rs
@@ -2,6 +2,7 @@ extern crate glium;
 #[macro_use]
 extern crate imgui;
 extern crate imgui_glium_renderer;
+extern crate imgui_glutin_support;
 
 use imgui::*;
 
--- a/imgui-examples/examples/hello_world.rs
+++ b/imgui-examples/examples/hello_world.rs
@@ -2,6 +2,7 @@ extern crate glium;
 #[macro_use]
 extern crate imgui;
 extern crate imgui_glium_renderer;
+extern crate imgui_glutin_support;
 
 use imgui::*;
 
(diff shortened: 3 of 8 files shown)


Gekkio:
> Update imgui-glutin-support to work with user-set HIDPI values

I didn't include this in the crate, because it feels like a hack and we should just trust the value coming from glutin. The hack is in the actual example support code because right now on Linux X11 winit's DPI choice is often bad and I wanted to force integer scaling in the examples to avoid people thinking that it's the fault of imgui-rs if things are blurry. However, it looks like winit is soon going to start taking the DPI value from same settings that other apps use (e.g. GTK apps), so the "artificial" DPI calculation is going to disappear and the hack should become unnecessary: https://github.com/tomaka/winit/pull/606

> Use imgui-glutin-support in glium-based examples

This looks good if you can get it to work :smiley: (right now the CI build is failing) After writing `imgui-glutin-support`, I noticed the same version conflicts and couldn't get the examples working so I just left it as it was. And yes, once glium and gfx agree on the glutin version, the problem goes away, but just temporarily. If one of those libraries bumps the glutin version and the other doesn't, we'll get the same situation again.

This is actually related to a bigger problem with how Cargo works: we'd like the glutin dependency in `imgui-glutin-support` to behave more like `peerDependency` in Node JS / NPM world. We'd just like to declare which versions are compatible, but Cargo should always use the version the user's project is already using and not pull in a conflicting version. I'm pretty sure this public/private dependency feature will help if it gets implemented at some point: https://github.com/rust-lang/rfcs/blob/master/text/1977-public-private-dependencies.md

malikolivier:
> This looks good if you can get it to work smiley (right now the CI build is failing) After writing imgui-glutin-support, I noticed the same version conflicts and couldn't get the examples working so I just left it as it was.

I can propose a fix. We can have two different projects for examples, as below:
- `imgui-glium-example`: contains all glium-based examples
- `imgui-gfx-example`: contains all gfx-based examples

Thus each project could agree on the version of glutin, and I think we can solve glutin version conflics. What do you think?

Gekkio:
I think you're right...we need to separate the examples. But instead of naming all the example crates based on the renderer, I think we should choose one renderer for the "main examples" and keep that crate as `imgui-examples`. Right now glium is the "main renderer", but I think we should switch to gfx since it's maintained a bit more actively. And once gfx-ll gets a release, we'll probably want to start using it immediately.

So, maybe something like this:

* `imgui-examples` (main examples, use gfx renderer here)
* `imgui-glium-examples` basic hello_glium example + maybe some texture stuff)
* `imgui-vulkano-examples` (once vulkano support lands, same idea as glium examples)

malikolivier:
Your suggestion looks good. I will make a PR implementing it.

PR #169, opened 2018-10-19
Use imgui-glutin-support in examples and split examples

malikolivier:
@Gekkio This is the suggestion we discussed in #163.

- Use `imgui-glutin-support` crate in all examples.
- Move examples in their own crates in their own workspace to prevent glutin version conflicts
- Allow setting HIDPI factor to `imgui-glutin-support` crate for current examples to work on all platform.

`imgui-glutin-support`'s glutin version is set in the Cargo.lock global to the workspace. I do not know of a way to using glutin 0.18 for `imgui-examples` and glutin 0.17 for `imgui-glium-examples`, unless we do not include the examples in the workspace. That's why I moved all the examples out of the main workspace. Then, I committed a working Cargo.lock for the glium example so that the correct version of glutin be selected.

Project tree now looks like this:

```
.
├── imgui
│   ├── Cargo.toml
│   ├── imgui-gfx-renderer
│   │   ├── Cargo.toml
│   │   └── src
│   ├── imgui-glium-renderer
│   │   ├── Cargo.toml
│   │   └── src
│   ├── imgui-glutin-support
│   │   ├── Cargo.toml
│   │   └── src
│   ├── imgui-sys
│   │   ├── build.rs
│   │   ├── Cargo.toml
│   │   ├── src
│   │   └── third-party
│   ├── src
│   │   ├── child_frame.rs
│   │   ├── ...
│   │   └── window.rs
│   └── target
├── imgui-examples
│   ├── Cargo.toml
│   ├── examples
│   │   ├── hello_gfx.rs
│   │   ├── ...
│   │   └── support_gfx
│   └── target
├── imgui-glium-examples
│   ├── Cargo.lock
│   ├── Cargo.toml
│   ├── examples
│   │   ├── color_button.rs
│   │   ├── ...
│   │   └── test_window.rs
│   └── target
├── LICENSE-APACHE
├── LICENSE-MIT
└── README.markdown
```

What are you thoughts on this?


malikolivier:
> Is it possible to use workspace exclusion instead of moving all the crates? I'm not sure how it exactly works, but in the 0.1-dev branch I've got a imgui-sys-bindgen crate that is excluded from the workspace.

Excellent! I did not know cargo had an exclude feature and that is exactly what we were looking for. I am not sure either how it works as I could not find it documented anywhere. I tried using "exclude" and made a new PR in #171. I close this PR.

PR #171, opened 2018-10-31
Use imgui-glutin-support in examples and split examples [use "exclude" Cargo.toml feature]

malikolivier:
@Gekkio This implement the suggestions we discussed in #163 and $168.

- Use `imgui-glutin-support` crate in all examples.
- Exclude examples from the main workspace to prevent glutin version conflicts
- Allow setting HIDPI factor to `imgui-glutin-support` crate for current examples to work on all platform.

Then, I committed a working Cargo.lock for the examples so that the correct version of glutin be selected in all cases.


malikolivier:
> I still don't like the extra DPI stuff in glutin support, but we can remove it later once winit improves DPI handling a bit.

Yes... I totally agree with you. Thank you!

PR #176, opened 2018-11-20
Bumped max acceptable glutin version to 0.19

agersant:
Transitively brings a lot of fixes from winit updates: https://github.com/tomaka/winit/blob/v0.18.0/CHANGELOG.md#version-0180-2018-11-07


agersant:
Uh, I think this will break the gfx renderer because the version on crates.io doesn't use glutin 0.19.

PR #180, opened 2018-11-23
Custom texture example

malikolivier:
This PR contains a new example that loads two textures. One is generated, while the other one is loaded from a JPG image. Uses the glium back-end.

image

NB:
- Moved all the resources for examples (the fonts and the images) to a shared `resources` folder.
- Updated glium support module to expose GL context and Textures.


rickyhan:
Your example is very clear and concise with different programming models. Bravo!

Issue #172, opened 2018-11-10
macOS Mojave (10.14.1) blank screen, no errors.


shousper:
Hi,

Just updated my mac, found my project rendering a blank screen 😞. So I ran the examples for `imgui-rs` and `gfx-rs` to confirm if one of those were broken, and it looks like it's just `imgui-rs`.

I tired both `master` branch and `v0.0.21` code, all examples, even with the `glium` renderer gives me a blank window. Also tried switching off dark mode, just in case 😛, no luck though, haha.

Sorry I can't offer much more than that right. Let me know if I can do anything more to help!

Cheers.

shousper:
Checked just now, the [imgui examples](https://github.com/ocornut/imgui/tree/master/examples) using metal or opengl2 work as well 👌. Trying to dig through `cimgui` and the `imgui-rs` bindings, but a little out of my depth!

malikolivier:
I see a fix on the [CHANGELOG](https://github.com/tomaka/winit/blob/master/CHANGELOG.md) of version 0.18.0 of winit saying `Fix issue whereby the OpenGL context would not appear at startup on macOS Mojave (#1069)`.

You will need to update `glutin` to 0.19.0.
This is impossible to upgrade `glutin` if you use `glium` as a backend as latest `glium` still depends on an old version of `glutin`.

I made a PR to update `glutin` for the `gfx` back-end. Could you confirm that it fixes your issue? Thanks.

PR #183, opened 2018-12-01
Update glutin to 0.19.0 on gfx examples

### LINKED ISSUE #150: Unexpected crash
Running of hello_world demo crashes program with following output: 

> thread '<unnamed>' panicked at 'assertion failed: validate_hidpi_factor(dpi_factor)', ...\winit-0.16.2\src\dpi.rs:153:9

malikolivier:
This should fix #150 and #172 for the `gfx` back-end.
Unfortunately, we cannot do the same for the `glium` back-end, as latest `glium` does not yet use `glutin` 0.19.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ rust:
   - stable
   - beta
   - nightly
-  - 1.26.0
+  - 1.28.0
 os:
   - linux
   - osx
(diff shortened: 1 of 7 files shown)


malikolivier:
glutin 0.19.0 depends on winit 0.18.0, which requires rust 1.28.0 or later. This explains the failing build.

Here is the full dependency chain:

glutin 0.19.0 > winit 0.18 > smithay-client-toolkit 0.4.2  > andrew 0.1.4 > rusttype 0.7.3 > ordered-float 1.0.1 (on linux). ordered-float uses the `[repr(transparent)]`.

The more time passes and the more I feel like rust is being npm-ified...

Maybe we should consider bumping the minimum supported rust version to support glutin 0.19?

PR #193, opened 2018-12-26
Update glium to 0.23

shockham:


Gekkio:
Thanks! :smile: 

This should fix some of the glutin version range issues, at least for now...

PR #198, opened 2018-12-31
Updates examples path in README (glium examples moved to their own folder)

JoshLambda:
**EDIT: I've submitted better PR #199**

~~Since the last release the examples for `glium` were moved into their own folder `imgui-glium-examples`.~~

~~The instructions in README specifically use these 'glium' examples, so the path needs to be updated there too.~~


JoshLambda:
I've submitted better PR #199

PR #203, opened 2019-02-25
Add support for image buttons

repi:
Found that image button support was missing, which we needed. So implemented it here.

Implementation is almost identical to `Image` just with some parameters being different, and that build returns if the button has been clicked or not.


Gekkio:
Thanks! :smile:

PR #204, opened 2019-03-01
add builder method to specify window pivot

sumeet:
could be used to center a window in the middle of the screen, like in
https://github.com/ocornut/imgui/issues/1657#issuecomment-370173339

---

i wasn't sure whether to call it `pivot()` or `position_pivot()`. if you want, i can rename it to whichever one you prefer! thanks :)


Gekkio:
Thanks! :smile:

PR #211, opened 2019-03-18
Allow InputText/InputTextMultiline to automatically resize their ImStrings if the user desires.

AlphaModder:
This PR adds a method `resize_buffer` to InputText and InputTextMultiline, which if called, sets the flag `ImGuiInputTextFlags::CallbackResize` and passes a callback that allows dear imgui to resize the `ImString` when it reaches capacity. 

Writing the callback required making some of `ImString`'s implementation details `pub(crate)` instead of private. If this seems ugly to you, I can think of a couple ways to avoid it: 
- Move the callback function from a freestanding function in `input.rs` to `ImString` itself. This avoids changing visibility, but still creates code nonlocality.
- Factor out the code that alters the capacity of the `ImString` into a public method, so that the callback can use it from its current location. The downside is that this method would appear to be a general-purpose resize operation, but have potentially confusing side-effects (refreshing the internal Vec's length and including the null terminator in it).

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -1,4 +1,5 @@
 use std::marker::PhantomData;
+use std::os::raw::{c_int, c_void};
 use std::ptr;
 use sys;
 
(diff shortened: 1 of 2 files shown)


Gekkio:
Thanks! :smile: 

Resize callbacks were not available when `ImString` was written, so the current 0.0.x API doesn't use them. I'll probably rework `ImString` and/or text inputs in 0.1 a bit based on this pull request (e.g. make resizing default, since there's no reason to not use it), but this is in any case a great addition for 0.0.x.

AlphaModder:
Yeah, I figured you had plans to rework this API for callbacks, but I thought this was so useful that it would be worth implementing even before then. Glad you agree! 😀

PR #213, opened 2019-04-17
Fixed a bug where get_time() returned bogus values

agersant:
See https://github.com/cimgui/cimgui/blob/67f3b097a5090dd074be959fb9c794bd75096a1c/cimgui.h#L1154


Gekkio:
:sob: 

Thanks! :smile:

PR #215, opened 2019-04-19
Fixed a bug where key states could get out of sync

agersant:
This fixes a bug where the state propagated to `imgui.set_key()` would be inaccurate when a key gets pressed while the glutin window is focused, but released while a different window is focused.

I discovered this issue while implementing a (native) File Open dialog which gets activated by pressing `Ctrl+O`. The window feeding the winit-support events would receive the two WindowEvents for `Ctrl` and `O` being pressed, but not the events for the keys being released. As a result, the `O` key got stuck in the pressed state from imgui's perspective.
As a side note, `Ctrl` key wasn't getting stuck because its state was being updated by `handle_modifiers(imgui, event.modifiers);` when I pressed escape to close the File Open dialog.

The `DeviceEvents` don't have this problem and are received regardless of the focused window. This change switches keyboard handling from using the `WindowEvents` to the `DeviceEvents` instead.

### DIFF
--- a/imgui-winit-support/src/lib.rs
+++ b/imgui-winit-support/src/lib.rs
@@ -83,7 +83,7 @@
 
 use imgui::{FrameSize, ImGui, ImGuiKey, ImGuiMouseCursor};
 use winit::{
-    ElementState, Event, KeyboardInput, ModifiersState, MouseButton, MouseCursor, MouseScrollDelta,
+    DeviceEvent, ElementState, Event, KeyboardInput, ModifiersState, MouseButton, MouseCursor, MouseScrollDelta,
     TouchPhase, VirtualKeyCode, Window, WindowEvent,
 };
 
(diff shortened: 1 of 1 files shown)


Gekkio:
Thanks! :smile: 

I actually noticed this same problem when doing stuff for the upcoming 0.1 version, and the problem should be fixed in the 0.1-dev branch.

I evaluated a similar fix as yours at first, but the problem is that it lets through events that should never arrive in the imgui window when it's not focused. For example, if you have a text input selected in an imgui window, pressing backspace while a completely different window is focused will erase characters from the imgui text input.

agersant:
I just looked at the code in the 0.1 branch and I agree that is much better!
Would you describe that branch as usable or should I hold on to master until it gets merged? I don't mind small and frequent breaking changes, but I do mind bugs and missing functionality.

PR #217, opened 2019-04-22
Add DirectX support

### LINKED ISSUE #55: imgui-gfx-renderer: Support for backends other than OpenGL
imgui-gfx-renderer includes shaders only for OpenGL, so other backends don't work right now.

tangmi:
Closes  #55

- Implement ImGUI shaders for HLSL Shader Model 4.0+
- Add feature flags "opengl" (default) and "directx" for `imgui-gfx-renderer`
- Implement constant buffer support in `imgui-gfx-renderer` behind "directx" feature flag
- Add DirectX support for `imgui-examples` behind the same feature flags

Some work that could be done:
- Add DirectX builds to TravisCI configuration
- `imgui-examples::support` has a lot of code duplication opportunities
- Code path for DirectX and OpenGL 3.1+ can be unified to use constant buffers (and have a special codepath only for OpenGL <= 3.0)
- Better design for enabling DirectX (e.g. implicitly via `target_os`?)
  - Edit: I've thought about this a little more and it might be doable with specialization over `Resources`, though I'm unsure if specialization has made it to stable.


tangmi:
I'm unfamiliar with TravisCI, but I'm wondering if the following would allow testing for this:

```yaml
script:
  - ...
  - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then cargo build --all --verbose --manifest-path imgui-examples/Cargo.toml --no-default-features --features directx ; fi
  - if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then cargo test --all --verbose --manifest-path imgui-examples/Cargo.toml --no-default-features --features directx ; fi
```

Gekkio:
Thanks! :smile:

tangmi:
@Gekkio Apologies that this is coming late, but I want to point out that my decision to make the backend based on cargo features has some problems when one is both depending on imgui-rs and selecting a backend via the compile target, e.g.:

```
[target.'cfg(windows)'.dependencies]
imgui-rs = { ... features = [ "directx" ] }

[target.'cfg(not(windows))'.dependencies]
imgui-rs = { ... features = [ "opengl" ] }
```

The issue arises because cargo doesn't (yet) handle target-specific features, and will compile with the union of all features across all targets. I'm happy to take another stab at this when I have some free time, if you want to revert this.

It's worth noting that usage is fine if one picks only one backend from imgui-rs.

The relevant issue is https://github.com/rust-lang/cargo/issues/1197

Gekkio:
Thanks for letting me know! I have very little time for imgui-rs at the moment, but I think I'll keep the change and revert only if somebody complains. :smile:

PR #222, opened 2019-05-18
Allow &[&ImString]  as well as &[&ImStr] in combo and list_box

AlexEne:
If you want to create a combo or list_box that has elements with dynamic names (read from a file or generated) things get a bit complicated.

now combo() and list_box api accepts both `&[&ImStr]` as well as `&[&ImString]`.
This makes it easy to push an `Vec<ImString>` into those UI elements.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1256,14 +1256,14 @@ impl<'ui> Ui<'ui> {
 
 // Widgets: Combos
 impl<'ui> Ui<'ui> {
-    pub fn combo<'p>(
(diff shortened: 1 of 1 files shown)


AlexEne:
Hi, Is this ok to merge or should I do some changes on it?

This is useful to populate a listbox of dynamic. Or is there another way to transform a Vec<String> into &[&ImStr] that I've missed?

Gekkio:
In 0.1 I'm experimenting with iterators instead of slices, but this change looks ok for 0.0.x. Thanks!

Issue #167, opened 2018-10-18
Discussion: the road to 0.1


Gekkio:
I've been thinking about doing some big refactorings in the project and doing a big push towards a 0.1 release. There's a couple of main topics I'd like to mention here:

### Automatic generation of bindings

In `imgui-rs`, we don't currently use bindgen and all the `imgui-sys` bindings are maintained by hand, which sometimes leads to human errors (fixed one mistake today I made in the latest cimgui upgrade). bindgen has been suggested in the past (most recently in #139) but there has been some cons in the approach, which is why I haven't been very excited about the idea. However, `cimgui` recently switched to automatic generation, which greatly changes the dynamics around the C bindings. I've got a WIP branch testing an approach that is slightly different to the previously suggested approach:

* instead of binding directly to Dear ImGui which generates non-portable C++ bindings, bind to `cimgui` which gives us portable C bindings. There's some non-portable preprocessor ifdefs, but those don't seem to end up affecting the generated Rust bindings
* instead of generating bindings at build-time which adds LLVM as a hard dependency, just regenerate bindings manually after every `cimgui` upgrade. Since the generated bindings are portable, there's no need to generate them on the fly

Let's take a look at the cons I listed in #139:

* *breaks pretty much all existing code (because many imgui-sys types are used in safe code too)*. 
This still applies, because we're replacing the entire imgui-sys API with an automatically generated one
* *user code becomes a mix of idiomatic and non-idiomatic code (because imgui-sys is used in safe code too)*.
This can still apply, but my next suggestion about the roles of -sys and the main crate should help
* ~~using bindgen with build.rs makes compiling much more complicated (e.g. LLVM is required on Windows. Do the bindings work with both msvc and gnu ABIs?)~~
Things should keep working on Windows just like they do now
* ~~LLVM 3.9 requirement means imgui-rs won't work out-of-the-box on Debian stable or old Ubuntu 16.04 LTS~~ 
LLVM won't be a requirement to use `imgui-rs`
* ~~a lot of magic in build.rs~~. 
Some "magic configuration" is still needed, but we can depend on `cimgui` to handle most of these things
* ~~bindgen has some limitations with C++. Are we sure the bindings work on all platforms and continue to work without causing a maintenance nightmare?~~ 
C++ limitations don't apply. We'll depend on `cimgui` to provide correct C bindings and there's a possibility of errors there, but `cimgui` is used by other bindings too so there's less risk than if we maintained complicated configuration ourselves

[...]

PR #223, opened 2019-05-25
Implement `get_*_draw_list` and `DrawList` drawing functions

### LINKED ISSUE #220: GetForegroundDrawList and GetBackgroundDrawList not exposed (upgrade Dear Imgui to 1.69)
These were recently exposed in Dear Imgui (https://github.com/ocornut/imgui/issues/2391, https://github.com/ocornut/imgui/releases) and are very useful - would be nice to have exposed in imgui-rs.

I guess this is a result of this being based on 1.66b while Dear Imgui is now at 1.69 (and so is cimgui).

tangmi:
Closes #220 (when 0.1 lands).

`ImDrawList_Get*DrawList` are very powerful functions and this PR exposes the capability in Rust.

- Uses `0.1`'s `AsRef<str>`-based way of passing in text.
- Uses `ImColor` from: https://github.com/Gekkio/imgui-rs/blob/1f5bb7c449e9f6f13dfd84640c426e7cdd91be49/src/window_draw_list.rs#L17-L53
- Does not implement the "advanced" (callbacks) or internal helper APIs
- Uses `ImVec2/4` and adds a conversion from `ImVec2/4_Simple` to `ImVec2/4`

### DIFF
--- a/imgui-examples/examples/draw_list.rs
+++ b/imgui-examples/examples/draw_list.rs
@@ -1,0 +1,37 @@
+use imgui::*;
+
+mod support;
+
+fn main() {
+    support::init(file!()).main_loop(|run, ui| {
+        let red = (1.0, 0.0, 0.0).into();
+        let green = (0.0, 1.0, 0.0).into();
+        let blue = (0.5, 0.5, 1.0).into();
+
+        Window::new(im_str!("Hello draw list"))
+            .opened(run)
+            .size([250.0, 250.0], Condition::Once)
+            .build(&ui, || {
+                ui.text("Hello world!");
+
+                let window_draw_list = ui.get_window_draw_list();
+
+                // in screen space
+                window_draw_list.add_circle((250.0, 250.0).into(), 50.0, red, 25, 1.0);
+                window_draw_list.add_text((160.0, 280.0).into(), red, "window screen space");
+
+                // in window space
+                let p1 = ui.get_cursor_screen_pos();
+                window_draw_list.add_line(p1, sys::ImVec2::new(p1.x + 50.0, p1.y + 50.0), red, 1.0);
+                window_draw_list.add_text(sys::ImVec2::new(p1.x + 50.0, p1.y + 50.0), red, "window local");
+            });
+
+        let bg_draw_list = ui.get_background_draw_list();
+        bg_draw_list.add_circle((250.0, 250.0).into(), 100.0, blue, 25, 1.0);
+        bg_draw_list.add_text((100.0, 160.0).into(), blue, "background");
+
+        let fg_draw_list = ui.get_foreground_draw_list();
+        fg_draw_list.add_circle((250.0, 250.0).into(), 75.0, green, 25, 1.0);
+        fg_draw_list.add_text((100.0, 200.0).into(), green, "foreground");
+    });
+}
--- a/imgui-sys/src/lib.rs
+++ b/imgui-sys/src/lib.rs
@@ -225,3 +225,17 @@ fn test_imvec4_simple_memory_layout() {
     assert_eq!(&ref_a.z as *const _, &ref_b[2] as *const _);
     assert_eq!(&ref_a.w as *const _, &ref_b[3] as *const _);
 }
+
+impl From<ImVec2_Simple> for ImVec2 {
+    #[inline]
+    fn from(v: ImVec2_Simple) -> ImVec2 {
+        ImVec2::new(v.x, v.y)
+    }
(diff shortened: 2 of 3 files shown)


hrydgard:
Cool! Looks great, exactly what I'll need. Any idea what the schedule for 0.1 is?

tangmi:
I think that's up to @Gekkio, the relevant links are #167 and #168.

tangmi:
Whoops, I just found https://github.com/Gekkio/imgui-rs/blob/master/src/window_draw_list.rs, which probably this probably should share code with.

hrydgard:
Right, indeed..

tangmi:
@Gekkio I've cleaned this up, added an example to `imgui-examples`, and squashed the changes into a single commit. This should be ready to review whenever you have time!

Gekkio:
Thanks! :smile: 

I didn't yet have a chance to do a full review, but here's some things I noticed quickly:

* this seems to target 0.1-dev, but that is a throwaway branch
* starting from 0.1, ImVec2/ImVec4 are not allowed in safe APIs
* what's the difference between this and the existing draw list API?
* what's the reason of having `ImColor`? Colours are `[f32; 4]` everywhere else
* all `get_xx_draw_list` functions cause UB, because they can generate references with any lifetime
* all `get_xx_draw_list` functions cause UB, because they can generate multiple mutable references to the same thing
* reusing the `DrawList` type for this use case causes UB, because you can have active references to idx_buffer/vtx_buffer/DrawCmdIterator while also modifying the data underneath

tangmi:
Thanks for the reply, I'll try and update the code to respond to this feedback, so please hold off on a full review for now!

>    this seems to target 0.1-dev, but that is a throwaway branch

I'll rebase this onto master

>    starting from 0.1, ImVec2/ImVec4 are not allowed in safe APIs

Sounds good, I'll change the usage to `[f32; 2]` and `[f32; 4]`

>    what's the difference between this and the existing draw list API?

No difference! I thought that `0.1-dev` was newer than master! I'll change this to use the existing draw list API in master.

>    what's the reason of having ImColor? Colours are [f32; 4] everywhere else

I change this to use `[f23; 4]`, although I think there's something to be said for having a type to document (and enforce) that colors stay within the [0, 1] range.

>    all get_xx_draw_list functions cause UB, because they can generate references with any lifetime

I'm still learning about unsafe Rust--I thought something like this implies `'ui: 'a`? (please correct me here!)

```
impl<'ui> Ui<'ui> {
    pub fn get_window_draw_list<'a>(&self) -> &'a mut DrawList {
        ...
    }
}
```

I think `DrawList` in `master` has a `'ui` lifetime, so I can have this function instead be

```
impl<'ui> Ui<'ui> {
    pub fn get_window_draw_list(&self) -> &'ui mut DrawList<'ui> {
        ...
    }
}
```

if that'll be better?

>    all get_xx_draw_list functions cause UB, because they can generate multiple mutable references to the same thing
>    reusing the DrawList type for this use case causes UB, because you can have active references to idx_buffer/vtx_buffer/DrawCmdIterator while also modifying the data underneath

I _think_ these can be addressed together. I'll try and find a solution for this. I initially had the `get_xx_draw_list` functions take `&mut self`, but I remember that causing API usage issues since one would often want other parts of `Ui` while doing a draw.

Gekkio:
> I'm still learning about unsafe Rust--I thought something like this implies 'ui: 'a? (please correct me here!)

I don't think it does...if we remove the effect of lifetime elision, I think it's basically equal to: 

```
impl<'ui> Ui<'ui> {
    pub fn get_window_draw_list<'self, 'a>(&'self self) -> &'a mut DrawList {
        ...
    }
}
```

The function takes a `&'self Ui<'ui>` (with some lifetime `'self` for the reference and `'ui` for the struct data) and returns a mutable reference `&'a mut DrawList` with *any lifetime 'a* (which could be `'static` and it would compile). I think the right thing in this case might be to drop the explicit lifetimes completely.

It's also worth noting that the current draw list API might have safety holes as well, so you shouldn't trust it without thinking...in 0.1 I've tried to fix many safety issues, but the window draw list API is one of the many things I didn't have time or energy to look at.

tangmi:
@Gekkio would you consider using bindgen (or checking in bindgen output, with instructions on how to regenerate)? I'm running into issues upgrading master to imgui >= 1.69, which is required for this PR. the 0.1-dev branch seems to use bindgen and be updated to 1.69.

tangmi:
Abandoning in favor of #242

PR #225, opened 2019-06-09
Update glium to 0.25.0

thomcc:
Mostly this seems to have been a new version of `glutin`.

Manually testing shows this working on all examples on macOS. (I have no reason to believe it wouldn't also work on other platforms).


Gekkio:
Thanks! :smile:

PR #228, opened 2019-06-30
Add support for winit 0.20-alpha via feature toggle

aloucks:
Winit's next version introduces many breaking changes. This PR adds support for `0.20` via a feature toggle while keeping backwards compatibility with version `0.19` by default.


Gekkio:
Thanks! :smile: 

Upgrading to winit 0.20 is definitely something we want to do, but I'm a bit unsure at this point about supporting both winit versions at the same time...other crates don't seem to do it and it's going to be much more complicated to maintain.

aloucks:
The intent was to only support `0.19` until `0.20` left the alpha phase. This would allow people to start using imgui-rs with winit `0.20` sooner rather than later while not disrupting others with the `0.20` alpha churn. I didn't clarify that on the initial PR, so apologies on that front.

Gekkio:
I've given this some more thought and think we need to support both versions at least temporarily.

The main reason for supporting both versions is that gfx-rs is still on 0.19, so we can't just upgrade to 0.20. I've got a hunch we'll need to drop gfx-rs support soon completely because it's not maintained actively...

PR #232, opened 2019-07-03
Do not reexport ffi imgui functions on windows

Veykril:
This define disables reexporting all imgui functions on windows builds reducing the library size when building dynamic linked libraries with it. See https://github.com/cimgui/cimgui/blob/2b03f434d4afd5cae9d352035edc1fbfd1ac3c2d/cimgui.h#L8-L12

Edit: Travis seems to fail since the 0.3 version of memoffset has been yanked

### DIFF
--- a/imgui-sys/build.rs
+++ b/imgui-sys/build.rs
@@ -27,6 +27,7 @@ fn main() -> io::Result<()> {
     build.cpp(true);
     // Disabled due to linking issues
     build
+        .define("CIMGUI_NO_EXPORT", None)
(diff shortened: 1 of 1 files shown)


Gekkio:
Thanks! :smile: 

Could you rebase? I've upgraded memoffset so this pull request should now be ok with Travis.
Any downsides to disabling the re-exporting? I'm not very familiar with Windows myself.

Veykril:
There shouldn't be any downsides regarding this, since no one would want to be able to access the c imgui functions from this as a dll I believe.

PR #237, opened 2019-07-15
Replace static mut bool with AtomicBool

nico-abram:
Replace possibly UB-happy and racy static mut boolean with an AtomicBool.

If this is not appropiate, and there's ensurance that this is thread-safe, at the very least I think the RacyUnsafeCell described here should be used https://github.com/rust-lang/rust/issues/53639#issuecomment-415515748 .


Gekkio:
Thanks! :smile:

PR #242, opened 2019-07-23
Implement `get_*_draw_list` and `DrawList` drawing functions

### LINKED ISSUE #220: GetForegroundDrawList and GetBackgroundDrawList not exposed (upgrade Dear Imgui to 1.69)

tangmi:
Closes #220

- Merge 1imgui::render::DrawList1 with 1imgui::WindowDrawList1
- Rename `WindowDrawList` to `DrawList` and make `[Window|Foreground|Background]DrawList` into a wrapper for `DrawList`
- Add `'ui` lifetime to `DrawData` and `*DrawList`
- Make all `DrawList` methods take `&mut self` instead of `&self`

**Breaks DrawData/DrawList API?**

- [ ] Rename `window_draw_list` to `draw_list`, but for now, this keeps the diff sane.
- [ ] Expose `DrawList`, `ImColor` publicly
- [ ] Path API?

### DIFF
--- a/src/render/draw_data.rs
+++ b/src/render/draw_data.rs
@@ -1,14 +1,15 @@
 use std::slice;
 
-use crate::internal::{RawCast, RawWrapper};
+use crate::internal::RawCast;
 use crate::render::renderer::TextureId;
 use crate::sys;
+use crate::window_draw_list::DrawList;
 
 /// All draw data required to render a frame.
 #[repr(C)]
-pub struct DrawData {
+pub struct DrawData<'ui> {
     valid: bool,
-    cmd_lists: *mut *mut DrawList,
+    cmd_lists: *mut *mut DrawList<'ui>,
     cmd_lists_count: i32,
     /// For convenience, sum of all draw list index buffer sizes
     pub total_idx_count: i32,
@@ -29,9 +30,9 @@ pub struct DrawData {
     pub framebuffer_scale: [f32; 2],
 }
 
-unsafe impl RawCast<sys::ImDrawData> for DrawData {}
+unsafe impl RawCast<sys::ImDrawData> for DrawData<'_> {}
 
-impl DrawData {
+impl<'ui> DrawData<'ui> {
     /// Returns an iterator over the draw lists included in the draw data
     pub fn draw_lists(&self) -> DrawListIterator {
         unsafe {
@@ -40,9 +41,9 @@ impl DrawData {
             }
         }
     }
-    pub(crate) unsafe fn cmd_lists(&self) -> &[*const DrawList] {
+    pub(crate) unsafe fn cmd_lists(&self) -> &[DrawList] {
         slice::from_raw_parts(
-            self.cmd_lists as *const *const DrawList,
+            self.cmd_lists as *const DrawList,
             self.cmd_lists_count as usize,
         )
     }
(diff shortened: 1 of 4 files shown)


src/render/draw_data.rs
```diff
@@ -107,54 +108,8 @@ fn test_drawdata_memory_layout() {
...
-            DrawCmdIterator {
-                iter: self.cmd_buffer().iter(),
-            }
-        }
-    }
-}
-
 pub struct DrawCmdIterator<'a> {
-    iter: std::slice::Iter<'a, sys::ImDrawCmd>,
+    pub(crate) iter: std::slice::Iter<'a, sys::ImDrawCmd>,
```
tangmi: needed since the `DrawList` impl above was moved to `imgui::window_draw_list`

src/render/draw_data.rs
```diff
@@ -40,9 +41,9 @@ impl DrawData {
             }
         }
     }
-    pub(crate) unsafe fn cmd_lists(&self) -> &[*const DrawList] {
+    pub(crate) unsafe fn cmd_lists(&self) -> &[DrawList] {
```
tangmi: This returns a DrawList straight up, since DrawList now owns a pointer to ImDrawList, rather than being an ImDrawList itself

src/render/draw_data.rs
```diff
@@ -68,14 +69,14 @@ impl DrawData {
 
 /// Iterator over draw lists
 pub struct DrawListIterator<'a> {
```
tangmi: This type no longer needs to exist (it's simply a wrapper for slice::Iter currently), but I'm keeping it to avoid changing the interface. Could be replaced with
```rust
/// Iterator over draw lists
pub type DrawListIterator<'ui, 'a> = std::slice::Iter<'a, DrawList<'ui>>;
```

tangmi:
@Gekkio I believe this PR should be a safe implementation of the drawlist apis.

Notably:
- All DrawList functions take &mut self, as they mutate the underlying draw list
- `DrawList` is bound to `'ui` now
- `add_circle`, etc return a builder with a mutable reference to the DrawList with a new lifetime `pub fn add_circle<'a, ...>(&'a mut self, ...) -> Circle<'ui, 'a>`

saucesaft:
What can be done to implement this functions into imgui-rs? Anything i can help with?

tangmi:
@saucesaft Aside from the merge conflict and failing CI checks, the bit in my last comment about lifetimes regarding `with_clip_rect` and `channels_split` I think is still true. Another thing I recall when using this branch is that while there is `WindowDrawList`, `ForegroundDrawList`, and `BackgroundDrawList`, there is no easy way to write code abstracted across them (this might be as simple as making `DrawList` `pub`.

I think @Gekkio is pretty busy (I am too!) and I believe neither of us have looked at this since the summer! Feel free to clean this up and re-propose the changes, I think some test cases that sanity check the API would catch things like in the second half of my last comment.

Boscop:
Any update on this? :)

In my use case I need access to the drawlist..

tangmi:
Closing, as I believe this was addressed in #414!

PR #245, opened 2019-07-27
Add name to FontConfig

aloucks:
Allows for fonts to be named when loaded from memory. 

image


Gekkio:
Thanks! :smile: 

Name is marked as "internal" field in upstream code, but I don't see any problem using it

PR #250, opened 2019-08-18
Update the winit version requirements to be compatible with wgpu-rs.

Yatekii:
Hi!

I had to bump the version so it works with up to date crates such as wgpu-rs.
If there is any problems with this I cannot see right now, please tell me and I'll fix it.

Best,
Yatekii


Yatekii:
Closing this PR in favor of a way better one!

PR #251, opened 2019-08-23
Mark a bunch of methods that return a builder as `must_use`

Jasper-Bekkers:
I noticed I kept forgetting to call `.build()` on a bunch of UI elements, so I marked them as `must_use` so you get a compile-time warning if you ignore the result. Should save some time debugging why certain widgets won't show up.

I'm not sure I've covered all of them, so feel free to point out missing ones & I'll add those as well.


Gekkio:
Thanks! :smile:

PR #252, opened 2019-08-28
Disable return-type-c-linkage warning

aclysma:
When building, I see a bunch of warnings:

```
warning: In file included from third-party/cimgui/cimgui.cpp:5:
warning: third-party/cimgui/cimgui.h:974:19: warning: 'igGetWindowPos' has C-linkage specified, but returns user-defined type 'ImVec2' which is incompatible with C [-Wreturn-type-c-linkage]
warning: CIMGUI_API ImVec2 igGetWindowPos(void);
warning:                   ^
warning: third-party/cimgui/cimgui.h:975:19: warning: 'igGetWindowSize' has C-linkage specified, but returns user-defined type 'ImVec2' which is incompatible with C [-Wreturn-type-c-linkage]
warning: CIMGUI_API ImVec2 igGetWindowSize(void);
```

I don't think these warnings are useful in this case. The returned values in question (ImVec2, ImVec4, ImColor) shouldn't pose any practical trouble. This change will suppress the warning. (Although I understand if you don't want to take this PR since it isn't targetted at just those types.)

### DIFF
--- a/imgui-sys/build.rs
+++ b/imgui-sys/build.rs
@@ -30,6 +30,8 @@ fn main() -> io::Result<()> {
         .define("CIMGUI_NO_EXPORT", None)
         .define("IMGUI_DISABLE_WIN32_FUNCTIONS", None)
         .define("IMGUI_DISABLE_OSX_FUNCTIONS", None);
+
+    build.flag("-Wno-return-type-c-linkage");
     for path in &CPP_FILES {
         assert_file_exists(path)?;
         build.file(path);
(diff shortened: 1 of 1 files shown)


Gekkio:
Thanks! :smile: 

> The returned values in question (ImVec2, ImVec4, ImColor) shouldn't pose any practical trouble.

The warnings are relevant and calling those functions cause UB on some platforms (Mac OS at least...?). As far as I understand, this is caused by "fun" C++ concepts like user-defined types vs PODs, which affect how the return values are handled in the calling conventions. In other words: calling `igGetWindowPos` from C/Rust code on Mac OS most likely returns garbage data and/or crashes the program.

However, imgui-sys doesn't expose these functions at all, and instead uses the "nonUDT"-variants defined in cimgui that are fully C-compatible. The warnings for the normal functions are still there, because we can't easily stop the C++ compiler from compiling these (unused) functions. I think it's ok, at least for now, to disable the warning.

PR #253, opened 2019-08-28
Fix backspace text entry

aclysma:
I had trouble with backspace not functioning properly on winit/macos. The issue was that character \u{7f} was being inserted, and then the backspace would delete that spurious character. I also fixed a compile warning from deprecated '...' since it was in the same file.


Gekkio:
Thanks! :smile: 

I released v0.2.0 yesterday, but will most likely backport this fix with v0.2.1

repi:
Thanks, this is something we ran into as well

PR #258, opened 2019-09-06
Small fix for missing .end() statement on new ComboBox causing a panic

jaynus:
This is just a 1-liner fix for adding an end statement inside the new `build_simple` function in `ComboBox`. This statement wasn't calling `end()`, causing a panic from the `drop` of the `ComboToken`.

I just wanted to push this up so I don't diverge from master too much.


Gekkio:
Thanks! :smile:

PR #259, opened 2019-09-09
Fix ImageButton bool return value for pressed

repi:
This was a regression in 0.2.0, in that large change the `build` function lost its return value so one couldn't check if the image button had been pressed or not.

Would it be possible to have a `0.2.1` published on crates.io with this fix?

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


Gekkio:
Thanks! :smile: 

Yeah, I'll release 0.2.1 of `imgui` + `imgui-winit-support` later today

PR #269, opened 2019-10-13
Update example code in README

toyamaguchi:
Hello! I updated the example code in README because many beginners refer to this code.
This code is almost same as following hello world code, except the argument "&ui" of build().
https://github.com/Gekkio/imgui-rs/blob/master/imgui-examples/examples/hello_world.rs


Gekkio:
Thanks! :smile:

PR #280, opened 2019-11-06
Remove resources dir from published crate

Jake-Shadle:
The fonts and image in /resources are only used by examples, but they are still included in the published crate, along with their licenses, even though they aren't used in downstream crates, so this just removes them from the published crate.

Alternatively, I could change it to move the resources into the examples directory and fix up the paths in the examples instead.


Gekkio:
Thanks! :smile:

PR #284, opened 2019-12-06
Fix toggling of MenuItem and Selectable with build_with_ref

malikolivier:
`Selectable` or `MenuItem` should be toggled on click when called with `build_with_ref`.
    
This is the behavior that is done in ImGui when  `bool ImGui::MenuItem(const char* label, const char* shortcut, bool*p_selected, bool enabled)` is called.

## Before the fix: Cannot toggle
bug

## After the fix: Can toggle (expected behavior)

fix

With this fix, the `test_window_impl` example works just as the upstream C++ ImGui demo works.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -10,6 +10,11 @@
 
 - Various things that were deprecated in imgui-rs 0.2.0
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


Gekkio:
Thanks! :smile:

PR #285, opened 2019-12-06
Add initial bindings for docking branch

rasky:
This PR adds initial support for the Dear Imgui `docking` branch. 

The docking/viewport branch is a long-living branch that will eventually be merged to master and brings many new features to Dear Imgui. While some features can be used almost without changing code (docking), others require extensive support in the backend (viewports). Since it might take a while to correctly implement for viewports in our several backends, I thought we might as well start implementation.

This PR has initial support: it passes cargo test by fixing the layouts, but does not yet wrap new APIs. It was written off v0.2, as 0.3 is still not stable.

*NOTE*: This PR should probably be merged into a new `docking` branch. There's no way for me to target a branch that doesn't exist yet; if somebody creates a `docking` branch, I can change this PR and let it point there.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "imgui-sys/third-party/cimgui"]
 	path = imgui-sys/third-party/cimgui
 	url = https://github.com/Extrawurst/cimgui.git
+	branch = "docking"
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "imgui-sys/third-party/cimgui"]
 	path = imgui-sys/third-party/cimgui
 	url = https://github.com/Extrawurst/cimgui.git
-	branch = "docking"
+	branch = "1.77dock"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ travis-ci = { repository = "Gekkio/imgui-rs" }
 bitflags = "1.0"
 glium = { version = "0.25", default-features = false, optional = true }
 gfx = { version = "0.18", optional = true }
-imgui-sys = { version = "0.2", path = "imgui-sys" }
+imgui-sys = { version = "0.2.1", path = "imgui-sys" }
 lazy_static = "1.1"
 parking_lot = "0.9"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,10 @@ imgui-sys = { version = "0.2.1", path = "imgui-sys" }
 lazy_static = "1.1"
 parking_lot = "0.9"
 
+[features]
+default = ["docking"]
+docking = []
+
 [dev-dependencies]
 memoffset = "0.5"
 
--- a/imgui-sys/src/bindings.rs
+++ b/imgui-sys/src/bindings.rs
@@ -160,6 +160,7 @@ pub type ImGuiBackendFlags = ::std::os::raw::c_int;
 pub type ImGuiColorEditFlags = ::std::os::raw::c_int;
 pub type ImGuiConfigFlags = ::std::os::raw::c_int;
 pub type ImGuiComboFlags = ::std::os::raw::c_int;
+pub type ImGuiDockNodeFlags = ::std::os::raw::c_int;
 pub type ImGuiDragDropFlags = ::std::os::raw::c_int;
 pub type ImGuiFocusedFlags = ::std::os::raw::c_int;
 pub type ImGuiHoveredFlags = ::std::os::raw::c_int;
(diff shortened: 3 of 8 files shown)


rasky:
Ping @Gekkio

rasky:
Ping @Gekkio, any feedback is appreciated

Gekkio:
Sorry, I don't have time for imgui-rs at the moment and know nothing about the dear imgui `docking` branch, so it's difficult to give feedback. :cry:

Jasper-Bekkers:
@Gekkio If you're looking for an additional maintainer I'm more then happy to step up, us at Embark have been using this crate quite extensively.

jaynus:
@Jasper-Bekkers @Gekkio 

I would be willing to take over maintenance of it as I also use it heavily in the amethyst org. I already had a branch for docking and a PR for it as well, which was denied (https://github.com/jaynus/imgui-rs/tree/docking).

I think we would all much appreciate if you could please allow a maintainer (the embark folks are AWESOME!) to take the crate over so it can continue to be developed and maintained. I'm sure they would agree to strictly 'maintenance' if that was your desire, but frankly there are some critical changes that weren't finished (the change from functions to builders for everything) as well as support for docking and platform backends which are critical to get in.

Alternatively, the amethyst org would be willing to take over a fork of imgui. (I'm sure embark would as well).

rasky:
I hope there’s no need to fork, Rust community is small enough without having the need to further split maintenance of packages. @Gekkio I hope you will allow a co-maintainer of this very useful crate if you don’t have time to follow it.

Gekkio:
Let's continue the discussion here to avoid spamming this pull request with general stuff about the project: https://github.com/Gekkio/imgui-rs/issues/292

rasky:
@Gekkio since you say that this PR has little information, what information do you need? The changes in this PR are simply “translations” from C++ to rust, adjustment of data structure to match the layout in C++, constants, and stuff like that: whatever is required to make tests pass. I didn’t write any new API (yet) so there’s zero design choices. I’m happy to provide missing information.

Gekkio:
* What is the docking branch? (links to *up-to-date* upstream docs is fine, you don't have to write it yourself). "brings many new features to Dear Imgui" as a description requires me to hunt down additional information to understand what's going on, which I'd gladly do *if* I had time for imgui-rs
* What is the expected lifecycle of the docking branch? Are we planning to add docking as a feature flag (easy in theory but in practice complicated to maintain in this case), or are we expecting it to land in upstream master soon and this pull request to be merged into imgui-rs master?
* How are viewports and docking related? The name of this pull request only mentions docking, but viewports are also mentioned in the description. It sounds like both are wanted...?
* Does the docking branch change semantics of existing APIs in any way (including addition of new asserts in the C++ code, which might be triggered in existing code paths)?

jaynus:
@Gekkio I use this docking branch heavily, so I'll answer,

1. The docking branch is the current working/community review branch by ocornut for docking abilities for imgui windows. It adds all the features necessary for allowing it 
2. I would honestly recommend the following: A new cimgui fork is created by you and maintained by the same collab list, copying the changes from this fork in this PR. (So they can sync). Then, imgui-rs master tracks this new cimgui. The code itself is basically the community testing for docking and viewports, so it *is* going to land in imgui master at some point. However, due to ocornut's development lifecycle this could be a fair long while before it does. 
3. Viewports were implemented specifically for docking; they are one-in-the-same a part of the imgui docking features. This is because the docking branch *also* allows dragging docked things "out of the window" and for the platform backend to spawn an entirely new window, if it supports it. The docking features are built upon these viewport features, so in the imgui source they are tightly coupled.
4. There are no breaking API changes from an existing API standpoint, aside from the new viewport/platform registration needs. The rest of the API feature set is hidden behind a imgui global setting, that is off by default. 

This PR basically allows the internals of the docking branch to be accessed. There still has to be various builders created for the actual internals; however, the imgui docking feature *Just Works(tm)* by flipping the config flag for docking. The additional API surface that needs to be built for it, currently, only relates to custom docking, e.g. programatically (not user driven) docking control.

rasky:
I personally disagree on point 2. For me, the best course of action is to use the standard `cimgui` upstream that already tracks the docking branch in [their own docking branch](https://github.com/cimgui/cimgui/tree/docking), which is what my PR is based upon (one of the thing that my PR does is to change the submodule to track the docking branch).

Then, I'd personally prefer a `docking` branch in `imgui-rs` as well; the feature flag would be more difficult to maintain and I'm not sure it gives a big benefit, especially once you think that in some time there will not be two branches anymore but just one.

rasky:
I also want to provide an additional datapoint: while it's true that using the docking feature for end-developers is very easy (just flip the docking flag), using the multi-viewport support requires more extensive changes in the backend plugins as the internal API for backend plugins is extended to support API to create multiple windows.

One of the goal of this PR is to help starting adapting these backends. For instance, I use `rust-imgui-sdl2` that will have to be adapted. It's easier to do once there is a `docking` branch in `imgui-rs`, that can be tracked to verify progress and to check that the implementation is working.

Once DearImgui merges the main docking branch, ideally all the binding chain should just merge their own docking branch (cimgui -> imgui-rs -> rust-imgui-sdl2).

PR #302, opened 2020-02-17
Updated winit to use new modifier framework in imgui

sanbox-irl:
Winit now emits a DeviceEvent when modifier keys are altered, which is more accurate than tracking the individual key changes. A common scenario where is happen is on macOs when you try to take a screen recording of the window -- the modifier keys will never be reset with a keydown event. Winit WILL however, emit a ModifiersChanged deviceevent, so I switched this implementation to track those and now we're successfully tracking input!

I left in tracking keycodes for the modifiers from the WindowsEvent since I don't think Winit sends us false-positives, just missing some inputs. There may be some OS/environment where Winit won't give us DeviceEvents for modifier keys (I doubt it, but maybe), so double checking might help that weird framework. If you think it's unnecessary, let me know.

### DIFF
--- a/imgui-winit-support/src/lib.rs
+++ b/imgui-winit-support/src/lib.rs
@@ -391,13 +391,19 @@ impl WinitPlatform {
                 ..
             } => {
                 io.keys_down[key as usize] = false;
-                match key {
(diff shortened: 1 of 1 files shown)


sanbox-irl:
👍 if anyone out there is on a platform like Linux on X11, please check this commit!

PR #303, opened 2020-03-10
Font name length check should use the minimum length

aloucks:


Gekkio:
Thanks! :smile: 

Could you rebase the branch on top of master? The winit version range problems prevented the CI from running

aloucks:
@Gekkio done!

PR #309, opened 2020-04-02
Add support for winit 0.22, fix some clippy warnings

jaynus:
This simply adds support for `winit 0.22`,  adding the appropriate changes where necessary and sharing feature flags with `winit-20` where possible.

Also fixed a few clippy warnings, since they were annoying me in my project.


aloucks:
I think you need to update the CI config so that `winit-22` feature is tested.

Gekkio:
Thanks! :smile: 

Could you fix the formatting errors and also rebase? I just had to bump minimum Rust version to 1.40 in master because a dependency uses features not available in 1.38.
Also, as suggested earlier, the CI config should test `winit-22` as well.

PR #311, opened 2020-04-05
Clamp textures to border (BorderClamp / GL_CLAMP_TO_BORDER)

Boscop:
I have a zoomable texture that I render with `ImageButton` (I also need to handle mouse click event on it).
When I zoom out, the texture is repeated, because GL_REPEAT is the default for textures.
But I need to see where the texture begins and ends!
This PR changes the texture wrap_function to `BorderClamp`.

Before:
image

After:
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
         rust:
           - stable
           - beta
-          - 1.38.0
+          - 1.40.0
         os:
           - ubuntu-latest
           - macos-latest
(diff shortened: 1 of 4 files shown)


Gekkio:
Thanks! :smile: 

Could you rebase? I just had to bump minimum Rust version to 1.40 in master because a dependency uses features not available in 1.38.

Instead of changing globally to `GL_CLAMP_TO_BORDER` to suit your use case, what we really need is the ability to specify per-texture sampling parameters (like gfx-renderer allows you to do). But that would require much more work and `GL_CLAMP_TO_BORDER` is a sane default, so I'll accept this change. :slightly_smiling_face:

Boscop:
Thanks, I just rebased. (For some reason it only let me push after I pulled from my origin/master again, but that put my commit below your latest one..)

PR #318, opened 2020-05-08
implementation of TabBar and TabItem

SnoozeTime:
Hi, I added a simple implementation of the tab bar and items.
I also added an example to the imgui-example folder.

Please let me know if you need anything more :)


SnoozeTime:
It looks like some files that are not in my PR are failing the rustfmt check. Do you want me to fix them at the same time?

Gekkio:
Thanks! :smile: 

A lot of stuff has happened in master, so please rebase. I also have one review comment: the test_window_impl example is meant to be a reimplementation of the C++ imgui_demo.cpp and not a freeform place for examples. So, you should either implement the tab bar example code based on the C++ demo *or* make a separate example

SnoozeTime:
Hi there, thanks for the review! I changed window_test_impl example to follow the ccp version.

PR #319, opened 2020-05-10
Add support for winit 0.22

### LINKED ISSUE #308: Update to glium 0.27
glium has a new version: 0.27

yodaldevoid:
Updates @jaynus's PR (#309) with the comments provided.

I don't like doing this, but I needed winit 0.22 support and ended up doing the work, so I figured I should provide a PR since @jaynus doesn't seem to have the time to fix up their PR. Feel free to close this if I am stepping on any toes.

Also updates `glium` to 0.27.

Closes #307 
Closes #309 
Closes #308

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,11 @@ jobs:
         with:
           command: check
           args: --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-20
+      - name: Run cargo check (imgui-winit-support with winit-22)
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-22
       - name: Run cargo fmt
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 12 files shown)


yodaldevoid:
It's looking `glium` will need to be updated to update `glutin` so that the same version of winit is use across the board. This will probably require feature flags to be added to `imgui-glutin-renderer` because the `glutin` update is not a drop in replacement.

yodaldevoid:
I've added a commit explicitly selecting `winit` 0.21 for `imgui-examples`. I've also figured out what the problem was with upgrading `glium`: I simply forgot to update `glium` in all `imgui` crates. Once I updated everything to `glium` 0.27, everything compiled without issue. We can stick with the solution I currently have pushed and update `glium` in a different PR, or we can update `glium` in this PR as well.

Gekkio:
Thanks! :smile: 

I think updating glium at the same time would be good, because I try to have all the examples and the main crate work out of the box when used with the "primary platform", which happens to be glium at the moment.

yodaldevoid:
Assuming that all the PR tests pass, that should also update to `glium` 0.27 across the board.

Gekkio:
Sorry, can you rebase again? We just merged another pull request to master so this one is no longer up to date :smile_cat:

yodaldevoid:
No problem. It's all taken care of.

PR #326, opened 2020-05-23
Make cmd_lists_count public in DrawData struct

filnet:
Also updated comments with latest from imgui.

Relates to https://github.com/Gekkio/imgui-rs/issues/325

### DIFF
--- a/src/render/draw_data.rs
+++ b/src/render/draw_data.rs
@@ -4,15 +4,18 @@ use crate::internal::{RawCast, RawWrapper};
 use crate::render::renderer::TextureId;
 use crate::sys;
 
-/// All draw data required to render a frame.
+/// All draw data to render a Dear ImGui frame.
 #[repr(C)]
 pub struct DrawData {
+    /// Only valid after render() is called and before the next new frame() is called.
     valid: bool,
+    // Array of DrawList.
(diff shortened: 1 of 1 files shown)


aloucks:
This should probably be a getter method instead. Making it public allows the value to be mutated which makes `draw_lists()` unsafe: 

```rust
    pub fn draw_lists(&self) -> DrawListIterator {
        unsafe {
            DrawListIterator {
                iter: self.cmd_lists().iter(),
            }
        }
    }
    pub(crate) unsafe fn cmd_lists(&self) -> &[*const DrawList] {
        slice::from_raw_parts(
            self.cmd_lists as *const *const DrawList,
            self.cmd_lists_count as usize,
        )
    }
```

filnet:
Changed to a setter as requested.

aloucks:
cc @Gekkio

Gekkio:
Thanks! :smile: 

Would it make more sense to return `usize` instead of `i32`? The public count fields are `i32` to match the underlying memory layout, but in a getter function we have the option of casting the value to something more appropriate for Rust code. Note also that you can already get the count with `draw_lists().count()`, which returns `usize`.

filnet:
Changed getter to return usize instead of u32.

PR #327, opened 2020-05-25
Manually managing tooltips

barsoosayque:
Add ability to create tooltips without passing a lambda. Essentially, it's a copy-paste from `Ui::begin_group()` and uses the same Token-based implementation. As with other methods like this, it's very useful when dealing with mutable references both before and inside lambda.


Gekkio:
Thanks! :smile:

PR #331, opened 2020-06-11
Added a feature to support the imgui features to use u32 draw indices

JoNil:


JoNil:
I have tested this on the glium backend and looked at the index buffer in render doc. There i can see that it is allocated correctly as u32.

aloucks:
Why not use `ImGuiBackendFlags_RendererHasVtxOffset` and `ImDrawCmd::VtxOffset` instead?

JoNil:
Cool didn't know about that! Doesn't seem like the imgui-glium-renderer supports it though? Do we want to expose both ways for people with back-ends that don't support it?

miloyip:
> Why not use `ImGuiBackendFlags_RendererHasVtxOffset` and `ImDrawCmd::VtxOffset` instead?

I just experiment this but the rendering results are glitches. It seems that the imgui-xxx-renderer have not support this.

I noticed that https://github.com/ocornut/imgui/issues/3129 are also related to this. Hopefully the fix can be merged into this, have renderers handling VtxOffset and have an example show correct rendering of large mesh.

Gekkio:
Thanks! :smile: 
Is this ready for merging? I don't have a good understanding of the changes so I'll just trust them and merge this if everything is ready.

aloucks:
I think a better solution is to update the renders to make use of vtx offset. It's both more efficient and the [recommended way for ImGUI](https://github.com/ocornut/imgui/blob/master/imgui.h#L1935). Using 32bit indices is a work-around.

Gekkio:
Can somebody test https://github.com/Gekkio/imgui-rs/pull/346 if you have a real-world project where the 16-bit limit is hit and the vtx offset becomes relevant?

norcalli:
I might resurrect this. We're hitting this limit at work.

sanbox-irl:
heya -- it's been nearly a year and a half since this was opened. I'm not sure what the status is, but can I get an update on it?

dbr:
Example!

```rust
use imgui::*;

mod support;

fn main() {
    let mut text = String::new();
    for _row in 0..1_000 {
        for _char in 0..100 {
            text += "a";
        }
        text += "\n";
    }
    system.main_loop(move |_, ui| {
        ui.window("Hello world")
            .size([300.0, 110.0], Condition::FirstUseEver)
            .build(|| {
                ui.child_window("Hm").build(||{
                    ui.text_wrapped(&text);
                });
            });
    });
}
```

If you scroll the window it will crash with:

```
hello_world: ./third-party/imgui-master/imgui/imgui.cpp:4364: void AddDrawListToDrawData(ImVector<ImDrawList*>*, ImDrawList*): Assertion `draw_list->_VtxCurrentIdx < (1 << 16) && "Too many vertices in ImDrawList using 16-bit indices. Read comment above"' failed.
```

aloucks:
@dbr What renderer are you using? The renderer should be using the vertex offset to avoid having to use 32bit indices.

dbr:
@aloucks Hmmm, that's a good point! I'm using `imgui-glow-renderer` and the same problem happens with the glium renderer used in `imgui-examples` - both of which have `RendererHasVfxOffset`

I can draw, say, a million lines with the draw list API and it is fine - which leads me to think it might be a bug or limitation with imgui's `TextWrapped` method. It appears to draw everything in a single `DrawCmd` (so the VfxOffset cannot help):

text_wrapped_drawlist

dbr:
Made an issue upstream about this, https://github.com/ocornut/imgui/issues/5720

dbr:
Ah the text-wrapping assertion error was a upstream bug where the characters were output in a way which would not benefit from the vertex-offset. This is fixed in upstream v1.89

Tempted to close this for now given the lack of real-world examples (..at least those which didn't turn out to be unrelated bugs like mine :blush: )

PR #339, opened 2020-06-27
Expose include path and native defines via `DEP_` env variables

aloucks:
This allows downstream projects to access the include path for `cimgui.h` and `imgui/imgui.h` for compiling native imgui dependencies. 

The `third_party` directory path will be available to `build.rs` scripts via the  `DEP_IMGUI_THIRD_PARTY` environment variable. The defines are available via `DEP_IMGUI_DEFINE_xxx`.

### DIFF
--- a/imgui-sys/Cargo.toml
+++ b/imgui-sys/Cargo.toml
@@ -9,6 +9,7 @@ repository = "https://github.com/Gekkio/imgui-rs"
 license = "MIT/Apache-2.0"
 categories = ["gui", "external-ffi-bindings"]
 build = "build.rs"
+links = "imgui"
 
 [build-dependencies]
 cc = "1.0"
--- a/imgui-sys/build.rs
+++ b/imgui-sys/build.rs
@@ -2,6 +2,7 @@
 
 use std::fs;
 use std::io;
+use std::path::Path;
 
 const CPP_FILES: [&str; 5] = [
     "third-party/cimgui.cpp",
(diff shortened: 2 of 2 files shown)


4bb4:
@aloucks ping me when this is merged :)

Gekkio:
Thanks! :smile: 
I don't have a good understanding of the practical effects and possible concerns of this change, but I'll just merge it anyway

4bb4:
@Gekkio thanks for addressing this! I'll be using this in https://github.com/4bb4/implot-rs/blob/master/implot-sys/build.rs to figure out where the C++ imgui headers are in `implot-rs`, and I assume @aloucks created the PR here for doing the same in a project of his. I haven't been able to get things to work yet because my knowledge of the `sys` crate mechanisms is still limited though.

aloucks:
@4bb4 I started building a wrapper for ImGuizmo. The wrapper is still rather clunky and unfinished, but the sys crate should be a good-enough example for how to use the `DEP_` env vars.  Implot was on my list too, but it sounds like you've already taken the lead with that one!

https://github.com/aloucks/imguizmo-rs/blob/master/imguizmo-sys/build.rs

4bb4:
@aloucks Thanks for the hint, I appreciate the guidance. The link 404s for me though, is the repo private?

4bb4:
@aloucks 
> Implot was on my list too, but it sounds like you've already taken the lead with that one!

I'd be open to collaborating. We can talk about this more over at https://github.com/4bb4/implot-rs to avoid spamming @Gekkio with notifications.

PR #341, opened 2020-06-28
imgui-glium-renderer: User-selectable texture filters per-texture

### LINKED ISSUE #287: Customize sampling with glium
The glium renderer always samples with linear interpolation:

https://github.com/Gekkio/imgui-rs/blob/ccec55c10a787f92b21054476f5817506c38692e/imgui-glium-renderer/src/lib.rs#L198-L200

I have a use case where I'd like to set this to Nearest: displaying sprites at magnified scales, where the original pixels should be clearly visible. For now, I've simply replaced the filter in the source of the renderer, but it'd be nice to customize this on a per-texture basis.

cosmicchipsocket:
Specific textures used in imgui-glium-renderer can now have their own MagnifySamplerFilter and MinifySamplerFilter defined. There's an extra hash lookup per texture, but unfortunately I don't see a better way unless additional libraries are pulled in or the existing API is broken.

Fixes #287


cosmicchipsocket:
Was unsatisfied with the original change so I replaced it with the API-breaking version that changes the template argument of `Textures` to use a plain-old-data object containing the texture and the min/mag filters. I just hope that it's okay to change something like this without a proper deprecation warning.

cosmicchipsocket:
For now, fork is here as a separate crate: https://crates.io/crates/imgui-glium-renderer-filter

cosmicchipsocket:
@Gekkio I don't think you can use a SamplerBehavior to set the properties of a Sampler created from a glium Texture in one go. Should I just set all the properties individually? What about properties that wouldn't make much sense, such as DepthTextureComparison?

Gekkio:
Note that `Sampler` is literally made from a texture reference and `SamplerBehavior` and these are public fields, so you should be able to construct it in one go with `Sampler(texture, behavior)` instead of using `texture.sampled()` or `Sampler::new`:

```
pub struct Sampler<'t, T: 't>(pub &'t T, pub SamplerBehavior);
```

cosmicchipsocket:
Look good now?

PR #347, opened 2020-07-09
Add support for TabItem flags

RoyJacobs:
Since support for `TabBar` and `TabItem` got merged, here's a small PR to also be able to set the flags of the `TabItem`.


Gekkio:
Thanks! :smile:

PR #354, opened 2020-08-12
Add glutin support to `imgui-winit-support`

8bitkitkat:
Surprisingly trivial, all I needed to do was add a cfg condition to wherever there was a winit-22 condition.

Working test [Gitlab - 8bitkitkat/imgui-winit-supoort-gluin-test](https://gitlab.com/8BitKitKat/imgui-winit-support-glutin-test)


Gekkio:
Doesn't glutin just re-export winit stuff, so you should be able to use winit-22 feature directly?
See the discussion in https://github.com/Gekkio/imgui-rs/issues/188 and how `imgui-glutin-support` was changed into `imgui-winit-support` around two years ago.

8bitkitkat:
huh, it does work. Don't know why I thought it required its own feature.

PR #355, opened 2020-08-18
Fixed `SharedFontAtlas` not being used

N3xed:
When creating a new imgui context, the optional shared font atlas was not being used.
This caused imgui to crash because no font atlas existed.


Gekkio (approved):
Thanks! :smile: 

Clearly not that many people use this feature, since this has been broken for quite a long time now...

PR #369, opened 2020-09-23
Add comments to DrawCmd

toyboot4e:
This PR adds two line of comments that are useful to who want to implement custom renderer.

It lets users know:

* `clip_rect`: it's `[left, up, right, down]` not `[x, y, w, h]`
* `count`: it's the number of indices not the number of triangles
---
Thank you for sharing the awesome crate!


Gekkio (approved):
Thanks! :smile:

PR #371, opened 2020-10-02
Winit 0.23 support for imgui-winit-support

JMS55:
Of note is that `MouseScrollDelta::PixelDelta` now returns a `PhysicalPosition`, and not `LogicalPosition`. I just converted it using `self.scale_factor`.

This bumps the crate version, and makes winit 0.23 the default.

The glium examples don't compile anymore, glium needs to update glutin which needs to update winit. I tested with the gfx examples.


m4b:
Glutin 0.25 uses winit 0.23; should be able to bump it and should just work afaics?

JMS55:
Glium hasn't updated to a new version of Glutin. I suppose I could use `glutin` directly, and not `glium::glutin`? Any thoughts?

m4b:
Can you just send a PR or open an issue there ? Winit changes need to propagate throughout the ecosystem :)

JMS55:
My bad, closer inspection actually shows glium will do a new release soon https://github.com/glium/glium/commit/303f4f95fb4a64acedac0be1732b69b35e2b7b9c

JMS55:
Now that glium updated I went through and tried to get the examples to compile. In test_windows.rs, line 169, I get the error "the trait `glium::Vertex` is not implemented for `imgui::DrawVert`". Glium says to use the implement_vertex!() macro, but warns that struct fields must match their names in the shader exactly. However, there are multiple different shaders for different glsl/opengl versions. Any thoughts on what to do?

m4b:
So what’s the status on this ? Would be really great to get winit 0.23 support, since most crates in the ecosystem have now moved

m4b:
Actually is it not possible to just add the imgui winit support changes and default to winit-22 feature ? And then publish this crate ? That should allow downstream users to upgrade to winit 0.23 and to continue allow glium deps, etc., to continue compiling ?

benmkw:
@m4b I opened https://github.com/Gekkio/imgui-rs/pull/381#issue-505515187 by accident and it worked for a basic example with wgpu backend for me but pretty sure there will be follow up issues cause that seems too easy :D

JMS55:
Agreed that defaulting to winit 0.22 could be an option, and allow upgrading for those using wgpu or otherwise. @Gekkio, what do you think is the best option?

maxded:
What exactly still needs to happen to get this PR in? Can't bump my project's winit version cause of imgui-winit-support. If needed, I can of course always help out. :)

Gekkio:
> What exactly still needs to happen to get this PR in?

Well, passing the tests would be a good start :smile:
I'd greatly prefer upgrading glium and winit at the same time, because glium is supposed to be the most supported renderer in imgui-rs. In fact, I'll almost certainly remove all support for gfx-rs in the near future.

I don't have time to investigate the problems in this pull request, but I noticed that the lock file has two conflicting `glium` versions. Maybe the upgrade is incomplete?

PR #376, opened 2020-10-11
Add wrapper for PopupContext

phoracek:
This element opens an overlay menu when user clicks on a Window or into the void of the screen.

See an example:

```
cd imgui-examples
cargo run --example test_window_impl
# Examples -> Simple Overlay -> Right click
```

Signed-off-by: Petr Horacek <phoracek@redhat.com>


Gekkio:
Thanks! :smile: 

Some comments:

* The `begin_*` functions conflict with current conventions, because the functions that take a closure are generally called `build_*`
* The actual `begin_*` functions which don't need the use of closures are missing
* You should take `&Ui` as a parameter only in the final functions that build the widget (= `build_*` and `begin_*`. This also reduces or even eliminates the need for lifetimes in the builder type
* Demo code looks good! It's a bit more verbose than the original C++ (and more readable IMHO :wink: ), but it looks like a reasonable implementation of the same thing

phoracek:
@Gekkio thanks for the review! I appreciate it. Great point with the redundant lifetimes. I think it should be all set now \o/

PR #381, opened 2020-10-18
update to winit 0.23 

benmkw:
I tested this with a small project of mine using the imgui-wgpu backend and it worked, are there other tests I should run?


benmkw:
sorry did not see the open PR

PR #385, opened 2020-11-09
Winit 0.23 support for imgui-winit-support 

maxded:
Continuation on @JMS55 winit 0.23 support PR. 

- Bumps the `imgui-winit-support` crate version, and makes winit 0.23 the default.
- Bumps the `glium` dependency to 0.28
- Bumps `imgui-rs` minimal Rust version to 1.41


maxded:
`owned_ttf_parser-0.6.0` seems to break the minimal 1.40.0 requirement. What is the normal procedure when this happens? Update the minimal requirements?

Note: tested these changes locally on Windows, my imgui project seems to work fine!

JMS55:
@maxded thank you for continuing my PR! I've been busy with school and haven't had time to take another look at it.

I think the issue I had was that the way you defined Vertex's with Glium changed, and imgui has like 5 different shader versions with different parameters, which made implementing the Vertex trait confusing.

cwfitzgerald:
Could we get a winit dependency with no defualt features? I use winit without wayland support nad this is currently forcing it on.

Gekkio (approved):
Thanks! :smile: 
Looks good to me :+1:

PR #391, opened 2020-11-21
rework imgui-sys and imgui-bindgen to support docking branch

tgolsson:
Hello! 

I know there is already one PR open to add support for the upstream imgui docking branch, but I took a slightly different approach, and wanted to see if it would be more palatable as the other one got stuck. I choose to make this available as a feature toggle, requiring users to actively opt-in.

This PR is deliberately incomplete, focusing on the binding generation and imgui-sys, as those became very awkward with a feature-based approach. The primary change is that the git submodule has been removed in favor of downloading imgui on-build and checking out the right commit then. This whole chain is driven from the cimgui update script; which decides the two commits that'll be used. 

In short, the flow is as follows:

* cimgui generates bindings for both main branch and docking
* imgui-sys-bindgen multiplexes the above sets with wasm, generating four variants
* imgui is downloaded from git for builds and a hardcoded commit is
checked out - this is read from a file generated in the cimgui step, maintaining synchronization
* the docking feature flag locks out all docking-related functionality when not enabled, at all levels

Also:
* also adds the minimal set of config attrs and fields to the high-level wrappers to make it safe to mix in imgui-sys calls in high-level code and make tests pass with and without the feature
* adds a single test step for the docking feature to make sure the structures are kept valid in both modes
* marks the generated files and dictionaries for linguist

I've got some initial high-level wrappers too; but I feel like that fits more as a separate PR once the groundwork has been laid. 

Let me know what you think and if we can make this code accessible! Looking at the network graph it seems as if every other fork is about docking, so I think it'd be very useful to concentrate all that effort here.


tgolsson:
After sleeping on this, it'd probably be neater at this point to generate the rust bindings at build time too; and only check in the cimgui bindings (since they depend on luajit, which we probably don't want as a build dependency...).

PR #397, opened 2020-12-05
added new for texture id

sanbox-irl:
This very small commit adds a constructor for `TextureId`. I have forgotten how to make it for a usize a few times. Now that IDE support is much better in Rust, finding `::new` is generally easier than `From` methods, though obviously, I didn't remove the `From` impls, so those who prefer those are good to go too.


thomcc (approved):
LGTM, probably could be a `const fn`, but I'll make that change if it seems valuable after looking more closely.

PR #398, opened 2020-12-06
Ownership transfer: fix clone URL

sadika9:


thomcc (approved):
Ah, thanks! I wonder why I didn't find that when I did a search.

PR #406, opened 2020-12-16
Use MAIN_SEPARATOR to make consistent behavior.

dskkato:
As is the path separator in Windows is '\' instead of '/', It seems to be better to use MAIN_SEPARATOR that Rust provides to make consistent behavior across platforms.


thomcc (changes requested):
Instead we should probably just use `Path`, no?

dskkato:
Right, that's what I wanted to do.

dskkato:
I think the result of `to_str()` here can be safely unwrap since original title is valid utf-8 bytes array.
If there is a better way, I would follow the way.

Thanks.

PR #404, opened 2020-12-11
support winit-0.24

willemv:
I've checked the breaking changes in winit 0.24, and none of them affect this `imgui-winit-support` crate. Simply allowing the 0.24 version for the `winit-23` feature should be sufficient.

I also have the change ready where there is a separate `winit-24` feature, adjusting the `imgui-winit-support/src/lib.rs` file to take it into account. Let me know if you prefer that approach over this one.

PR #407, opened 2020-12-16
Support both glium 0.28 and 0.29

willemv:
As discussed in https://github.com/imgui-rs/imgui-rs/pull/404 : allow both `glium` 0.28 and 0.29.


thomcc (approved):
Thanks!

PR #411, opened 2020-12-23
Cimgui upgrade to 1.79

dneumair:
Hi i restructured the imgui-sys crate to reflect the original layout of the original cimgui repository. 
Cimgui was not a git submodule but the files where part of this repository. Furthermore the file structure was NOT the same as the original cimgui repository.

I also added the missing "internal" functions, which enables experienced users to create more complex widgets.
Everything internal is part of imgui_sys::* and unsafe!

Unfortunatly bindgen seems to either have changed the way enums are represented or (even worse) generates different representations on windows and unix
Imgui uses u32 while bindgen (at least on my windows machine) generated i32. I did not test unix yet

Finally i added a new example "imgui_sys" which can be run with 
"cargo run --examples imgui_sys" and shows what is possible with access to the internal functions

but:

THIS SHOULD NOT BE MERGED TO MASTER YET

@thomcc please create a new branch, i do not want to accidentally introduce bugs which break the build system


thomcc:
It might not be until the start of the year until I get to reviewing this, sorry for any trouble.

dneumair:
no problem
by the way i badly burned my finger with direct use of c++ from bindgen!
everything just worked fine but msvc decided to change the abi of the ImVec2 because as soon as you introduce a constructor to even a small struct it is no longer an plain old datatype...

everything compiled and ran just fine but calling into a function with a specific signature (ImVec2 as first parameter) corrupted the stack pointer register silently...
the parameters suddendly where shifted (e.g. the value of the second parameter was in place of the first one).

the result was a silent crash without even an exeption or error code. 
Only "Process finished with exit code 0"

long story short: Do not use Bindgen with c++ unless rustc introduces an extern "c++" calling convention ;)

thomcc:
> long story short: Do not use Bindgen with c++ unless rustc introduces an extern "c++" calling convention ;)

Yeah, I hadn't looked yet, but was going to ask you not to change it if this was using bindgen-for-C++. Anyway, thanks for the PR, I'll take a look after the new years. Cheers.

sanbox-irl:
I'm going to close this since we're behind 1.79. If this used a new technique that you'd like me to take a look at, please lmk

PR #414, opened 2020-12-28
Refactor WindowDrawList into DrawListMut and fix #413.

### LINKED ISSUE #413: Problem with get_background_draw_list()
**Describe the bug**
If one uses ui.get_background_draw_list() outside of the window context, a new window is being created with "Debug" title.

**To Reproduce**
I modified the "hello_world" example with this code, note that it should draw a thick line across the screen, but additional window is being created:
```rust
use imgui::*;

mod support;

fn main() {
    system.main_loop(move |_, ui| {
        let dl = ui.get_background_draw_list();
        dl.add_line([0.0, 0.0], [640.0, 480.0], [1.0, 0.0, 0.0])
            .thickness(5.0)
            .build();

        Window::new(im_str!("Hello world"))
            .size([300.0, 110.0], Condition::FirstUseEver)
            .build(ui, || {
                ui.text(im_str!("Hello world!"));
                ui.text(im_str!("こんにちは世界！"));
                ui.text(im_str!("This...is...imgui-rs!"));
                ui.separator();
                let mouse_pos = ui.io().mouse_pos;
                ui.text(format!(
                    "Mouse Position: ({:.1},{:.1})",
                    mouse_pos[0], mouse_pos[1]
                ));
            });
    });
}
```

**Expected behavior**
It should not create any additional windows when using drawlists.

**Screenshots**
image

**Please describe your environment**
I tried it on "master" right now, but it's valid for 0.6.1 as well. OS is irrelevant, but it's linux.

**Other information**
Quick glance at the source code shows this:
https://github.com/imgui-rs/imgui-rs/blob/eada611c2ce822054596833c8edb3122e77ee7db/imgui/src/lib.rs#L490-L493

And this:
https://github.com/imgui-rs/imgui-rs/blob/eada611c2ce822054596833c8edb3122e77ee7db/imgui/src/window_draw_list.rs#L87

My guess is that because igGetWindowDrawList() is called for background draw list as well, it causes window to be revealed in imgui when used outside of the window context. Quick glance at imgui's source code indeed confirms this, they always have "Debug" window in the stack:
https://github.com/ocornut/imgui/blob/fa963b9aafde7f05b1d32d42e3d2b084c9685e31/imgui_internal.h#L2140

Most likely avoiding the call to igGetWindowDrawList() will fix this. I.e. this sould be a "static method": 
https://github.com/imgui-rs/imgui-rs/blob/eada611c2ce822054596833c8edb3122e77ee7db/imgui/src/window_draw_list.rs#L92

nsf:
- Rename WindowDrawList -> DrawListMut. It's not about window draw lists, but
  about background/foreground draw lists as well. The naming was not an easy
  choice, but seems like in rust it's a common convention to add a Mut suffix for
  mutable entities. Imgui-rs already has DrawList and it acts as an immutable
  reference type for rendering implementations to consume. Hence the name
  DrawListMut, which becomes a mutable reference to draw list with methods
  to modify it.
- Add Ui::get_foreground_draw_list(). Same as Ui::get_background_draw_list()
  but for foreground.
- Add draw_list example which shows the use of all three draw lists
  (window, bg, fg).

### DIFF
--- a/imgui-examples/examples/draw_list.rs
+++ b/imgui-examples/examples/draw_list.rs
@@ -1,0 +1,90 @@
+use imgui::*;
+
+mod support;
+
+// rect is [x, y, w, h]
+fn draw_text_centered(
+    ui: &Ui,
+    draw_list: &DrawListMut,
+    rect: [f32; 4],
+    text: &ImStr,
+    color: [f32; 3],
+) {
+    let text_size = ui.calc_text_size(text, false, 0.0);
+    let cx = (rect[2] - text_size[0]) / 2.0;
+    let cy = (rect[3] - text_size[1]) / 2.0;
+    draw_list.add_text([rect[0] + cx, rect[1] + cy], color, text);
+}
+
+fn main() {
+    let system = support::init(file!());
+    system.main_loop(move |_, ui| {
+        {
+            let bg_draw_list = ui.get_background_draw_list();
+            bg_draw_list
+                .add_circle([150.0, 150.0], 150.0, [1.0, 0.0, 0.0])
+                .thickness(4.0)
+                .build();
+            draw_text_centered(
+                ui,
+                &bg_draw_list,
+                [0.0, 0.0, 300.0, 300.0],
+                im_str!("background draw list"),
+                [0.0, 0.0, 0.0],
+            );
+        }
+
+        {
+            let [w, h] = ui.io().display_size;
+            let fg_draw_list = ui.get_foreground_draw_list();
+            fg_draw_list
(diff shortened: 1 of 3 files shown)


thomcc (commented):
This looks great. It will be a bit until I can merge it since I'd like to check the example on different platforms first but overall very nice.

thomcc:
Sorry for the delay here, I finally have some time to test stuff today.

thomcc (approved):
This is great, and works great. Thanks a ton. I don't know when this will be in a release but probably relatively soon.

PR #416, opened 2020-12-29
inline calls to key_index into methods which operate on keys

### LINKED ISSUE #415: UI::is_key_released may be easy to use incorrectly because of u32 key_index
I had the following bug:  `ui.is_key_released(imgui::Key::A as u32)` which should have been `ui.is_key_released(ui.key_index(imgui::Key::A))`.

If a breaking change would be acceptable I would think making `is_key_released` accept a `Key` (or marking it a method on the Key and passing the ui) would be harder to use incorrectly. This also applies to other related methods which take a `key_index`.

benmkw:
I am not sure if someone needs to call `key_index` so maybe this should still remain pub but I could not see an obvious reason (and also not specific reason for the casts but don't have a strong opinion about it either).

closes #415

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Changed
+
+- is_key_down, is_key_pressed, is_key_released and key_pressed_amount now make a call to key_index internally and thus take a Key instead of u32
+
+### Removed
+
+- key_index is now only used internally
+
 ## [0.6.0] - 2020-11-15
 
 ### Added
--- a/imgui/src/input/keyboard.rs
+++ b/imgui/src/input/keyboard.rs
@@ -94,31 +94,35 @@ impl<'ui> Ui<'ui> {
     /// Returns the key index of the given key identifier.
     ///
     /// Equivalent to indexing the Io struct `key_map` field: `ui.io().key_map[key]`
-    pub fn key_index(&self, key: Key) -> u32 {
-        unsafe { sys::igGetKeyIndex(key as i32) as u32 }
+    fn key_index(&self, key: Key) -> i32 {
+        unsafe { sys::igGetKeyIndex(key as i32) }
     }
     /// Returns true if the key is being held.
     ///
     /// Equivalent to indexing the Io struct `keys_down` field: `ui.io().keys_down[key_index]`
-    pub fn is_key_down(&self, key_index: u32) -> bool {
-        unsafe { sys::igIsKeyDown(key_index as i32) }
+    pub fn is_key_down(&self, key: Key) -> bool {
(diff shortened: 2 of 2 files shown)


CHANGELOG.markdown
```diff
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Changed
```
thomcc, line 5: I'm realizing now I should have been updating this all along 😅

imgui/src/input/keyboard.rs
```diff
@@ -94,31 +94,35 @@ impl<'ui> Ui<'ui> {
     /// Returns the key index of the given key identifier.
     ///
     /// Equivalent to indexing the Io struct `key_map` field: `ui.io().key_map[key]`
-    pub fn key_index(&self, key: Key) -> u32 {
-        unsafe { sys::igGetKeyIndex(key as i32) as u32 }
+    fn key_index(&self, key: Key) -> i32 {
+        unsafe { sys::igGetKeyIndex(key as i32) }
```
thomcc, line 98: I do wonder if we should be doing `ui.io().key_map[key]` instead, which would allow for the compiler to optimize better.

Oh well, there are a lot of places like that in here, and this is by far not the worst offender

thomcc (approved):

imgui/src/input/keyboard.rs
benmkw: hm would probably be best to look at cargo asm or sth. like that, not sure where the key_map is populated (seems its just internal to imgui and then accessed by casting a pointer to the Io), so its not compile time known so it may not be that much difference, just two ways of accessing the array...

thomcc: The compiler will isn't ever able to inline calls to external functions, nor is it allowed to remove them, or reorder accesses around them unless it can prove those accesses are unrelated (which it generally can't).

So, it's good to avoid external calls when you can.

norcalli:
This entire MR is incorrect. For one, the keys_down array is 512 long, whereas the key_map is only 22 long. Forcing the key_conversion index means you can't actually use any of the key_down parts of the code anymore. I don't know why such a fundamental thing slipped through...

norcalli:
This entire MR is incorrect. For one, the keys_down array is 512 long, whereas the key_map is only 22 long. Forcing the key_conversion index means you can't actually use any of the key_down parts of the code anymore.

thomcc:
I think it's fair to state that this probably inadvertently harmed some use cases (use cases where keys that imgui doesn't use are). You should still be able to access it via `io`, e.g. `ui.io().key_map[N]`.

That said, it's probably worth adding back functions to call the ImGUI functions without going through the key map.

PR #418, opened 2020-12-31
Add Listclipper wrapper

henbr:
Adds a wrapper around [ImGuiListClipper](https://github.com/ocornut/imgui/blob/fa963b9aafde7f05b1d32d42e3d2b084c9685e31/imgui.h#L2100) api.

The "log" example of the demo window is using ListClipper so it felt apropriate to add that as well, or at least enough of it to show how ListClipper works.


thomcc (approved):
CI failure seems unrelated. Thanks.

PR #429, opened 2021-01-27
Fixed broken aarch64-apple build by swapping cty::c_char with os::raw::c_char

tbogdala:
The current master doesn't build examples with arm64 rust toolchains because there is a conflict between the way std::os::raw::c_char and cty::c_char define their values. The compilation will fail with a series of errors such as these (pointing to sys:igGetVersion() and str_id.as_ptr()):

-----------

error[E0308]: mismatched types
  --> imgui/src/lib.rs:80:36
   |
80 |         let bytes = CStr::from_ptr(sys::igGetVersion()).to_bytes();
   |                                    ^^^^^^^^^^^^^^^^^^^ expected `i8`, found `u8`
   |
   = note: expected raw pointer `*const i8`
              found raw pointer `*const u8`

error[E0308]: mismatched types
   --> imgui/src/lib.rs:350:35
    |
350 |         unsafe { sys::igOpenPopup(str_id.as_ptr(), 0) };
    |                                   ^^^^^^^^^^^^^^^ expected `u8`, found `i8`
    |
    = note: expected raw pointer `*const u8`
               found raw pointer `*const i8`

...

--------------

The root of this seems to be imgui-sys/src/bindings.rs using cty::c_char values. When these values are used with functions such as CStr::from_ptr() there will be a conflicting type.

My solution to this was to replace occurrences of cty::c_char with std::os::raw::c_char at the binding level instead of using casts throughout all calling code. AFAIK, this approach doesn't have a down side.

This PR compiles with stable-aarch64-apple-darwin 1.49.0 on a new MacBook Air and examples run as expected. I also tested the PR under stable-x86_64-pc-windows-msvc without issues.

This is my first PR in a long while (possibly ever, I don't remember) and my first PR for a rust project so feel free to drop constructive criticism if I've gone about this the wrong way.

### DIFF
--- a/imgui-sys/src/bindings.rs
+++ b/imgui-sys/src/bindings.rs
@@ -2,6 +2,9 @@
 
 #![allow(nonstandard_style, clippy::all)]
 
+extern crate std;
+use std::os::raw::c_char;
+
 #[repr(C)]
 #[derive(Copy, Clone, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct __BindgenBitfieldUnit<Storage, Align> {
@@ -321,7 +324,7 @@ impl Default for ImVector_ImWchar {
 pub struct ImVector_char {
     pub Size: cty::c_int,
     pub Capacity: cty::c_int,
-    pub Data: *mut cty::c_char,
+    pub Data: *mut c_char,
 }
 impl Default for ImVector_char {
     fn default() -> Self {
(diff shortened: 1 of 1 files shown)


thomcc:
This sounds like a `cty` bug, and I'd prefer it be fixed there if possible. I also have built this code on aarc64-apple-darwin, and it seemed to work then?

thomcc:
I filed https://github.com/japaric/cty/issues/18. I'd prefer to fix it there so the sys crate works under no_std.

thomcc:
Fixing this with #442 instead. Thanks for the PR though, it's just there was a reason we weren't using those to begin with.

PR #442, opened 2021-02-05
Replace cty with chlorine

thomcc:
Justification in the comments.

Managed to ~~bully~~ convince @Lokathor into fixing this in his `cty`-alike crate, so lets use that for now since https://github.com/japaric/cty/pull/19 has gotten no response.


thomcc:
Yep, works on M1 with this.

PR #445, opened 2021-02-10
Add `DrawListMut::add_image` etc

dbr:
Exposed a few `DrawList` methods related to drawing images, they are quite useful for things like making custom controls (using `ImageButton` makes it

I personally only needed `add_image` but added the quad and rounded ones for completeness (well completeness plus the wavy Lenna button is quite entertaining)

There's a few things I questioned about these new bindings:

1. Could they be combined into one struct - my decision was nope, as you can't add rounded corners to `add_image_quad` (so some of the builder arguments would be mutually exclusive). Could have combined `Image` and `ImageRounded` and left `ImageQuad` separate, but it seemed neater just to keep each separate
2. There is already an `Image` struct at `src/widgets/image.rs` (`Image` as in `ImageButton`). I couldn't think of a good alternative name for these without being inconsistent with the other similar builders in `draw_list.rs` so I left it as-is
3. `ImDrawCornerFlags` is in crate::legacy` - is this okay? It felt slightly odd using something in a legacy folder to implement some new functionality!


thomcc:
This seems great.

Not sure about the naming vs the other widget. I don't think its good to have two with the same name... I'd almost be inclined to name the other ImageButton or something, but maybe it's better to just call this something like DrawListImage. @sanbox-irl any thoughts?

> ImDrawCornerFlags is in crate::legacy - is this okay

Yeah, I think unless there are actually deprecated parts in the C++ ImGui API it's likely worth taking that out of legacy when there isn't a replacement or an obvious downside...

sanbox-irl:
First, most importantly, the `legacy` stuff does need to get updated by hand, unfortunately, for some semblance of maintenance.
I'll write it out here, and then later add it to some future collection:

1. Take the legacy bitflags out of legacy and put it in the correct module near where it's used.
2. Redefine it to be `: u32` 
3. Redefine each flag to, instead of being a raw number itself, refer to a const from `sys::FlagName`. Fundamentally, legacy is from a time when automatically binding those constants was difficult.

For an example, I uh...mistakenly did not remove the legacy DragDropFlags out of legacy.rs recently. Will fix that immediately, but take a look at those legacy flags, and the new `drag_drop` module and see how the flags changed.

If you need extra documentation then that which is already there, try to pull it from the dear imgui lib, but feel free to add more as necessary!

Okay, now on the names:
there's `Image` *and* `ImageButton` in the widget/image.rs file, and `Image` is a struct just for images, so I think we want to protect that.

That being said, I'm not really against the local names within the modules being the same. DrawList is more or less an advanced feature, and doing `pub mod draw_list` to access these builders seems appropriate. If not, `DrawListImage` is a good call imo.

One minor note on the PR itself, and this is something that's not true in the past here -- these builders (and the ones previously made) take a long lived borrow of `DrawListMut`. Would it be possible for us to edit this PR so `draw_list_mut` is taken as a parameter in `build`? Some of imgui generally does this, and in the future, most of the functions will move towards this. I know it takes an immutable borrow, so it shouldn't matter too much, but the less we're asking users to give the draw_list, the nicer an API I suspect it'll be. 

Unfortunately, if you only did that for the new stuff, we'd have a different flavor of API versus the old one, and it'd be unfortunate to have a breaking change after only 1 version. I think either is fine depending on what @thomcc thinks.

All of that being said, this is a great PR, and all this work on the drawlist has literally made my life better (saved me from having to do it myself!) so thank you so much for putting in all this work. This looks pretty substantial!

sanbox-irl:
Looking over this more, it looks like you can remove and fix up all the flags in Legacy, and then delete that file for good. That's a nice simplification

dbr:
> Would it be possible for us to edit this PR so draw_list_mut is taken as a parameter in build?

If I understand right, you mean to remove the `DrawListMut::add_*` methods, and instead users would interact directly with `imgui::draw_list::Image` etc? For example:

```rust
imgui::draw_list::Image::new(...).build(&my_draw_list)
```

..so it is more consistent with things like the widget builders such as
```rust
imgui::Slider::new(...).build(&ui)
```

> That being said, I'm not really against the local names within the modules being the same. DrawList is more or less an advanced feature, and doing `pub mod draw_list` to access these builders seems appropriate. If not, `DrawListImage` is a good call imo.

In the current (0.7.0) API, none of the objects are ever constructed directly (`Image::new`, `BezierCurve::new` etc are private), and only accessed via `DrawListMut::add_*` methods - so I think having these be named same as `imgui::Image` is fine (e.g currently there's no reason for a user to `use imgui::draw_list::Image` so the chance for inconvenient name clashes is low)

If we change the API so draw-items structs are what users interact, then it's much more likely someone might want to

```rust
use imgui::{Image, ImageButton};
use imgui::draw_list::{Image, BezierCurve};
```

...in which case prefixing the draw list builders would make sense I think - `DrawListImage` makes sense (or maybe the slightly shorter `DrawImage`? So user "adds a DrawImage and a DrawRect to the DrawList")

> `1. Take the legacy bitflags out of legacy and put it in the correct module near where it's used.`

Ahh, I see - `legacy.rs` meant "flags done in old hard-to-maintain way" not "these are deprecated"!

This change looks pretty straight forward, I'd be happy to do this as part of this PR (since most of the flags belong to draw list, only `ImGuiInputTextFlags` moves elsewhere)

## Summary

I'll remove the remaining `legacy.rs` flags.

If @thomcc is in agreement, I can also:
1. make the draw list builders consistent with the widget builders (i.e BezierCurve::build etc takes reference to draw list, rather than calling methods on DrawListMut)
2. Prefix the draw-item builder structs (`Image`, `ImageQuad`, `BezierCurve`, `Rect` etc) to something else (probably either a) `DrawListImage`, `DrawListImageQuad` etc - or b)`DrawImage`, `DrawImageQuad` etc?)

thomcc:
> imgui::draw_list::Image::new(...).build(&my_draw_list)

Hmm, not totally sold on this. Requires more inputs increases the visibility of an API that exists mainly to work around the lack of default parameters in Rust...

sanbox-irl:
Ah, my bad, we access these as a method on `draw_list_mut`. The long borrow is fine then -- that's a better API anyway than an indendent builder. In that case, For simplicity, I'm voting on `DrawListImage`. I agree that `DrawImage` is nicer, but someone browsing the docs may confuse this for `Image`. `Image` is a very useful beginning level struct, whereas `DrawListImage` is slightly more advanced, and if accessed through `draw_list_mut` will probably not ever even need to be named by users.

I think going with `pub mod draw_list` is a good idea too, so if users *want* to find the draw list structs and create them with builders on their own, they can if they want, but since that's not the normal path, they should probably go through some hoops for it (also cleans up the autocomplete).

What do y'all think?

dbr:
Yep I agree!

Another point in favour of `pub mod draw_list` is the returned objects aren't documented otherwise:
https://docs.rs/imgui/0.7.0/imgui/struct.DrawListMut.html#method.add_bezier_curve
returns BezierCurve but searching returns nothing useful:
https://docs.rs/imgui/0.7.0/imgui/struct.DrawListMut.html?search=BezierCurve#method.add_bezier_curve

Would we would prefix all the similar structs for consistency? (or just `Image`, or all three `Image/ImageQuad/ImageRounded`?)

```
pub struct Line<'ui> {
pub struct Rect<'ui> {
pub struct Triangle<'ui> {
pub struct Circle<'ui> {
pub struct BezierCurve<'ui> {
pub struct Image<'ui> {
pub struct ImageQuad<'ui> {
pub struct ImageRounded<'ui> {
```

Also do we want to make the `Line::new` etc public to allow manual creation?

So, to summarise:

1. Make `mod draw_list` public
2. Prefix the builders
3. Maybe make the builders constructor public?

sanbox-irl:
1. Yup
2. Yup, up to @thomcc if we want to break and rename the old ones. I think for consistency we should.
3. I see no reason not to -- perhaps someone else thinks that's more elegant code. It doesn't really hurt us

dbr:
Made these changes!

I made the rename to Line/Rect/BezierCurve etc a separate commit so it can be reverted pretty trivially. I'm not too convinced about prefixing them all - it means almost everything in `imgui::draw_list` has `DrawList` as a prefix which looks a little weird.

I'd bebe  inclined to say none of them need the prefix, since they are all within the draw_list module, and it should be very uncommon for someone to directly access it anyway (since `DrawListMut::add_image` is the preferred access). Now the module is public, the rendered docs makes it pretty clear what is what (the return type now links to `draw_list::Image` docs instead of being a plain text `-> Image`)

sanbox-irl:
I agree that writing this module probably feels very silly using all those, but I'm thinking that a user who, for example runs this code:

```
let draw_image = draw_list_mut.add_image() /* etc */ ; 
```
will probably have an IDE which will either inlay the type, or on hover, show them the type `DrawListImage` which they may find helpful at a glance.

However, I'm not a strong proponent of this, so if you think that this is too awkward, I'm alright with either way

dbr:
Yeh the inlay type hints in rust-analyzer etc are a good point, but that is something that could potentially change (it could show qualified paths instead of just the struct name in future), so I reckon it's best to stick with the more concise, non-prefixed names


I've also tweaked and tidied some of the documentation strings, so I think this is mostly done. There's a spurious failure in one of the tests (rustup failing with a "Connection reset by peer" error), but ignoring that all seems good as far as I'm aware

thomcc:
Yes, I prefer the non-prefixed names. If we're going to prefix them we should do that in the future in a dedicated commit.

Can you also add a changelog entry?

dbr:
Ah I meant to ask if the changelog is generally updated in PR's or done after merging; now updated!

Reverted the `DrawListTriangle` change too, unsure how I missed that one

thomcc (approved):
This looks good to me. @sanbox-irl, is there anything else you wanted here? Otherwise I'll merge tomorrow or so.

sanbox-irl:
Ah, I never answered -- let's do the merge after the conflicts have been fixed.

Re: the changelog -- I think PRs should feel fine to edit the changelog. We obviously might edit it some more before release when it matters, but it seems to be that the author's are in the best situation to describe what they did.

dbr:
Fixed up a few merge conflicts

> but it seems to be that the author's are in the best situation to describe what they did

Agreed; the main issue I see is just that the changelog becomes an inevitible source of merge conflicts - but the conflicts are extremely easy to resolve, so it's not a big deal

sanbox-irl:
Perfect! @dbr merging this in. Thank you so much for all the hardwork!

PR #449, opened 2021-02-18
add input_text hinting

lwiklendt:
Added hinting to `InputText`, including the addition of an example in `test_window_impl`.


thomcc (approved):
Great, thanks!

PR #450, opened 2021-02-27
allow creation of ui and context objects

greaka:
Hey there,
it is impossible to use imgui-rs in combination with a backend provided by an external c application.
This PR would allow creating the necessary instances of Context and Ui.

### DIFF
--- a/imgui/src/context.rs
+++ b/imgui/src/context.rs
@@ -90,6 +90,19 @@ impl Context {
     pub fn create_with_shared_font_atlas(shared_font_atlas: Rc<RefCell<SharedFontAtlas>>) -> Self {
         Self::create_internal(Some(shared_font_atlas))
     }
+    /// Gets the current context
+    pub fn current() -> Self {
+        let ctx = unsafe { sys::igGetCurrentContext() };
+        Self {
+            raw: ctx,
+            shared_font_atlas: None,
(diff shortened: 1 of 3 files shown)


sanbox-irl:
Hi! Thanks for the PR! 

On the surface, glancing over this, it seems to me like the worry would be that these operations aren't guarded, and competing calls could easily happen. A lot of the FFI safety in imgui-rs relies on it having unique access to the Context.

I'd be okay with a function like these at the minimum guarded around some unsafety, but I'd also probably be interested in some way that's more type aware as well to reduce unsafety. 

@thomcc you have a better handle on this area than I do

greaka:
Thank you for the fast response!
I will happily adapt the PR as necessary.

Using it the way it is currently requires to never drop the created objects, or else bad things happen. So getting some guidance on how I can make it safe for other people to use is very much appreciated.

thomcc:
It's probably fine if the API for this is unsafe (e.g. `unsafe fn`s). I'll think more though, busy this weekend (through to like Tuesday or so).

greaka:
Hey there,
I just wanted to check in on this, any ideas? :slightly_smiling_face:

greaka:
I completely forgot that I had this PR open, closing because it is dead anyway.

PR #452, opened 2021-02-28
added ext variants to imgui functions

sanbox-irl:
So, I made a list of all the of the functions in the imgui.h file that had default arguments which we weren't already covering (excluding a few obscure functions which we haven't wrapped around yet, which is a separate battle for another day).

My general method was:
1. Name them with a `with_x` name as appropriate to the parameter, attempting to be short (`pos` over `position`). I used `opts` over `options` in here, so please complain away if you don't like that.
2. Use `with_x` of the *outermost* parameter when there is a function with more than 1 default argument. So for example, `same_line` really is a call to `same_line_with_pos(0.0)` and `same_line_with_pos` is really a call to `same_line_with_spacing(0.0, -1.0)`. These default parameters are noted in each "simpler" function.
3. I do not call the underlying sys::x function for each variant, but rather the more "complete" functions, so, to use the above example, `same_line` -> `same_line_with_pos` -> `same_line_with_spacing` -> `sys::igSameLine`. Because rust can inline at the crate level, we don't need to inline each of these functions and can trust that LLVM will inline that chain for `same_line` at this crate level. I did not mark these functions are inlined because these functions are high use, and I don't think users will generally want to trade compile times for tiny speed gains. If there's a disagreement on this, however, I do not feel strongly.

Okay, so, the major downside of this change is that it obviously gives TONS of breaking changes. How'd we end up?

In the example code in the repository, this change created 43 distinct errors (reported at 86, but really 43). This isn't great -- that's a lot of changes. However, merely 7 of those 43 locations actually needed the `with` variant -- or in other words, 34 function calls were simplified. Over half of these were `same_line` and `button`, which got significantly nicer.

This PR also includes a rewrite of `calc_text_size` to allow it to take `AsRef<str>` instead of `&ImStr`. This was trivially possible, and is not a breaking change, because `&ImStr` is `AsRef<str>`, but this can help some users have simpler code.

Finally, I didn't edit the changelog, since I imagine there will be some complaining about naming and what not and I'll wait to do the changelog until then.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -2,12 +2,15 @@
 
 ## [Unreleased]
 
+- BREAKING: Created `with_x` variants for most functions which previously took multiple parameters where some had default arguments in the C++. This makes calling most functions simpler and more similar to the C++.
+ - The most likely breaking changes users will see is `button` and `same_line` now take one fewer parameter -- if you were calling `button` with `[0.0, 0.0]`, simply delete that -- otherwise, call `button_with_size`. Similarly, for `same_line`, if you were passing in `0.0.` simply delete that parameter. Otherwise, call `same_line_with_pos`.
+
 - Removed legacy `ImGuiDragDropFlags` from `legacy.rs`, which were accidentally not cleared when they were remade in `drag_drop.rs` in v0.7.0.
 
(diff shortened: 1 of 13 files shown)


thomcc:
It'll be a couple days until I can take a look here, but everything you said in the description sounds good!

dbr:
This looks good to me; not having to specify these arguments is a small but very pleasant improvement to using the library!

sanbox-irl:
@dbr thank you :)) once thom takes a look, I'll merge it in

thomcc (approved):
This looks good to me!

Issue #359, opened 2020-09-08
Is there a way to use ImGui's Freetype support?


ctrlcctrlv:

dbr:
As of imgui-rs 0.7 there isn't an easy way to enable [imgui's freetype support](https://github.com/ocornut/imgui/tree/master/misc/freetype)

With imgui 1.81 (imgui 0.7 uses 1.80), the freetype integration was simplified. Once imgui-rs is updated to use >=1.81, then this might be fairly easy to do..

What I tried:

1. Add define for `IMGUI_ENABLE_FREETYPE` into `imgui-sys/build.rs`
2. Include `imgui_freetype.cpp` into `imgui-sys/src/include_all_imgui.cpp` (this also required `build.include("third-party/imgui");` so `imgui_freetype.cpp` could find `imgui.h`)
3. Add in the appropriate compiler/linker flags for freetype (I'm not sure the best cross-platform approach to this is - maybe good idea to look at the `-sys` crate for `freetype-rs` or `rust-freetype`?)
4. Add a feature flag to opt into this (there's already an example in `build.rs` to respect a feature flag in build.rs)

Then in theory that is about it, I think.. however I did all that got some compiler errors from the C++ part along the lines of
```
.../imgui-rs/imgui-sys/./third-party/cimgui/imgui/misc/freetype/imgui_freetype.cpp:147: undefined reference to `FT_New_Memory_Face'
```
..which I suspect may be to do with the order everything is getting passed to the compiler, although I didn't look into it closely

dbr:
Ohh, just realised there is some stuff in cimgui to enable Freetype support, which may help matters,

https://github.com/cimgui/cimgui/blob/ceda975a5b189e140d52aa2699811dbcbc769bb1/CMakeLists.txt#L23

dbr:
Made a PR with this: #496

PR #457, opened 2021-03-03
Update to imgui v1.81

dbr:
Mostly because I was wanting to try out the freetype support in 1.81 ("almost worked", see https://github.com/imgui-rs/imgui-rs/issues/359#issuecomment-789462561 ), I've updated the bindings to the latest imgui release.

There's one breaking change which is `ListBox::calculate_size` is gone as the imgui functions behind it have been deprecated

The more interesting change is I have shuffled around how cimgui is brought in - basically now `imgui-sys/third-party/cimgui` exists and it has a imgui (sub-)sub-module is used as the source of the `.cpp` files

This removes the need for the separate cimgui checkout and copying files around, so hopefully reduces the number of steps involved and reduces the room for errors - just cleanly checkout `cimgui` to the tagged release, run `cargo xtask bindgen`.

I think the main downside to this is someone will need to have their own `cimgui` fork in order to fork `imgui-rs` to a custom uptsream imgui version (i.e arbitrary revision or branch).. since you can't trivially commit the `generator/output/` directory anymore. Other side of this is that the imgui-rs repo no longer needs to store the intermediate generated files (only `bindings.rs`)

Thoughts?

Oh in `imgui-sys/README.md` I've documented the steps to update the bindings in future in (there's probably more steps than before, most of them are just "check the nested submodules update properly", as I'm not 100% sure how they act)


.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "imgui-sys/third-party/imgui"]
-	path = imgui-sys/third-party/imgui
-	url = https://github.com/ocornut/imgui
+[submodule "imgui-sys/third-party/cimgui"]
```
thomcc, line 1: Unfortunately, I'm planning in moving in the other direction here. I'd rather not "double-down" on submodules for now, since I think for our use cases subtrees will be better (It also should help us use stuff like other branches long term).

So, for now I'd prefer if this update stuck with the way cimgui has been managed in the past in this repo, even though it is tedious.

thomcc (changes requested):

imgui-sys/README.md
```diff
@@ -0,0 +1,96 @@
+# Updating to new imgui versions
```
thomcc, line 1: I'd prefer if docs like this should probably be in `<project-root>/docs/updating.md`.

One reason not to use the readme is that it will show up on crates.io, and so if we have one (which we should, tbh, but probably not as part of this changeset) it would be better for that to be more user-facing.

imgui/src/widget/list_box.rs
```diff
@@ -4,42 +4,22 @@ use crate::string::ImStr;
...
-            size: Size::Vec(sys::ImVec2::zero()),
+            size: sys::ImVec2::zero(),
         }
     }
-    /// Sets the list box size based on the number of items that you want to make visible
-    /// Size default to hold ~7.25 items.
-    /// We add +25% worth of item height to allow the user to see at a glance if there are more items up/down, without looking at the scrollbar.
-    /// We don't add this extra bit if items_count <= height_in_items. It is slightly dodgy, because it means a dynamic list of items will make the widget resize occasionally when it crosses that size.
-    #[inline]
-    pub const fn calculate_size(mut self, items_count: i32, height_in_items: i32) -> Self {
```
thomcc, line 36: Hmm, if it's deprecated in imgui 1.81, maybe we should just deprecate it and remove in a future version? Unless there's a compelling reason to just remove it outright

thomcc (commented):
Couple more smaller issues. Overall thanks for doing this, it looks good aside from the submodule thing, which I probably should have better communicated somewhere.

imgui-sys/README.md
dbr: Ohh, yes, I completely forgot imgui-sys gets published like that! Shall move update and move this doc to that folder, and maybe put a placeholder imgui-sys README in its place

imgui/src/widget/list_box.rs
dbr: Main reason I removed it was the binding function disappeared, and I didn't put much thought into why..

Looking into it now, cimgui defines `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` by default so you don't get obsolete functions.

Thoughts?

thomcc: Fair enough. We'll need to remember to mention it in the changelog, then.

.gitmodules
dbr: Yep fair enough, done!

thomcc (approved):
Looks good, can you add a changelog entry?

dbr:
Oh, of course - added!

Issue #456, opened 2021-03-01
add original imgui names as doc aliases to functions 


benmkw:
same as https://github.com/4bb4/implot-rs/issues/11

https://blog.rust-lang.org/2020/11/19/Rust-1.48.html#adding-search-aliases

> An interesting use case for aliases is FFI wrapper crates, where each Rust function could be aliased to the C function it wraps. Existing users of the underlying C library would then be able to easily search the right Rust functions!

This is busy work but might help people to get startet/ find what they are looking for when porting cpp imgui code.

sanbox-irl:
wow, great idea! this is however, non-trivial in practice, though welcome. Our big tasks this release are, hopefully, completing the drawlist wrapper, make our bindings to optional arguments simpler, nicer tokens for state on the stack, and removing im_str! from the codebase. So things are a bit busy right now.

If someone wants to help though, your contribution would be more than welcome (please comment and say "i'll take these modules", since this is probably a bigger task than a single person wants to do and we can divide the codebase up):

1. Functions on the `Ui` struct are highest priority. Those should be aliased to their underlying FFI call, which should be obvious in practice what it's referring to (most of them as just wrappers around their call). For example:
```rs
    pub fn separator(&self) {
        unsafe { sys::igSeparator() }
    }
```
should be bound like:
```rs
    #[doc(alias = "Seperator")]
    pub fn separator(&self) {
```

Note the capitalization -- ImGui functions are PascalCase. If our goal is that someone can search imgui functions in docs.rs and find the bindings, then we should follow that to make it useful.

As a minor note, depending on when this task is accomplished, there may be some wrapper functions which handle default arguments. If it's possible (I'm not sure) to have multiple functions have the *same* search alias, then all should have the same search alias. If now......leave a comment and we can handle it on a case by case basis. Probably the shortest version should be bound, since it's easy to walk up from the shortest to the longest version, while the functions will all parameters in them and no default arguments are not as obviously related to their overriden cousins. 

2. Builders:

So a huge category of this API is making builders to emulate extensive default arguments / immediate mode stack building. The Window struct is a great example of this. 

Binding the aliases to these functions could prove very tricky. Ideally, the builder functions which correspond to underlying usage should be flagged with their parent function, but that might even be more confusing. Take for example:

```rs
    #[inline]
    pub fn position(mut self, position: [f32; 2], condition: Condition) -> Self {
```
[...]

PR #458, opened 2021-03-04
Add doc aliases with ImGUI function names

toyboot4e:
Quick (and dirty) additions for #456

* I skipped `token.rs` and `drag_drop.rs`

### DIFF
--- a/imgui/src/context.rs
+++ b/imgui/src/context.rs
@@ -79,6 +79,7 @@ impl Context {
     /// # Panics
     ///
     /// Panics if an active context already exists
+    #[doc(alias = "CreateContext")]
     pub fn create() -> Self {
         Self::create_internal(None)
     }
@@ -87,10 +88,12 @@ impl Context {
     /// # Panics
     ///
     /// Panics if an active context already exists
+    #[doc(alias = "CreateContext")]
     pub fn create_with_shared_font_atlas(shared_font_atlas: Rc<RefCell<SharedFontAtlas>>) -> Self {
         Self::create_internal(Some(shared_font_atlas))
     }
     /// Suspends this context so another context can be the active context.
+    #[doc(alias = "CreateContext")]
     pub fn suspend(self) -> SuspendedContext {
         let _guard = CTX_MUTEX.lock();
         assert!(
@@ -183,10 +186,12 @@ impl Context {
         self.renderer_name = renderer_name;
     }
     /// Loads settings from a string slice containing settings in .Ini file format
+    #[doc(alias = "LoadIniSettingsFromMemory")]
     pub fn load_ini_settings(&mut self, data: &str) {
         unsafe { sys::igLoadIniSettingsFromMemory(data.as_ptr() as *const _, data.len()) }
     }
     /// Saves settings to a mutable string buffer in .Ini file format
+    #[doc(alias = "SaveInitSettingsToMemory")]
     pub fn save_ini_settings(&mut self, buf: &mut String) {
         let data = unsafe { CStr::from_ptr(sys::igSaveIniSettingsToMemory(ptr::null_mut())) };
         buf.push_str(&data.to_string_lossy());
@@ -236,6 +241,7 @@ impl Context {
 }
 
 impl Drop for Context {
+    #[doc(alias = "DestroyContext")]
     fn drop(&mut self) {
         let _guard = CTX_MUTEX.lock();
         // If this context is the active context, Dear ImGui automatically deactivates it during
@@ -270,6 +276,7 @@ pub struct SuspendedContext(Context);
 
 impl SuspendedContext {
     /// Creates a new suspended imgui-rs context.
+    #[doc(alias = "CreateContext")]
     pub fn create() -> Self {
         Self::create_internal(None)
     }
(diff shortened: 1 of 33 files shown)


imgui/src/context.rs
```diff
@@ -87,10 +88,12 @@ impl Context {
     /// # Panics
     ///
     /// Panics if an active context already exists
+    #[doc(alias = "CreateContext")]
     pub fn create_with_shared_font_atlas(shared_font_atlas: Rc<RefCell<SharedFontAtlas>>) -> Self {
         Self::create_internal(Some(shared_font_atlas))
     }
     /// Suspends this context so another context can be the active context.
+    #[doc(alias = "CreateContext")]
```
sanbox-irl, line 96: Lets not have this doc comment here... It's not exactly accurate

imgui/src/context.rs
```diff
@@ -270,6 +276,7 @@ pub struct SuspendedContext(Context);
 
 impl SuspendedContext {
     /// Creates a new suspended imgui-rs context.
+    #[doc(alias = "CreateContext")]
```
sanbox-irl, line 279: For now, let's keep the SuspendedContext out of the doc comments. Its current implementation is wrapped up in our own single-threaded safety guarantees, and not necessarily in ImGui's create context. I don't want to confuse someone who searched "CreateContext" and sees something like this in docs.rs

imgui/src/context.rs
```diff
@@ -498,6 +508,7 @@ impl Context {
     /// # Panics
     ///
     /// Panics if the context uses a shared font atlas that is already borrowed
+    #[doc(alias = "NewFame")]
```
sanbox-irl, line 511: Probably "Frame", but we're all searching for "Fame" I understand

imgui/src/style.rs
```diff
@@ -153,19 +153,22 @@ unsafe impl RawCast<sys::ImGuiStyle> for Style {}
 
 impl Style {
     /// Scales all sizes in the style
+    #[doc(alias = "ScaleAllSizes")]
     pub fn scale_all_sizes(&mut self, scale_factor: f32) {
         unsafe {
             sys::ImGuiStyle_ScaleAllSizes(self.raw_mut(), scale_factor);
```
sanbox-irl, line 159: Why do we bind these differently? This looks right though

imgui/src/widget/color_editors.rs
```diff
@@ -192,6 +192,7 @@ pub struct ColorEdit<'a> {
 
 impl<'a> ColorEdit<'a> {
     /// Constructs a new color editor builder.
+    #[doc(alias = "ColorEdit3", alias = "ColorEdit4")]
```
sanbox-irl, line 195: Hahaha, this really shows how much we need to update this. We now have a significantly improved color struct which we'll probably use instead of this in the near future; how odd these aliases look reflects how odd the binding is.

imgui/src/widget/text.rs
```diff
@@ -36,14 +37,17 @@ impl<'ui> Ui<'ui> {
         style.end();
     }
     /// Renders text wrapped to the end of window (or column)
+    #[doc(alias = "TextWrapperd")]
```
sanbox-irl, line 40: Probably "Wrapped" instead of "Wrapperd" although "Wrapperd" has some linguistic advantage -- what is a wrapper is not something where wrapperd's things

sanbox-irl (changes requested):
Overall, looks great.

There are a couple of typos which should be fixed before merging, but then I'll add it right in. There will likely be another commit landing before this one is merged in (unless you fix the typos very fast), so please rebase this commit onto master after you fix the typos. There will be some new functions there (specifically, this PR is about to land https://github.com/imgui-rs/imgui-rs/pull/452#pullrequestreview-603754852) but I will add doc aliases for those new functions too. In the future, we'll have to remember to check for doc-aliases in the future.

All that being said though, this is a heroic PR. Thank you very much! I cannot imagine what sort of Zen state you entered to churn through effectively the entire surface of the library. It is appreciated!

imgui/src/context.rs
toyboot4e: hehe

imgui/src/style.rs
toyboot4e: Are you referring to these alises?
```rust
    #[doc(alias = "StyleColors", alias = "StlyeColorsClassic")]
```
If so, it was because aliases are searched with (case-insensitive) exact match. But such aliases don't work well after all, so I'll remove them.

toyboot4e:
Rebased! While rebasing, I added aliases for new coming functions.

imgui/src/style.rs
sanbox-irl: Oh no, that's all good! lol no I shouldn't have just been musing -- these are good. Sorry lol, even just *reviewing* all of this my mind started wandering. That you wrote them all was impressive.

sanbox-irl:
Looks fantastic. Thank you SO much for all this effort! Do you mind if we shout you out in the changelog? We're going to start tagging people who contributed

toyboot4e:
No, if I'm mentioned, I'm deeply honored. Thank you.
Btw, I think it is your issue comment and code review that is awesome :)

PR #461, opened 2021-03-06
(WIP) First cut at removing ImStr from the API.

thomcc:
Ulimately we're going to still transition to the imgui sized str branch here, but this means if there's some reason that doesn't pan out or can't be used universally (if it doesnt mix with docking or other branches or w/e) then it doesn't actually impact the API.

Concretely this version copies strings to a smallvec<[u8; 128]>, in order to NUL terminate them without allocating in common cases.

This has a nice benefit: it having this buffer means we can take allow users to pass in e.g. `format_args!(...)` directly, and we can use it as the scratch space, conveniently avoiding allocations for cases where you dynamically build the text on a widget, which is nice.

# Basic idea

- all functions taking &ImStr to move to `trait AsImStr`.
- `trait AsImStr` is implemented for all `&T: AsRef<str> + ?Sized` (e.g. references to AsRef<str> types — this handles a lot of things, including &str, &String, &ImStr, &ImString, ...
- `trait AsImStr` is also implemented for `fmt::Arguments` (not as a reference).
- (not done yet) `ImStr`/`ImString` are deprecated, but will be kept around (possibly with a much smaller API surface) since y'all don't update shit ever (we still support winit 0.19 in this code... 😭). 

The long term goal is for us to transition (possibly behind a feature) to the `string_view` branch, but this change is intended on making whether or not we use that branch an implementation detail that doesn't matter too much (it's pretty cheap either way).

# Open questions:

1. I'd like a better name than AsImStr. If ImStr goes away, why should this be called that? This ends up showing up in a *lot* of our function signatures, so we should try to do a good job here.

2. Needs user-facing docs. I'd like to solve the other questions here first tho to avoid writing them multple times.

3. Docs on the implementation stuff need cleaning up a lot

    - I kind of half-explained how the macro and such works in like 4 places in the hope that one of them stuck.
    - It's not even that complex — it's just weird looking and almost only happens inside a macro.

4. We could probably generalize to implementations of `Display` rather than AsRef<str> + format_args. Should we?

    - This is more flexible... Or, it looks that way. I think in reality, it doesn't give you anything that `format_args` doesn't also give you, but it might give you that in a cleaner way that's likely to be understood by a larger portion of Rust users...

    - Con: ommitting to this might make it hard to get benefits from the `string_view` branch (we'd need specialization).

    - Con: It also might be confusing if you pass an integer in an argument that takes a string, and it's accepted because it's Display-able, or w/e...

    - Pro: it likely helps solve the issue around "this shows up in all our func sigs, so it would be nice if users had a clue what it was" (how many APIs take a `format_args!` directly?)

    - More generally: there something other than Display that's better for our case? (Note that we're still limited by coherence — it's very easy to come up with a set of things that would be nice if they worked, but they do not)

5. Is it worth the trouble at all? I think yes, but @sanbox-irl was skeptical, and has more experience with using imgui from Rust, so maybe this doesn't actually help.

6. Does this hit any unanticipated issues with type inference? If it tends to fail to resolve, that would defeat the point, largely...

7. ...

Also, note that I only translated a few parts, @sanbox-irl indicated he might take the rest up but no rush, I might get the rest of them later today, it also might be worth deciding some of the above questions first. ~~That's why this is a draft PR tho.~~ (nm, clicked wrong button)


sanbox-irl:
I think it's worth it. I'll need to take a closer took tomorrow at all of this, but the idea is thrilling! lol a world of imgui-rs WITHOUT im_str. A magic world

thomcc:
> Is it worth the trouble at all

Sorry, this wasn't referring to doing this at all, but the approach I took in the patch (as opposed to just accepting, say, `impl AsRef<str>` everywhere). Sorry, that was super unclear.

sanbox-irl:
No, it was! I actually meant, now that I'm looking at this, I think it's worth the effort to accept `format_args`. The notion of that becoming free now that it's infront of me has made me very greedy

dbr:
I don't have any strong opinions on the implementation, but if this means `ui.button(im_str!("This"))` becomes `ui.button("This")` then.. :partying_face:

>  I'd like a better name than AsImStr. If ImStr goes away, why should this be called that?

I am very likely missing some detail, but.. if ImStr goes away, wouldn't `AsImStr` also be deleted?

As in on the mystical day when ImString goes away, I imagine the function signature would become `fn button(label: &str) -> ... {` or something along those lines?

sanbox-irl:
ahh fair. we've had some discord conversations that have gone over it. here's the thought:

we occasionally have some label which is a formatted string. For example, let's say you're outputting this button:

`if ui.button(&format!("Entity {}", entity_id)) {}`

This forces an allocation per call by the user. The current `im_str` macro actually does this!

However, we realized here, that we can actually prevent that for a significant number of short strings (a very long string will need to be allocated stuff) by using the `format_args!` macro:

``if ui.button(format_args!("Entity {}", entity_id)) {}``

When thom says above that I was initially against this complexity, that's because I thought most people never would know about `format_args` (it's mostly used in macros). So in the above, right now, the big win is that we never actually need to allocate that string -- we just write it into our backing buffer directly, and then hand that off to imgui like normal.

So we need a trait like `AsImStr` that we use in the parameter, so we can take `&str`, probably `String` (though that's a little suboptimal though would nice to take), and the output of `format_args`

What do you think about all that @dbr ?

aloucks:
The [`As`](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) naming convention is generally considered to be "free". From a naming perspective, I think it should be `IntoImStr`.

Although I'm strongly in favor is moving away from `im_str!` / `ImStr` / `ImString`, I don't know if using `SmallVec` here is ideal. It should prevent most allocations with labels, but the text functions will likely allocate frequently.

I think a better solution is to just keep a cache of text buffers in the `Context` (and could possibly be reset per frame or at the users request to reclaim the memory). 


```rust
#[derive(Debug, Default)]
struct TextBuffers {
    // Assumes no imgui function needs more than 5 strings in use at the same time
    buf: [Vec<u8>; 5],
}

impl TextBuffers {
    /// Create a null-terminated string from `text`. The pointer remains valid until the next
    /// call to `c_str` with the same `index`, the `TextBuffers` is dropped, or `reset`
    /// is called.
    fn c_str(&mut self, index: usize, text: &str) -> *const std::os::raw::c_char {
        fn set_text(buf: &mut Vec<u8>, text: &str) -> *const std::os::raw::c_char {
            buf.clear();
            buf.extend_from_slice(text.as_bytes());
            buf.push(0);
            buf.as_ptr() as *const _
        }
        set_text(&mut self.buf[index], text)
    }

    fn reset(&mut self) {
        for buf in self.buf.iter_mut() {
            *buf = Vec::new();
        }
    }
}
```

thomcc:
Into's not totally accurate either, and long term the goal is to switch to the [`string_view`](https://github.com/ocornut/imgui/tree/features/string_view) branch, which would make the "As" name more accurate, and remove the allocations (and copies) 

However, unless the string_view branch becomes the only branch, there's no way for us to (efficiently) support stuff like:

- linking to custom builds of Dear ImGui
- linking to other freature branches, such as the Docking branch
- etc

Which we'd like to support, unless we come up with an API that can work in either.

A buffer cache isn't the worst idea though, I'll think about it. It's worth noting that the Text apis already don't work ideally with this design, which only really works for immutable string inputs, so that's why not much progress has been made, need to come up with a solution for it.

norcalli:
Can there be an escape hatch to allow users who don't need hand holding so that they can pass pre-checked strings themselves? I'd rather just pass strings that I format from a bump allocated arena. And the worst thing that happens if I forget a null terminator is probably a segfault, and I'm entirely willing to live with that possibility. It really doesn't scare me. At least allowing the users to make the choice to take care of things themselves in addition to an easier option where things are "just handled" is a generally a good idea IMO.

We use the imgui library extensively at work, so this would be fairly useful for us.

Side note, is there a chance of stack overflow from having a bunch of SmallVecs be created inside a string heavy function?

thomcc:
Note: It's unlikely this is the exact approach I'll take — I think the idea of using a scratch buffer is a better idea.

That said:

> Side note, is there a chance of stack overflow from having a bunch of SmallVecs be created inside a string heavy function?

No, because the stack will be cleared after each one is used.

---

Also, even in the current version of the patch, IIRC, the smallvec isn't used if you pass in an already nul-terminated string (Well, I'm *pretty* sure this is the case — it should be, but it's been a while since I looked at this code, and can't at the moment). This property will still be true if/when I change it to use cached buffer, which would allow the zero cost solution you're describing.

We'd still check for the nul-terminator for you, but that's literally trivial — checking the last byte of a slice.

Anyway, I have planned getting back to this and some other open source stuff next week, after I'm done recovering from vaccine 2.

norcalli:
Actually, I thought about this again, why not just retain a single ImString and write into that rather than use a SmallVec? imgui copies the string immediately anyway. I'm about to do this myself as a workaround in the meantime.

thomcc:
This is what I mean by using a cached buffer.

norcalli:
You're commenting too quick for me, lol.

Issue #455, opened 2021-03-01
Why only part of keyboard inputs provided?


Danielmelody:
```rust
#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
pub enum Key {
    Tab = sys::ImGuiKey_Tab,
    LeftArrow = sys::ImGuiKey_LeftArrow,
    RightArrow = sys::ImGuiKey_RightArrow,
    UpArrow = sys::ImGuiKey_UpArrow,
    DownArrow = sys::ImGuiKey_DownArrow,
    PageUp = sys::ImGuiKey_PageUp,
    PageDown = sys::ImGuiKey_PageDown,
    Home = sys::ImGuiKey_Home,
    End = sys::ImGuiKey_End,
    Insert = sys::ImGuiKey_Insert,
    Delete = sys::ImGuiKey_Delete,
    Backspace = sys::ImGuiKey_Backspace,
    Space = sys::ImGuiKey_Space,
    Enter = sys::ImGuiKey_Enter,
    Escape = sys::ImGuiKey_Escape,
    KeyPadEnter = sys::ImGuiKey_KeyPadEnter,
    A = sys::ImGuiKey_A,
    C = sys::ImGuiKey_C,
    V = sys::ImGuiKey_V,
    X = sys::ImGuiKey_X,
    Y = sys::ImGuiKey_Y,
    Z = sys::ImGuiKey_Z,
}
```
where is the key `B`, `D` ... and everything else?

sanbox-irl:
Those are the keyboard inputs which imgui *itself* tracks. You set up your keyboard bindings which allow you to ferry keyboard inputs throughout your application using the keymaps API. Check out the winit platform implementation, and see how it handles keys.

Effectively, those are just the keys which imgui needs to operate (it doesn't actually ever listen to `b`, `d` or anything else!)

dbr:
I ran into the same difficulty today, trying to make a shortcut as obscure as `ctrl+q`

## History

In imgui-rs 0.6 you could do

```rust
if ui.is_key_down(winit::event::VirtualKeyCode::Q as u32) {
    println!("Cool");
}
```
..assuming you were using the `imgui-winit-support` or similar.

In imgui-rs 0.7.0 was a change (#416) which made this harder, as the key functions take `imgui::Key` directly - which I think is a better API, but is currently limited to a very small subset of keys (A/C/V/X/Y/Z etc as mentioned)

## Workaround

[...]

PR #463, opened 2021-03-09
Add back "_index" versions of all the keyboard related methods

dbr:
As discussed in #455 - this restores the ability to query keys by arbitrary, backend-defined indexes which is currently very necessary as `imgui::Key` only contains a few keys.

When this is expanded further, querying keys by index may still be necessary in some cases for less common key codes

The existing (v0.7.0) `is_key_pressed` API is retained as the default, and `is_key_index_pressed` has been added - so I don't think this is a breaking change.

### DIFF
--- a/imgui-examples/Cargo.toml
+++ b/imgui-examples/Cargo.toml
@@ -16,3 +16,7 @@ image = "0.23"
 imgui = { path = "../imgui" }
 imgui-glium-renderer = { path = "../imgui-glium-renderer" }
 imgui-winit-support = { path = "../imgui-winit-support" }
+
+# Direct dependency only required because `winit::VirtualKeyCode` is
+# used for one example.
+winit = "*"
--- a/imgui-examples/Cargo.toml
+++ b/imgui-examples/Cargo.toml
@@ -16,7 +16,3 @@ image = "0.23"
 imgui = { path = "../imgui" }
 imgui-glium-renderer = { path = "../imgui-glium-renderer" }
 imgui-winit-support = { path = "../imgui-winit-support" }
-
-# Direct dependency only required because `winit::VirtualKeyCode` is
-# used for one example.
-winit = "*"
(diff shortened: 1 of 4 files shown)


sanbox-irl (approved):
Looks great. will wait on @thomcc to check

imgui-examples/Cargo.toml
```diff
@@ -16,3 +16,7 @@ image = "0.23"
 imgui = { path = "../imgui" }
 imgui-glium-renderer = { path = "../imgui-glium-renderer" }
 imgui-winit-support = { path = "../imgui-winit-support" }
+
+# Direct dependency only required because `winit::VirtualKeyCode` is
+# used for one example.
+winit = "*"
```
thomcc, line 22: glob import will be a problem here... kind of a pain to fix, for now please just hardcode the constant winit uses for it.

thomcc (changes requested):

imgui-examples/Cargo.toml
dbr: Not sure I understand - you mean pin this to, say, `winit = "0.24.0"` ?

thomcc: No, remove it. It will prevent us from actually running tests on different winit versions. The way version resolution works, this isn't actually a wildcard like you'd think.

I think we have two choices here: either we figure out what value `VirtualKeyCode::Home as i32` is and hardcode that, or reexport winit from imgui-winit-support. My gut was to do the first one, but the 2nd is probably reasonable too, although probably I'd prefer we `#[doc(hidden)]` it, since people should probably depend on it explicitly.

dbr: Ahh, I see! That makes sense. I'll just hardcode it - if we use the re-export in examples hiding in docs wont stop copy-pasting!

dbr:
Removed the example's direct dependency on `winit`, and added a changelog entry for this

sanbox-irl:
@dbr @thomcc as far as I can tell, we're good to merge here, right?

thomcc (approved):
Sorry, missed this

sanbox-irl:
thanks @dbr !

PR #468, opened 2021-03-26
Modifies build functions to allow returning data from the build closures.

### LINKED ISSUE #467: Build method closures returning a user-specified type rather than ()
Often times when I'm using the `.build(ui, || { ... })` APIs, I want to return data to the outside scope.  It's kind of inconvenient as it is right now because I have to create a mutable variable outside the scope.  I'm wondering if we could consider changing the return type of these closures to something more usable to the user.

Here's an example re-definition for the `build` method on PopupModal, to explain what I mean:
Today:
```Rust
pub fn build<F: FnOnce()>(self, ui: &Ui<'_>, f: F) {
    if let Some(_popup) = self.begin_popup(ui) {
        f();
    }
}
```
Suggestion:
```Rust
pub fn build<F: FnOnce() -> T, T>(self, ui: &Ui<'_>, f: F) -> Option<T> {
    self.begin_popup(ui)
        .and_then(|_popup| f())
}
```
Alternative:
```Rust
pub fn build<F: FnOnce() -> T, T>(self, ui: &Ui<'_>, f: F) -> Option<T> {
    if let Some(_popup) = self.begin_popup(ui) {
        Some(f())
    } else {
        None
    }
}
```

AngelOfSol:
Resolves #467

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+- BREAKING: Modifies `build` style methods to allow the provide closure to return a value.  The build call will then return Some(value) if the closure is called, and None if it isn't.
+ - The most likely breaking changes users will see is that they will need to add semicolons after calling `build`, because these function no longer return `()`.
+
 - BREAKING: Created `with_x` variants for most functions which previously took multiple parameters where some had default arguments in the C++. This makes calling most functions simpler and more similar to the C++.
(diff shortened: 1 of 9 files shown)


thomcc (approved):
This looks fine to me, but I'd like @sanbox-irl to weigh in since he missed most the discussion afaict.

Also, can you add a note about this to the changelog?

AngelOfSol:
Will do, and I'm not in a rush to get this merged!  

It looks like there some doc errors that are causing the tests to fail.

sanbox-irl:
Okay this looks good. I'll push a fix to the doc comments (nice catches n the beta branch -- nice that running CI on them helped fix errors!) after accepting this for simplicity.

@AngelOfSol thanks so much! I'm going to take a look at your other PR today too

PR #469, opened 2021-03-26
Adds callbacks from rust.

### LINKED ISSUE #430: Callback input flags for Text/Floats/etc are more or less useless right now
Hi all,

I'm planning on spending some effort rewriting the inputflags system in the next few days. 

On TextInput, we allow users to indicate that they would like a callback with a given flag, but, somewhat hilariously, we have not implemented a way to provide a callback (*wah wah*), so you can be like "i would like a per char callback", but can't actually give it one, and Dear ImGui apparently just ignores your flag then. There are some comments which indicate a boxed trait object approach, but I suspect just a mutable pointer generic over a trait will be sufficient (I do not think imgui needs to hold onto this struct, but it may, so I'll look into it).

If anyone else is working on this/had some PR to merge, lmk so I don't spend the effort unnecessarily!

AngelOfSol:
This is a WIP pull request implementing text edit callbacks in rust.

Current status:
~~I got imgui to call our basic callback, and I'm working on integrating a more rusty API for the end user.~~
I've started working on the end-user API, but I'm stuck on how we should allow the user to manipulate the ImString buffer.  I think it might be worth while to provide helper methods that do work the user could do themselves, but are pretty repetitive.

This should eventually resolve #430

### DIFF
--- a/imgui/src/input_widget.rs
+++ b/imgui/src/input_widget.rs
@@ -1,10 +1,12 @@
 use bitflags::bitflags;
-use std::marker::PhantomData;
 use std::os::raw::{c_int, c_void};
 use std::ptr;
+use std::{convert::TryFrom, marker::PhantomData};
 
 use crate::sys;
 use crate::{ImStr, ImString, Ui};
+
+/// @angelofsol -- notice the "CALLBACK_EDIT" flag that I added too. We need to cover that also...probably.
 
 bitflags!(
     /// Flags for text inputs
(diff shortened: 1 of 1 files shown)


imgui/src/input_widget.rs
```diff
@@ -199,6 +199,297 @@ extern "C" fn resize_callback(data: *mut sys::ImGuiInputTextCallbackData) -> c_i
...
+            }
+            true
+        } else {
+            false
+        }
+    }
+
+    pub fn event_char(&self) -> char {
+        // pretty sure EventChar is utf16 but i need to double check
+        char::decode_utf16(std::iter::once(unsafe { (*self.inner).EventChar }))
```
thomcc, line 267: std::char::decode_utf16 works here without the unstable feature.

AngelOfSol: Fixed

sanbox-irl:
Closing in favor of #516

PR #471, opened 2021-04-05
Update Dear Imgui to v1.82

thomcc:


sanbox-irl (approved):
looks good to me. nice that we largely dodged all the breaking changes

PR #473, opened 2021-04-05
Support 32 bit unicode (-DIMGUI_USE_WCHAR32). (Also, teach xtask how to wrangle submodules)

### LINKED ISSUE #424: Full unicode range support
Currently in imgui-rs (version 0.6) there is support for a subset of the range to `0xFFFF`, with some ranges predefined in [`FontGlyphRanges`](https://docs.rs/imgui/0.6.1/imgui/struct.FontGlyphRanges.html) as discussed in https://github.com/imgui-rs/imgui-rs/issues/157

Introduced in https://github.com/ocornut/imgui/pull/2815 (and [discussed here](https://github.com/ocornut/imgui/issues/2538)) there is support full 32-bit unicode range, enabled by a `IMGUI_USE_WCHAR32` define. This allows a much larger set of characters (including things like emoji and so on)

Seems like this should be a relatively simple change to implement into `imgui-rs` - I tried literally just shoving in the `IMGUI_USE_WCHAR32` define to `imgui-sys/build.rs` and it "almost worked", but aborted when displaying a test window and I didn't look into debugging it yet

This is something we may need at some point in the nearish future, at which point I'll look into it further and hopefully make a PR for it - but making this ticket now in case the information is of use to anyone else

thomcc:
The unicode stuff in the API is a little tricky since most people use codepoint even if what they mean is unicode scalar value, and the distinction means whether or not we can use `char`.

It also was hard to write the changelog. In theory we're allowed to change the flags since we provide ways of accessing them, but we've changed them rarely enough that its worth mentioning.

Rerunning bindgen and stuff caused some weird changes, though, but since I ran it last, probably my fault.

The submodule bit was because i did this in a work tree, and forgot to fixup the submodule. It's taken mostly from my game engine's xtask, and is extremely overkill for now, but it's plausible that we'll get more submodules, in which case it won't be. It really should be a separate PR, but whatever. I also futzed with the docs on the flags some, etc.

Fixes #424.


imgui-sys/third-party/cimgui_impl.h
```diff
@@ -1,38 +0,0 @@
-typedef struct SDL_Window SDL_Window;
```
thomcc, line 1: Unsure what the story here is but we don't provide the functions these correspond to in the static lib our -sys crate builds, nor do we expose them as rust functions, so I think it was just a mistake or cimgui quirk.

thomcc:
Shoot, that's what i get for using nightly.

PR #474, opened 2021-04-16
Initial Tables API

cfrantz:
This is a naive and incomplete binding to the ImGui Tables API.

1. APIs are thin rust wrappers around their C++ counterparts.  I believe I've followed the naming convention for overloads.
2. A couple of Table APIs are not implemented yet (I didn't have immediate need for them and they weren't dead-trivial to implement).
3. `setup_columns_with_id` uses an `Id` function from my `docking2` branch.
4. I've not yet written doc comments.

This draft PR is to get a review on the general approach for the Tables API.  In my own experiments, I like the look of Tables much better than what I can achieve with the Columns API:
[Example](https://github.com/cfrantz/z2edit/blob/tables/resources/z2e2-enemyattr.png)

Signed-off-by: Chris Frantz <frantzcj@gmail.com>

PR #476, opened 2021-04-17
Initial Docking API

cfrantz:
This is a naive and incomplete binding to the ImGui Docking API.

1. APIs are thin wrappers around their C++ counterparts.
2. Some of the wrappers are not implemented.
3. I haven't written doc-strings for anything yet.

I've tried to structure this as a series of commits which separate the bindings from the code-base preparation steps (ie: cimgui updates, bindgen, etc).  Similar to https://github.com/imgui-rs/imgui-rs/pull/474, this draft PR is to solicit feedback on the approach and shape of the docking bindings.


sanbox-irl:
Hello! This is very interesting though we currently aren't planning on switching the repo over to track the docking branch. There might be future plans to do something for the docking branch (perhaps our own long lived branch to mirror it, perhaps a cargo feature if we can get away with that), but for now I don't think this PR, even in perfect shape, could be something we could accept

dbr:
Personally I'd really like if imgui-rs could have support for docking enabled via feature flag (mainly because it'd save me having to rebase my own similar changes atop each release!)

That said, I just noticed the docking PR has been put under a release milestone, which means.. something, probably: https://github.com/ocornut/imgui/milestone/14 - however it's 1.90 (and current release is 1.82, and at roughly 1 release per month would put it somewhere early next year maybe)

There was an older PR for adding docking support, https://github.com/imgui-rs/imgui-rs/pull/285 (and a few other similar forks mentioned in there)

In terms of wrapping the dock-builder API, I found the approach I used [here](https://github.com/dbr/imgui-rs/blob/969d45d4636dd4b289a2f8062386012fd229666d/imgui/src/docking.rs) worked quite well (which was mostly stolen from another fork I've somehow lost track of currently?! Main change was to use `im_str!` more for ID's, which was more consistent with the C++)

```rust
let space = ui.dockspace(im_str!("MainDockArea"));
space.split(
    imgui::Direction::Left,
    0.7,
    |left| {
        left.split(
            imgui::Direction::Left,
            0.6,
            |sub_left| {
                sub_left.dock_window(im_str!("Left"));
            },
            |sub_right| {
                sub_right.dock_window(im_str!("Middle"));
            },
        );
    },
    |right| {
        right.dock_window(im_str!("Tab in right column"));
        right.dock_window(im_str!("Another tab"));
        right.split(
            imgui::Direction::Down,
            0.4,
            |down| {
                down.dock_window(im_str!("Small window at bottom"));
            },
            |_up| {},
        );
    },
);

imgui::Window::new(im_str!("Left")).build(...etc..);
```

This seems pretty flexible - you can either create a dockspace over the entire viewport, or within a window, then dock and undock windows to it as needed. It also worked with imgui's built-in settings thing for saving window layouts.

It did need some work as there was still some things you could easily do with the API which would make imgui fail some asserts, but for the most part the above has been working nicely for me

sanbox-irl:
closing in favor of https://github.com/imgui-rs/imgui-rs/pull/538

Issue #462, opened 2021-03-09
Tracking Issue for v0.8.0


sanbox-irl:
Hello!

Over the past month, we've all been working away on v0.8.0 of the repository. I'm making an issue here (and pinning it) so people can see what work has been done, what work is ongoing, and what work might be useful to do. I will try, in the followup comments, to provide some direction on issues. We're also planning on having some breaking changes in the next few releases (and obviously the possibility of more after, but these next two are gonna be larger).

## Work Done

1. Added permissive dropping to the library, and restructured drop tokens in general to have a lifetime but not be as punishing to misuse. Now, instead of panicking when a token (such a "TabBarToken") is dropped without running `.end` on it, we'll just run `.end` in the drop code if necessary (if you manually call .end, that simply drops the token early), invoking the drop code. This will produce significant breaking changes, but they should be trivial to fix. This change was merged <https://github.com/imgui-rs/imgui-rs/pull/440>
2. Add `DrawList::add_image` and other functions. Allows for the draw image in the draw list functions to be used. Thank you @dbr for that! This is going to be very nice for *me* in particular :)) (I use a tilemap editor with your drawlist functions, and add image will be very nice!). This PR landed here https://github.com/imgui-rs/imgui-rs/pull/445
3. Added `_with_x` variants for most functions which take default arguments in C++. So for example, `same_line` previously took a float, but now `same_line` takes no parameters, but `same_line_with_pos` has been added. This should make `same_line` and `button` significantly less annoying to use. This was PRed here: https://github.com/imgui-rs/imgui-rs/pull/452
4. A few smaller commits: we added doc aliases to all of the functions for their original imgui counterparts, which should make finding imgui-rs equivalents to the imgui_demo.cpp file a lot easier. Thanks a lot @toyboot4e . Additionally, we added some input hinting for the input_text module (more on that below). thanks for that @lwiklendt. And finally, @dbr once again being excellent, upgraded to imgui v1.81, which was a smaller release than v1.80.

## Work in Progress:
[...]

PR #477, opened 2021-04-30
Disable winit default features

dzil123:
Allows users to set their own winit features. 

Previously the winit default features would have been enabled even if the user disabled it in their own Cargo.toml. This is because of the way Cargo does [feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification).

This modification does not change any functionality because imgui-winit-support does not use winit::platform.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,7 @@ jobs:
             ${{ runner.os }}-target-lint-
       - run: cargo clippy --workspace --all-targets
       # supported winit versions
-      - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --features winit-19 --features winit-20 --features winit-22 --features winit-23/default --features winit-24/default --all-targets
+      - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --all-features --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-19 --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-20 --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-22 --all-targets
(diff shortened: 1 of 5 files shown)


thomcc (approved):
Thanks, looks good assuming CI is green.

thomcc:
I don't think I can get this in a release easily soon, since there are a lot of breaking changes and we have a lot more planned (see #462). If you need it badly I could probably backport it and cut a patch release, but probably not until after the weekend.

dzil123:
Thank you for the patch release offer, but this isn't urgent. I understand v0.8.0 is a big change.

thomcc:
Ah, it looks like it won't be fully backwards compatible anyway...

dzil123:
The default features can be re-enabled only during CI with the `--features winit-2x/default` flag.

I haven't figured out how to do this for the `--all-features` clippy run.

.github/workflows/ci.yml
```diff
@@ -46,7 +46,7 @@ jobs:
             ${{ runner.os }}-target-lint-
       - run: cargo clippy --workspace --all-targets
       # supported winit versions
-      - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --all-features --all-targets
+      - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --features winit-19 --features winit-20 --features winit-22 --features winit-23/default --features winit-24/default --all-targets
```
thomcc, line 49: Ugh, this pretty much sucks. I'd like to find a better way. Note that among other things, you can pass more than one feature at a time, but is there no way to make --all-features work?

Also, please reflect any changes you make here to https://github.com/imgui-rs/imgui-rs/blob/master/xtask/src/main.rs#L38-L61, which is used for local dev (it's not used for CI too so that we get more insight on which operation in particular failed)

thomcc:
This will need some updates after #485, sorry.

dzil123:
How's that? I wasn't sure if I should add winit 0.25 to Github CI and xtask.

sanbox-irl:
hey @dzil123 looks like we lost the thread on this one here.

I think we're all good to go right? Thom's primary issue was losing `-- all-features` and you did a trick with a fake feature to take care of that. I think we just need a rebase and then it can merge?

dzil123:
Thanks for checking in! Yeah, I think this is good to go.

I modified the default features list to include the winit default features as well, to avoid breaking projects that somehow depend on imgui-winit-support but not winit. So to opt out of winit default features in a consumer crate, default-features should be set to false for both imgui-winit-support and winit.

I also added winit 0.25 to Github CI and xtask as I said above.

One thing however is that cargo clippy/test for winit 0.19 is failing on imgui-rs:master, so Github may not allow this to be merged.

sanbox-irl:
Fixed the winit19 error, rebased this commit to it, and added to the changelog. Thanks so much for the work! This is an area that I am not well equipped, and our winit dependency specifically is a mess, so I appreciate you doing the leg work here. Thank you!

Issue #444, opened 2021-02-08
Removing `imgui-gfx-renderer`?


dbr:
Would it make sense to remove the renderer which uses the deprecated `gfx` crate?

> https://github.com/gfx-rs/gfx#the-gfx-crate
>
> This repository was originally home to the [`gfx`](https://crates.io/crates/gfx) crate, which is now deprecated. You can find the latest versions of the code for that crate in the [`pre-ll`](https://github.com/gfx-rs/gfx/tree/pre-ll) branch of this repository.

and the last release of `gfx` (0.18.2) was close to a year ago and is presumably the final release

The reason it might be worth removing is it is maybe (at least it was to me) a little confusing as the `gfx` repo is still active, but it is now being used for the `gfx-hal` package. So I initially assumed the ` imgui-gfx-renderer` was as-valid a thing to use for a new project as `glium`. To add further confusion, the [gllium repo says it is "no longer actively developed by its original author"](https://github.com/glium/glium) (which on first glance reads like the project is dead, but it's actually just maintained by different people - edit: made issue about clarifying this with glium project [here](https://github.com/glium/glium/issues/1912))

I can see the value in having more than one example integration codebase - which is probably where #243 comes in I guess!

sanbox-irl:
i am pretty in favor of this. It's useful to have the rendering backends in *some* form for CI to identify any reversions, but it'd be nicer instead if we could figure out how to check out imgui-wgpu and use that as a test bench too.

sanbox-irl:
@thomcc opinions?

thomcc:
In favor of it, but not really in favor of having `glium` (which is very lightly maintained and not commonly used) be the only example impl here. I think long term we probably want:

- A glow renderer instead of the glium renderer.
- A wgpu renderer instead of the gfx renderer, although for this I suspect the community does a better job than we would (or than I would anyway), so it requires more thought / coordination.

That said, that stuff doesn't need to block removal of the gfx one.

sanbox-irl:
I do not believe there is a currently publicly available glow implementation. My own imgui bindings are using `glow` however, and are closely based on the dear imgui opengl bindings. I doubt the https://github.com/Yatekii/imgui-wgpu-rs would be interested in being consumed by us, but maybe they are? I can reach out

dbr:
@Hperigo has recently started working on https://github.com/Hperigo/imgui-glow-renderer

[...]

PR #482, opened 2021-05-06
WIP: A software renderer

dbr:
Something I wanted for writing test cases for some imgui widgets, but might be useful for a few reasons:

1. An alternative backend as discussed in #444 - not being dependent on any particular hardware or OS makes it quite a nice reference (assuming it works properly)
2. Writing more comprehensive tests for imgui-rs code, since it can run on any CI platform (doesn't need actual graphics hardware, or platform-specific things like Xvfb)
3. I have a half-formed idea that it would be possible to write mdbook preprocessor which would allow you to write docs including snippets like:

      ```rust

      # let ui = setup_renderer().display_size([100.0, 200.0])
      ui.button(im_str!("Example button!")
      # ui.render_png()
      ```

      ..and so building the mdbook would magically and automatically generate all the screenshots (which would be much less tedious than manually cropping screenshots, and also serve as some form of high-level test case)

The renderer seems to work quite well in the limited testing I've given it - but there are a few oddities/issues:

1. The first few frames never render properly - I'm pretty sure this is due to how imgui calculates layouts etc
2. Had to duplicate the `Transform::invert` code from tiny_skia as it's currently private. Need to open an issue with that project to see if it's something that can be made `pub` (if not, the implementation can be rewritten in a much shorter way)
3. The `TestHelper` struct could also be improved drastically, and should probably have a better name. It's intended to make the "just render this widget and give me a PNG" case as simple
4. It would be helpful to have some kind of `imgui-software-support` which lets you easily emulate mouse/keyboard input etc

Thoughts?

### DIFF
--- a/imgui-software-renderer/Cargo.toml
+++ b/imgui-software-renderer/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "imgui-software-renderer"
+version = "0.7.0"
+edition = "2018"
+authors = ["The imgui-rs Developers"]
+description = "Software renderer for imgui-rs"
+homepage = "https://github.com/imgui-rs/imgui-rs"
+repository = "https://github.com/imgui-rs/imgui-rs"
+license = "MIT/Apache-2.0"
+categories = ["gui", "rendering"]
+
+[dependencies]
+tiny-skia = "0.5"
+imgui = { version = "0.7.0", path = "../imgui" }
--- a/imgui-software-renderer/README.md
+++ b/imgui-software-renderer/README.md
@@ -1,0 +1,28 @@
+# Software renderer for imgui-rs
+
+A renderer backend for imgui-rs to allow easy capture of an "dear imgui" interface to a file, without requiring any graphics hardware or complex dependencies.
+
+Some use cases:
+
+1. In a test case, run an application in a "headless" mode, interacting with the interface programatically. If an assertion fails, save the rendered image to a PNG file for debugging purposes.
+2. In a test case, render a widget and compare it again a "known good" reference image to check a complex custom widget hasn't been altered
+3. Use the renderer to automatically generate screenshots of an application or widget for use in documentation/tutorials.
+
+
+## Notes/performance
+
+Performance is not a high priority for this project.
+
+The renderer is reasonably fast (the basic example with the "Dear ImGUI demo" nad a few other windows renders in around 6ms per frame in release mode), but this is almost entirely thanks to the speed of [`tiny_skia`](https://github.com/RazrFalcon/tiny-skia) used for rasterisation.
+
+The renderer is inspired by [this C++ implementation](https://github.com/emilk/imgui_software_renderer/blob/master/src/imgui_sw.cpp) which contains many optimisations (e.g combining the polygons for each text character into a single square).
+
+The primary goals of this renderer are:
+
+1. Small, simple to follow code base
+2. Consistent output - the same draw list should produce the same pixel data
+
+
+## Usage
+
+See the `examples/` directory.
--- a/imgui-software-renderer/examples/software_renderer_basic.rs
+++ b/imgui-software-renderer/examples/software_renderer_basic.rs
@@ -1,0 +1,99 @@
+use tiny_skia::Pixmap;
+use imgui::{im_str, FontConfig, FontSource};
+
+fn main() {
+    // Size of our software "display"
+    let width = 1000;
+    let height = 500;
+
+    // Create imgui Context as per usual
+    let mut imgui_ctx = imgui::Context::create();
+
+    // Don't save window layout etc
+    imgui_ctx.set_ini_filename(None);
+
+    // Tell imgui to draw a cursor, and set the cursor position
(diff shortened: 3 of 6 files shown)


sanbox-irl:
Hey dbr -- I really like this PR and this idea. It would be stupendous if our examples ran on this, and therefore would allow us to do actual GUI tests. Like I'd love to mock up mouse inputs to make sure that we're handling over frame checks and what not.

That being said, I'm not sure about `tiny-skia`. Have you seen https://github.com/zesterer/euc? It's pure rust, which is nice, but it also looks a bit simpler in API to use.

What do you think?

sanbox-irl:
also dbr, if you have a second, would you mind getting in contact with me? No pressure if you don't want to, but would love to chat in a faster way -- my discord is `sanbox_irl#7777` and my email is `jjspira` on gmail

thomcc:
Neat! I have a few concerns though... ideally, we'd either:
1. Avoid exposing `tiny_skia` as part of the public API.
2. Call it something more specific like `imgui-tiny_skia-renderer`.

Personally, I'd greatly prefer the first of these, as I don't think skia-style path-based drawing is actually what we want here, and so we'd want to move away from it eventually (either to some other crate or whatever. For this kind of thing I suspect we could do a rasterizer that supports what imgui needs in under 1kloc even, tbh)

Anyway, I only care that we work this stuff out before publishing it on crates.io. If we just wanted to use it for internal testing, we can figure it out later.

dbr:
> That being said, I'm not sure about tiny-skia. Have you seen https://github.com/zesterer/euc? It's pure rust, which is nice, but it also looks a bit simpler in API to use.

tiny-skia is pure Rust also, it doesn't have much to do with the skia C++ library aside from using it as a reference

I mostly selected it as we were already using resvg in the project I prototyped this in, and it was a fairly small dependency and fast (if I remember right, it ran quicker than the C++ implementation linked in readme, which is a hand written rasterizer with various optimizations like merging text tris into single boxes)

That said, replacing it with something more directly aimed as rendering polygons would likely simplify the code a bit, and either way hiding the implementation from public API seems perfectly reasonable. The C++ implementation was about 700 lines so is certainly plausible to just write the rasterizer from scratch, it is almost entirely just requires basic filled/textured polygons

I wonder if there's something spiritually similar to [mint](https://crates.io/crates/mint) but for pixmaps - as the `PixmapRef` was probably the only real benefit of exposing it (I've written to many "convert this image to `[B, G, R, ...]` to `[R, R, R, R, ... G, G, G G, ... B, B, B, B, ...]` etc methods to want to expose a simple slice of pixel values :sweat_smile: )

> Anyway, I only care that we work this stuff out before publishing it on crates.io. If we just wanted to use it for internal testing, we can figure it out later.

Agreed entirely - my initial plan for this is to try and get a `mdbook` showing generated screenshots (/GIF's) of various IMGUI widgets, as part docs/part integration test. I think I have a good idea of how to go about this now, shall give this a go hopefully some time this week!

sanbox-irl:
Ah sorry I misunderstood the skia! I don't mind at all what implementation we use for our own example purposes. I think eventually releasing this would be excellent, but just getting it for tests would be really nice. Really, really nice.

thomcc:
> tiny-skia is pure Rust also, it doesn't have much to do with the skia C++ library aside from using it as a reference

My concern is more that path rendering (like the skia api) and triangle rendering (like what the draw data expects) are... different enough that it seems likely to be a bit of whackamole for weirdness. To the point where I worry it would not be an effective tool for debugging certain kinds of issues.

You linked https://github.com/emilk/imgui_software_renderer/blob/master/src/imgui_sw.cpp in the readme, and IMO its closer to what we want, although that code is more complex than I think we need.

Anyway I actually have background in software rendering (I've written production ones, and often use triangle rasterization for game programming stuff[^1]) so I'll take a quick stab at this to put my money where my mouth is. (Either way I think the important thing is to not tie down the approach much)

[^1]: Really; it's great for the kind of "oh no this part of my game is grid-based, but now needs to work with a shape which is... *too weird for the grid* 🙀" problems that comes up all the time. (I actually don't know how else people solve them, TBH)

dbr:
Oh forgot to update here - the WIP code is here (barely presentable, but wouldn't take much to tidy it to a "releasable" state)

https://github.com/dbr/imgui-book-prototype

You run `mdbook build` in the docs folder, and it will:
1. Invoke the preprocessor `cargo run --release --bin imgui-book-processor --`
2. The `build.rs` for the preprocessor extracts all fenced codeblocks with start like ` ```imgui-example`, wraps them up in functions which execute them with the software renderer and produce an image
3. The processor is then run, it then does the same codeblock extraction, invokes the appropriate function created above, and inserts the image

The end result is something like this:

    # The Second
    
    Simplest example, display code and render output
    
    ```imgui-example
    use imgui::im_str;
    ui.text(im_str!("Basic example"))
    ```
    
    Hide code, but display resulting image:
    ```imgui-example,hide_code
    use imgui::im_str;
    ui.text(im_str!("Example with no code visible!"))
    ```

..becomes something like this:

image

PR #485, opened 2021-05-22
Add support & upgrade to winit 0.25

repi:
Adds a new `winit-25` feature that works just like the previous `winit-24`.

Leaves default for now on `winit-24` though as `glutin` is still using it and can't mix and use multiple winit versions within this repo

### DIFF
--- a/imgui-gfx-examples/Cargo.toml
+++ b/imgui-gfx-examples/Cargo.toml
@@ -22,7 +22,7 @@ glutin = "0.26"
 image = "0.23"
 imgui = { path = "../imgui" }
 imgui-gfx-renderer = { path = "../imgui-gfx-renderer" }
-imgui-winit-support = { path = "../imgui-winit-support" }
+imgui-winit-support = { path = "../imgui-winit-support", default-features = false, features = ["winit-24"] } # glutin still depends on winit 0.24
 old_school_gfx_glutin_ext = "0.26"
 
 [target.'cfg(windows)'.dev-dependencies]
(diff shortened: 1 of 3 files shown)


thomcc (approved):
Hm, I do wonder if it's worth keeping support for winit-24. I suspect the valuable ones are 19, and latest (and everything in between doesn't matter). Oh well, this is fine for now.

thomcc:
(Ah, rather, it's fine pending green CI, ofc)

repi:
Believe need to at least support the previous version (0.24) in this case to not force complex dependent upgrades. Even some examples in this repo used `glutin` which still uses `winit` 0.24

repi:
Hmm ok that can't really be solved as the CI builds the entire workspace which forces cargo feature resolver to unify all the features and build the `imgui-winit-support` with multiple winit versions which it will fail on (one for the default and one for winit 0.24 for the tests).

Could change so just add support for 0.25 but keep the default here on 0.24 until glutin and the examples also can use 0.25, believe that is the only workable approach?

thomcc:
> Could change so just add support for 0.25 but keep the default here on 0.24 until glutin and the examples also can use 0.25, believe that is the only workable approach?

SGTM

thomcc:
Please file an issue for updating to winit-25 once glutin does.

Or I can remember to do so later today.

thomcc:
Hm? I would have been fine merging 92bed93.

thomcc:
Thanks. Sorry about the hassles here. I'll merge this in a bit and take care of filing the followup.

thomcc:
Since this doesn't change the default feature I think I can cut this as a patch release (0.7.1) of  imgui-winit-support too, which means I can probably get to it later today.

thomcc:
Actually, I just did it now https://crates.io/crates/imgui-winit-support/0.7.1

PR #491, opened 2021-06-02
Add example to Slider.range() doc

dbr:
Was a bit hard to work out what to pass to the `.range(...)` method, so added an example to clarify


sanbox-irl:
Thank you!

PR #494, opened 2021-06-08
Include LICENSE and README files where relevant

### LINKED ISSUE #493: Fedora packaging need license and readme file
Hi.
Fedora need to have access to the license and the readme file in the crate source. \
A ``include = [ "README.markdown", "LICENSE-APACHE", "LICENSE-MIT"] should be sufficent for fedora.

thomcc:
Fixes #493 

For the license files, I've just duplicated them, but for the readme, I symlinked. Symlinks don't really work the same way on windows, so I only opted to follow that for the file that actually gets changed.

That said, note that `imgui-winit-support`, `imgui-gfx-renderer`, and `imgui-glium-renderer` don't have a readme (giving them one is an open TODO, and making it the `imgui-rs` readme wouldn't really be accurate).

Anyway, I've confirmed just having these makes these show up in the package file (using `cargo package` locally) that there's no need to include these in the `include = []` list, which is good, because it would be quite painful to maintain manually.

For example this was the include I needed to match the default behavior for `imgui-sys`, and even this took a lot of trial and error. Getting the wrong one can easily result in a package that cannot build, as well.

```toml
include = [
    "../LICENSE-MIT",
    "../LICENSE-APACHE",
    "README.markdown",
    "Cargo.toml",
    "include_all_imgui.cpp",
    "build.rs",
    "/src",
    "third-party/",
    # exclude json, lua, and the imgui subdirs (imgui/examples, imgui/docs, etc)
    "!third-party/*.json",
    "!third-party/*.lua",
    "!third-party/imgui/*/",
    # Exclude imgui dotfiles (seems to follow behavior
	# of the default `include`...)
    "!third-party/imgui/.*",
]
```

### DIFF
--- a/imgui-gfx-renderer/LICENSE-APACHE
+++ b/imgui-gfx-renderer/LICENSE-APACHE
@@ -1,0 +1,202 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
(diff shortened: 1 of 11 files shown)


thomcc:
IDK when this will be in a release.

There are currently unreleased breaking changes on master, and so getting these in a release before 0.8 is ready (which is in progress, and I should be working on soon, but has a number of things we'd like to finish before it can go out) would require the creation of a hotfix branch to cut a new 0.7.x.

(That said, I guess you'd actually need it as 0.7 anyway, since if it were only in 0.8 then you couldn't package anything currently using 0.7 and not 0.8... So there's probably no way around that if you need this)

Anyway, this is doable just tedious a pain. @remilauzier is this blocking you packaging something else, or is it just the library itself?

remilauzier:
No it doesn't block me. Fedora just don't like to have a long list of thing to downloads.

thomcc:
Alright, sounds good then. I won't cut a patch release for this.

P.S. Unrelated question: Which Rust version are Fedora packages allowed to use? Is it like Debian in that it's using something kind of old that Rust devs should be aware of?

remilauzier:
Nah. Fedora follow rust stable release. We are at 1.52.1 for now.

PR #496, opened 2021-06-10
Add "freetype" feature

dbr:
Quick attempt at #359 - adds feature to enable the [freetype font rasterizer](https://github.com/ocornut/imgui/tree/d0c6dd9bafcfb0845153080fe00d74e62ad28c86/misc/freetype)

Not tested much beyond "seems to work" comparing before/after with:
```
cargo run --example multiple_fonts --features freetype
```
..which requires `libfreetype6-dev` on Debian. No idea how it'll get on under other platforms, particularly Windows

There is one compiler warning:

```
warning: ./third-party/imgui/misc/freetype/imgui_freetype.cpp:357:8: warning: ‘ImFontBuildSrcGlyphFT’ has a field ‘ImFontBuildSrcGlyphFT::Info’ whose type uses the anonymous namespace [-Wsubobject-linkage]
warning:  struct ImFontBuildSrcGlyphFT
warning:         ^~~~~~~~~~~~~~~~~~~~~
warning: ./third-party/imgui/misc/freetype/imgui_freetype.cpp:366:8: warning: ‘ImFontBuildSrcDataFT’ has a field ‘ImFontBuildSrcDataFT::Font’ whose type uses the anonymous namespace [-Wsubobject-linkage]
warning:  struct ImFontBuildSrcDataFT
warning:         ^~~~~~~~~~~~~~~~~~~~
```
..which I think is an upstream thing

PR #504, opened 2021-07-07
Fix examples DPI scaling when scaling is non-integer

drojf:
I recently ran into an issue where DPI scaling on Windows wouldn't work properly if scaling by a non-integer value (125%, 150%, 175% etc). My program was based on the example's `support/mod.rs` file.

When you start the program (not changing DPI while program is already running), the GUI would be offset, and the hitboxes for the buttons wouldn't line up with the buttons themselves. I re-tested this on the imgui-rs hello-world example, and the same thing happened.

Changing the HiDpiMode to `HiDpiMode::Default` in `mod.rs` seems to fix the issue - non-integer scaling values work correctly when the program is started, and the UI is still usable if you change DPI while the program is running.

I'm not sure if this is the correct way to fix this problem though, so I'm leaving this as a draft PR.

I also changed the value in the `imgui-gfx-examples` folder, but I haven't tested that as I haven't got it to run on my computer.

### DIFF
--- a/imgui-examples/examples/support/mod.rs
+++ b/imgui-examples/examples/support/mod.rs
@@ -46,7 +46,7 @@ pub fn init(title: &str) -> System {
     {
         let gl_window = display.gl_window();
         let window = gl_window.window();
-        platform.attach_window(imgui.io_mut(), window, HiDpiMode::Rounded);
+        platform.attach_window(imgui.io_mut(), window, HiDpiMode::Default);
     }
 
     let hidpi_factor = platform.hidpi_factor();
--- a/imgui-gfx-examples/examples/support/mod.rs
+++ b/imgui-gfx-examples/examples/support/mod.rs
@@ -64,7 +64,7 @@ pub fn init(title: &str) -> System {
     imgui.io_mut().font_global_scale = (1.0 / hidpi_factor) as f32;
 
     let render_sys = RenderSystem::init(&mut imgui, builder, &events_loop);
-    platform.attach_window(imgui.io_mut(), render_sys.window(), HiDpiMode::Rounded);
+    platform.attach_window(imgui.io_mut(), render_sys.window(), HiDpiMode::Default);
     System {
         events_loop,
         imgui,


sanbox-irl:
absolutely neither am i. if anyone knows or has a good understanding of this issue, it would be very appreciated

drojf:
I had a look at the git blame to figure out where the "Rounded" came from, and it appears to be from this commit 3 years ago: "Restrict DPI factor to integers in support code" https://github.com/imgui-rs/imgui-rs/commit/c22035250a3824e5297b450b93a836788961c2e5

> Restrict DPI factor to integers in support code 
> 
> The library still supports fractional DPI factors, but it seems like a
good idea to use integer factors in examples to avoid blurry fonts and
other issues which might give the wrong impression.

https://github.com/imgui-rs/imgui-rs/blob/c22035250a3824e5297b450b93a836788961c2e5/imgui-examples/examples/support/mod.rs#L27-L29

Oddly enough, using the "Default" (non-rounded) option looks good on Windows at all scaling values (it is not blurry). The only time it is blurry is if you change DPI while the program is running, but after restarting the program it looks good again.

dbr:
What I've found worked well is:

1. `HiDpiMode::Default` as in this PR
2. Round the font size to nearest integer (very important) - e.g `let font_size = ((13.0 * hidpi_factor) as f32).round();`
3. Increase the oversampling, something like `FontConfig { oversample_h: 4, oversample_v: 4, ...}` (no idea if those values are optimal - I think I arbitrary shoved in 4 and got less aliasing weirdness when subtly moving text around in the UI, haven't investigated the range of values)

I've checked this on Linux (with quite a variety of DPI screens), two Windows machines (on a 23" 4K monitor and a 21" 1080p one), and one macOS display (Macbook Pro) and it seemed to work reasonably well in all cases.

The only other major things which helps a little are the freetype rasterizer (#496) which mainly helps with smaller font sizes (e.g on 1.0 scaling factor screens); and I've not yet investigated what impact using a linear frame-buffer, so the fonts are blended in a "gamma-correct" manner, but in theory should help font rendering also.

dbr:
> 2. Round the font size to nearest integer (very important) - e.g let font_size = ((13.0 * hidpi_factor) as f32).round();

I since realized, at least when using `winit`, scaling by the `hidpi_factor` is wrong and results in things being "doubly scaled".

If you just set `let font_size = 13.0;` then draw everything without considering the scaling factor, it all "just works" - as in if you force the scaling to 1.0, it looks exactly half the size of same window when forced to HiDPI factor of 2.0 (i.e assuming that factor is determined properly, your UI will look same on different pixel-density screens)

I think by default everything in `imgui_winit_support` is using "logical pixels", which as described here https://docs.rs/winit/0.25.0/winit/dpi/index.html accounts for the scaling :partying_face:

dbr:
I've made a PR with a bunch of changes relating to this, which I think should make the standalone examples work about as well across different DPI screens as I can get - #570

drojf:
Ah nice - I don't think there's any need for this PR anymore (besides as a record of previous discussion), so I'm gonna close it.

colinmarc:
Hi, sorry to reopen this thread, but this isn't working as intended on my mac (using winit).

I can get added fonts to display correctly by doing the following:

```
let font_size = 13.0 * window.scale_factor();
imgui.io_mut().font_global_scale = (1.0 / window.scale_factor()) as f32;
```

However, it seems like this should be handled by the winit platform.

PR #507, opened 2021-07-27
Support for button

scottleesft:


sanbox-irl:
This may just be a doc comment issue -- we have support for Buttons. It's just a method on `ui` -- `ui.button` and `ui.button_with_size`. 

In general, builders with `.build` are something we're actively trying to move away from in this Repo, but I appreciate the work you put in! I think unfortunately it's not necessary

scottleesft:
@sanbox-irl Ah, thanks for the insight there - no worries at all. That's a goof on my part as I am learning the library.

Issue #402, opened 2020-12-06
Determine policy around MSRV and new MSRV


thomcc:
I suspect I'll be bumping the MSRV to at least 1.46 to take advantage of const fn, but more broadly, I'd like a policy here.

More broadly, I'd like to find a more tenable strategy for MSRV than either:

- waiting for our deps to force us to bump it.
- forcing latest stable.

My inclination here is to adopt a policy of `stable - N` (e.g. `stable - 1` would mean we support stable and the previous release). I'm thinking `N` is between 1 and 3. (It seems like a good balance that's predictable, avoids the worst problems of "latest stable", while not forcing you to make the semver number useless).

The other option I'm fond of is to consider it a minor (but not major) bump, which is not really relevant here, but avoids the version becoming almost entirely meaningless.

sanbox-irl:
I think the minor bump is the right call.

Most large projects will be using a fork, or a completely specified version, of this project anyway, so "surprising" someone with an update, when they aren't updated rustc, would surprise me.

I dislike keeping up with stable Rust since this forces significant amounts of updates just to keep to MSRV policy.

thomcc:
The issue with it being a minor bump is that prior to 1.0 (which I'm not ready for) is that it's equivalent to no policy at all.

sanbox-irl:
the underlying idea here is that you intend to move the repo to be MSRV on the new update of Rust which is expected to have min_const_generics, right?

thomcc:
I meant the const fn changes of 1.46, but it's possible there's some array code that would benefit from min_const_generics.

dbr:
For tools where they are likely to be packaged by distros, using a minimum version of "whatever is packaged in most distros" is a good approach to MSRV policy, however that probably isn't a major consideration for imgui-rs.

Otherwise I'd be happy with incrementing it "as needed" - e.g bumping the MSRV for a minor helper method which reduces 2 lines to 1 is probably not worthwhile, but for something like the const-generics, sure.

I don't have strong opinions on if bumping the MSRV being a major bump versus minor etc. The thing I *do* have strong opinions on is: bump the version intentionally, and note the change in the changelog

So on these lines, I've made a PR to test the current minimum version in CI:

https://github.com/imgui-rs/imgui-rs/pull/512

That way if a PR adds usage of something added after 1.48 it'll be easy to spot - and either the version can be bumped or the PR be altered

sanbox-irl:
[...]

PR #512, opened 2021-08-22
Test against Rust 1.48

dbr:
Currently Rust 1.48 is the minimum required by the project (excluding one usage of `bool::then` in one of the `xtask submodule` command), mainly due to the `#[doc(alias = ..)]` which was stabilized then

So I've added this version to be tested in CI, so use of any newer things can be noticed and, at the very least, noted in changelog for a new release (pending a policy being decided in #402)

Additionally clippy will be run from this version to avoid the linter-whackamole problem, when a new stable adds new lints and suddenly fails CI on previously-accepted code

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,20 +11,21 @@ jobs:
   clippy:
     name: Run linter
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


sanbox-irl:
lol "gentle nudge towards and msrv" I respect it. thanks for the work.

PR #514, opened 2021-09-05
init draft.

sanbox-irl:
needs doc work before it's ready to go. @dbr lmk thoughts


dbr:
Looks good to me - simpler to use, implementation is simpler, and I don't have to remember the `=` in `0.0..=1.0` (or the `..`) :partying_face:

PR #516, opened 2021-09-07
Input Widget Callback Draft

sanbox-irl:
An initial attempt at input widget callbacks, based on a trait object with dynamic dispatch.

It also does some fairly large wrappign and uses some very questionable UB-y code here,
so I'm still submitting this as a draft until I go over it in more detail.

### DIFF
--- a/imgui-examples/examples/keyboard.rs
+++ b/imgui-examples/examples/keyboard.rs
@@ -11,7 +11,8 @@ fn main() {
     let mut uncaptured_counter = 0u32;
     let mut home_counter = 0u32;
     let mut f1_release_count = 0u32;
-    let mut text_buffer = ImString::new("");
+    let mut text_buffer = ImString::new("with some buffer");
+    text_buffer.reserve(100);
 
     system.main_loop(move |_, ui| {
         Window::new(im_str!("Means of accessing key state"))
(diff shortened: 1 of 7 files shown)


sanbox-irl:
Okay, I changed this to make it non-ub (at least I think), and simpler to use. I also made examples

Here's the basic calling convention:

```rs
                struct AllCallback;
                impl InputTextCallbackHandler for AllCallback {
                    fn char_filter(&mut self, c: char) -> Option<char> {
                        println!("Char filter fired! This means a char was inputted.");
                        Some(c)
                    }
                    fn on_completion(&mut self, _: TextCallbackData<'_>) {
                        println!("Completion request fired! This means the tab key was hit.");
                    }

                    fn on_edit(&mut self, _: TextCallbackData<'_>) {
                        println!("Edit was fired! Any edit will cause this to fire.")
                    }

                    fn on_history(&mut self, dir: HistoryDirection, _: TextCallbackData<'_>) {
                        println!("History was fired by pressing {:?}", dir);
                    }

                    fn on_always(&mut self, _: TextCallbackData<'_>) {
                        // We don't actually print this out because it will flood your log a lot!
                        // println!("The always callback fired! It always fires.");
                    }
                }

                ui.input_text(
                    im_str!("All Callbacks logging"),
                    buffers.get_mut(0).unwrap(),
                )
                .callback(InputTextCallback::all(), AllCallback)
                .build();
                ```

sanbox-irl:
I am going to merge this in once it goes green after lunch in case anyone has any objections.

Obviously, this method precludes closures being used, which are very attractive in Rust. However, it wildly simplifies the types within InputText which seemed important to me, since frankly, this library has lacked support for callbacks ENTIRELY for years and we haven't gotten a single issue about them. In fact, dear imgui itself appears to be slightly borked as per its docs on callbacks (it gives a nullptr when i don't think it should), yet no one seems to mind too much. This is a niche feature and as such, I think it's okay to have a non-elegant API for the sake of keeping the non-obscure case simple.

PR #519, opened 2021-09-13
Add disable methods (and update to imgui 1.84.2)

dbr:
Most of the interesting change is in the last commit b24b726a434d99af0dc4659f498a8e572b5f114d - rest is fairly mechanical updating to latest imgui upstream.

Although it's worth noting I did change how `cargo xtask bindgen` works, so it doesn't call `cargo xtask fixmod` - which @thomcc may have some opinion on? (although we should either keep this change, or change the `docs/upgrading-imgui.md` doc as the two contradict each other)

For the interesting part, there is token and closure based methods to disable widgets.

I think the only bit of some contention might be: I have an ~irrational~ rational dislike of double-/triple-negative boolean logic in code (e.g `disabled(!false)` then introduce a `let not_enabled = ...` :cry: ), thus I've included both `enabled` and `disabled` methods (similar to Qt's `setEnabled(...)` and `setDisabled(...)`). Arguably a little redundant, but I think it's worthwhile to include to improve easy-of-reading


sanbox-irl:
Letting thom know to check if they have time. I do hilariously have the same feeling actually on negative bools -- I often write `== false` instead of `!false` in if checks, which is a tendency I've had to avoid in this repo since I'm told it's a little idiosyncratic

dbr:
Updated sans `im_str` :partying_face:

thomcc (approved):
lgtm

sanbox-irl:
thank you dbr!

dbr:
Oh just realized the CHANGELOG needs updated to mention 1.84.2 (currently it only mentions the previous update to 1.82)

sanbox-irl:
@dbr good memory. will add this to the tracking issue so I don't forget, but I'll be doing over the entire changelog/readme fairly extensively before release.

PR #522, opened 2021-09-15
Fix text input deletion

Jake-Shadle:
The previous code didn't take text deletion into account, so removing characters from a text input would result in the string being filled with null terminators.


sanbox-irl:
thanks -- good catch

Issue #525, opened 2021-09-17
Access to Style from Ui


lwiklendt:
When creating custom widgets we would like to follow the current styling, such as item spacing, rounding, etc. `Style` is about 1KB so it feels wasteful to get current style vars via `Ui::clone_style(&self) -> Style` inside a custem widget function for every widget instance of which there could be hundreds. Due to the homogeneity of color data we already have the neat function `Ui::style_color(&self, style_color: StyleColor) -> [f32; 4]`, but it would be good to have something for the non-color style variables too.

One solution could be to add one function per non-color `Style` field, e.g. `Ui::style_item_spacing(&self) -> [f32; 2]`,  `Ui::style_child_rounding(&self) -> f32`, `Ui::style_color_button_position(&self) -> Direction`, etc.

PR #526, opened 2021-09-17
Add Style accessor functions to Ui

lwiklendt:
Proposed solution to issue #525

### DIFF
--- a/imgui/src/utils.rs
+++ b/imgui/src/utils.rs
@@ -3,8 +3,8 @@ use bitflags::bitflags;
 
 use crate::input::mouse::MouseButton;
 use crate::style::StyleColor;
-use crate::sys;
 use crate::Ui;
+use crate::{sys, Direction};
 
 bitflags! {
     /// Item hover check option flags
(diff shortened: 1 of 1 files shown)


sanbox-irl:
Don't mind this @lwiklendt, but is there a reason we shouldn't just expose an `&Style` in the UI? This does add some maintenance burden onto the repo

lwiklendt:
I don't understand myself why originally `&Style` wasn't exposed. I think maybe @Gekkio [wrote](https://github.com/imgui-rs/imgui-rs/blame/5f441b7b1d40a3d2fe727cbc90e7e9464cee5636/imgui/src/lib.rs#L186) the clone version `Ui::clone_style(&self) -> Style` but I don't know how to get to the "0.1-dev" branch to see if there's a reason given. I agree if there are no problems, a `Ui::style(&self) -> &Style` would certainly be preferable.

sanbox-irl:
Okay, I have some answers for you on why it's set up the way it is -- effectively, we violate memory safety if we give an `&Style`.

I created this simple program:


```rs
        let style = ui.style();
        let _t = ui.push_style_var(imgui::StyleVar::Alpha(0.2));

        let new_style = ui.style();

        println!(
            "equals, {}\n{:?}\n{:?}",
            style == new_style,
            style,
            new_style
        );

    // in ui...
    pub fn style(&self) -> &crate::Style {
        self.ctx.style()
    }
```

My expectation was that this would yield `true` and `alpha` in the style struct would be set to `1.0`, the global default. Unfortunately, it yielded `true`, but with `alpha` at `0.2`. This means that we held the immutable pointer to the C++ style struct in `style` while ImGui was mutating it when we called `push_style_var`. Obviously, we can't do, since the compiler will mess with this tremendously with optimizations it believes to be sound.

I think we have three options:
1. accept this PR with the negative side effects referenced
2. change the signature of `style` to `&mut self`, and yet return an `&crate::Style` (imgui's docs are explicit that we *should not* write to this data after calling `NewFrame` -- I assume `push_style_var` also does other work in addition to writing to this struct). This is kind of funky, but actually I quite like it -- fundamentally, that's the whole idea of Rust really. I'd like to more to eventually have the ui methods use `&mut self` at some point.
3. We can use a getter function that takes `StyleVar` and match on that -- it already corresponds (I believe) to each field in the style struct, so that gives us a compile time warning as we update imgui.
4. give access to a raw ptr with some scare quotes in a safety doc. This can be accepted in addition to others.

What do you think?

thomcc:
I like 2, but I suspect it could require massive changes to peoples codebase to use. 3 and 4 might be the best option. Unfortunately, IIRC the style can be a number of different types, which... complicates things.

lwiklendt:
In the spirit of 3, there could be a `Ui::get_style_var(&self, style_var: &mut StyleVar)` where you supply (with dummy data) the variant you want to retrieve which gets filled with the correct data. It's probably not very "rusty" but it is kind of "imgui-y" considering functions like `Ui::checkbox(&self, label: impl AsRef<str>, value: &mut bool) -> bool` that take a mutable reference, albeit not with dummy data.

thomcc:
A better option might be an `unsafe fn` that returns a `&Style` that you arent allowed to hold across other calls into imgui, since it may be mutated. That's easier to document, at least.

imgui/src/utils.rs
```diff
@@ -173,4 +173,15 @@ impl<'ui> Ui<'ui> {
     pub fn style_color(&self, style_color: StyleColor) -> [f32; 4] {
         self.ctx.style()[style_color]
     }
+    /// Returns a shared reference to the current [`Style`]. This function is tagged as `unsafe`
```
thomcc, line 176: We should note that users who would like to avoid unsafe can use `clone_style()`.

imgui/src/utils.rs
```diff
@@ -173,4 +173,15 @@ impl<'ui> Ui<'ui> {
...
+    /// because pushing via [`push_style_color`](crate::Ui::push_style_color) or
+    /// [`push_style_var`](crate::Ui::push_style_var) or popping via
+    /// [`ColorStackToken::pop`](crate::ColorStackToken::pop) or
+    /// [`StyleStackToken::pop`](crate::StyleStackToken::pop) will modify the values in the returned
+    /// shared reference. Therefore, you should not retain this reference across calls to push and
+    /// pop. The [`clone_style`](Ui::clone_style) version may instead be used to avoid `unsafe`.
```
sanbox-irl, line 182: apologies for the specifics, but can you format this as:

```
Returns a shared reference to the current [`Style`].

## Safety

This function is tagged as `unsafe`...etc

sanbox-irl:
left one comment, once that's resolved, good to merge. Thank you and apologies that you ran into hotel order issue of mutability of the `Ui` object. I have some ideas but wow they;'d be a lot of breaking changes

imgui/src/utils.rs
lwiklendt: No worries, it's good to keep a clean uniform style.

sanbox-irl:
Thank you!

PR #527, opened 2021-09-17
Remove noisy println statements

vojd:
Removes what appears to be development/debug `println!` leftovers. 
Caused quite the confusion when working on clipboard integration in our project.


sanbox-irl:
Thank you for the effort! Sorry about that, hope it wasn't too much of a problem for ya

PR #528, opened 2021-09-17
Remove `println`s

davidpdrsn:
These were showing up while running our application.


sanbox-irl:
Thanks!

sanbox-irl:
Actually going to close in favor of #527 -- please let me know if removing the 'eprintln's' is a concern -- since we abort there, it's the only way we let users know what happened.

PR #532, opened 2021-09-18
Fix glow renderer doc links

toyboot4e:
For the win :)

### DIFF
--- a/imgui-glow-renderer/src/lib.rs
+++ b/imgui-glow-renderer/src/lib.rs
@@ -5,7 +5,9 @@
 //!
 //! # Basic usage
 //!
-//! A few code examples are provided in the source.
+//! A few code [examples] are provided in the source.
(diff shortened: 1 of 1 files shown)


toyboot4e:
`glow` also needs to be updated to 0.11.0

thomcc (approved):
lgtm

sanbox-irl:
@toyboot4e re: glow 0.11.0, we can't trivially upgrade. Unfortunately, we have some issues with their newtype pattern wrappers that we'll need to handle. I'll try to write it out tomorrow, but I've been talking to the maintainers of `glow` and they're working on it. The switches to newtypes in 0.11 made the getters/setters more or less impossible to use, so saving ogl state isn't really possible in glow 0.11. We need that in the glow-renderer, so they're working on ways to restore that in 0.12

PR #536, opened 2021-09-23
Switch [f32; _] types in the public api to `mint` equivalents

sanbox-irl:
This is a larger effort that I'm only just starting. I'm not sure this is the right decision (using language primitives like arrays has a lovely simplicity), but *especially* for handling sizing in complex layouts, I'm always using a real math lib (basically always `glam`), and all the `.into` calls are irksome.

The plan here is that we take `mint` types as *inputs* but still return `[f32; _]` as *outputs* (for example, in getters).

If anyone has opinions on this, I'd love to hear them

### DIFF
--- a/imgui-examples/examples/test_window_impl.rs
+++ b/imgui-examples/examples/test_window_impl.rs
@@ -545,7 +545,7 @@ fn show_test_window(ui: &Ui, state: &mut State, opened: &mut bool) {
             ui.input_float3("input float3", &mut state.vec3f)
                 .build();
             ColorEdit3::new("color 1", &mut state.col1).build(ui);
-            ColorEdit3::new("color 2", &mut state.col2).build(ui);
+            ColorEdit4::new("color 2", &mut state.col2).build(ui);
 
             TreeNode::new("Multi-component Widgets").build(ui, || {
                 ui.input_float2("input float2", &mut state.vec2f)
@@ -601,19 +601,19 @@ fn show_test_window(ui: &Ui, state: &mut State, opened: &mut bool) {
                     "Click on the colored square to open a color picker.
 CTRL+click on individual component to input value.\n",
                 );
-                ColorEdit3::new("MyColor##1", &mut s.color)
+                ColorEdit4::new("MyColor##1", &mut s.color)
                     .flags(misc_flags)
                     .alpha(false)
                     .build(ui);
 
                 ui.text("Color widget HSV with Alpha:");
-                ColorEdit3::new("MyColor##2", &mut s.color)
+                ColorEdit4::new("MyColor##2", &mut s.color)
                     .flags(misc_flags)
                     .input_mode(ColorEditInputMode::HSV)
                     .build(ui);
 
                 ui.text("Color widget with Float Display:");
-                ColorEdit3::new("MyColor##2f", &mut s.color)
+                ColorEdit4::new("MyColor##2f", &mut s.color)
                     .flags(misc_flags)
                     .format(ColorFormat::Float)
                     .build(ui);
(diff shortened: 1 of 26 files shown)


toyboot4e:
impl Into<[f32; 2]> also makes sense.

sanbox-irl:
It does, but for more complex types, mint support is nice. I'm not sure if that'll come up but it's a good crate to support I think

sanbox-irl:
Hello all -- this has now covered the API. Please let me know if you have any thoughts or opinions -- some of these bounds are a bit rough (especially the `InputFloatX` and `ColorPickerX` and `ColorEditX` widgets). Until `mint` upgrades, we can't simplify. I added an issue to them here https://github.com/kvark/mint/issues/67

since the above would be a patch we'd be safe to go to 0.5.7.

I'm failing some of the tests, but will fix over the weekend. that mostly is because this includes a breaking change where the `ColorPicker` and `ColorEdit` structs where de-genericed to just `ColorPicker3` and `ColorPicker4`, since their generic `Into` nature was just...a mess.

dbr:
This looks like a good improvement!

With the caveat that I have almost no idea what I'm talking about: could this potentially have a noticeable impact on binary size due to monomorphization? I have even less idea of how it works in regards to `#[inline]` methods, but basically could something like

```
#[inline]
pub fn size(mut self, size: impl Into<crate::math::MintVec2>) -> Self {
    self.size = size.into().into();
    self
}
```
...mean that user-code like
```
fn some_giant_function(ui: &imgui::Ui) {
    ui.whatever().size([0.0, 1.0]);
    ui.whatever().size(some_vec);
    ui.whatever().size(some_other_vec);
    // and lots more calls
}
```
..would end up with lots of copies of `some_giant_function` with the various permutations of `.into().into()`?

I had a quick search around and only related-ish thing I could find was https://github.com/rust-lang/rust/issues/77767 so I suspect it probably isn't a big problem

sanbox-irl:
not like that -- it would work the other way around though.

So there will only be one `some_giant_function` (you can tell since it has no generic type parameters), but there WILL be have at least three (assuming `some_other_vec` is another `T: Into<MintVec2>`) implementations of `.size` on it. 

Morever, if you noticed, I actually put the trait bound on the impl block itself, so you can't `.size` with different concrete types. Actually, supporting that wouldn't be that difficult, but that sounds NEEDLESS.
So that's the say, this wouldn't work...

```rs
ui.window("hello")
   .size([10.0, 20.0])
   .size(glam::Vec2::splat(100.0)) // error, expecting [f32; 2]
```

However, if a user is using two different math libraries, or freely using arrays + a math library, their binary size will definitely increase.

That doesn't *overly* concern me atm, but what are you thinking? I will say that the compile time increase that this would create somewhat worries me, but since we had `#[inline]` over most of these functions anyway (and generic code is always a candidate for inlining, even across crate boundaries), it seems like the impact would be too serious.

At least on my end, I've been fairly extensively using the draw API for a node graph editor at work, and being able to use glam instead of arrays will wildly simplify that work

sanbox-irl:
I will say, for a normal math library like `glam` or `vek`, we end up at a pretty nice performance situation. We *do* pay to switch to an from the [f32; 2]` types, though we use some fast instructions on this x86_64 backends: https://godbolt.org/z/sdx4exvMf

dbr:
Ahh yeh that makes much more sense (for some reason I half-thought inlining a generic function would make the calling function generic, but on actually thinking about it, that.. wouldn't make much sense!)

As long as there isn't much overhead if people exclusively use `[f32; 2]` then it's hard to see much issue with any potential overhead if someone decides to use 23 different vector types :grin:

> At least on my end, I've been fairly extensively using the draw API for a node graph editor at work, and being able to use glam instead of arrays will wildly simplify that work

I've been using [euclid](https://crates.io/crates/euclid/) for similar purposes and although adding `.to_array()` each time isn't too bad, removing the need for that would be another minor `im_str!`-like papercut gone :partying_face:

Issue #524, opened 2021-09-15
tables api causes segfault


sanbox-irl:
see: https://github.com/ocornut/imgui/issues/3740#issuecomment-920218302

rokups:
Could you please list those functions?

sanbox-irl:
@rokups just two functions actually as I go through this will deref a null possibly: `TableNextRow` and `TableSetBgColor`.

These functions return dummy defaults (mostly false, sometimes a NULL for a char*):
```
TableNextColumn
TableSetColumnIndex
TableGetColumnCount
TableGetColumnIndex
TableGetRowIndex
TableGetColumnName
TableGetColumnFlags
```

I think most pressingly, `TableNextColumn` and `TableSetColumnIndex` should both assert, and it's upstream's choice to assert on the Getters (personally, I like assertions).

Additionally, this repo would definitely appreciate a `IsTableActive` function -- C++ assertions don't play nicely with Rust, and it'd be nice to duplicate some assertions on the rust side in o0 contexts for nicer rust crashes, but not essential.

sanbox-irl:
I've decided to feature gate the tables api for our next release -- although this issue is part of it (segfaulting in Rust is a bit embarrassing tbh), really since the tables api is marked as "beta" I feel like I should signal that to users. I will write more here shortly: https://github.com/imgui-rs/imgui-rs/issues/462

sanbox-irl:
Not only is that a great idea, I've been experimenting with converting all of imgui-rs to this. For example, why can you access the window functions when you might not have a window? This is a huge change though that I'm working

wcassels:
> Not only is that a great idea, I've been experimenting with converting all of imgui-rs to this. For example, why can you access the window functions when you might not have a window? This is a huge change though that I'm working

Cool! If there’s anything in particular that needs working on, I’d be keen to get involved with this effort.

sanbox-irl:
check if out here https://github.com/sanbox-irl/imgui-rs/tree/mutability . These are just some sketches, but it ends up looking similar to `Iterator` I think. I'm also taking some design cues from Oddio https://github.com/Ralith/oddio

dbr:
> I read the 0.8 tracking issue and this suggestion does go against the "Make everything build off of Ui" long-term goal

[...]

PR #537, opened 2021-09-25
Window::new -> Ui::window attempt 2

dbr:
Like #521 without me deleting useful things

..and so I tangentially added a bunch of examples of using `push_id` too

### DIFF
--- a/imgui-examples/examples/creating_windows.rs
+++ b/imgui-examples/examples/creating_windows.rs
@@ -1,0 +1,69 @@
+mod support;
+
+fn main() {
+    let system = support::init(file!());
+
+    system.main_loop(move |_, ui| {
+        // If we don't explicitly create a window before creating some kind of widget, then Dear Imgui will automatically create one
+        ui.text("This text will appear in a default window titled 'Debug'");
+
+        // However, in almost all cases it's best to make a window, so it has a useful title etc
+
+        // imgui-rs has two main methods of creating windows (and these same approaches
+        // apply to many other widgets). First, callback based:
+
+        ui.window("My window via callback")
+        .build(||{
+            ui.text("This content appears in a window");
+
+            // Everything in this callback appears in the window, like this button:
+            ui.button("This button");
+        });
+
+        // Often the callback approach is most convenient, however occasionally the callbacks can be hard to use.
+        // In this case, there is the "token based" approach. You call a method and get a "window token",
+        // everything that happens until the token is dropped is included in the window this is more-or-less how
+        // the Dear ImGui C++ API works)
+        
+        // Here we (maybe) get a window token:
+        let window_token = ui.window("Token based window").begin();
+        if let Some(_t) = window_token {
+            // If the token is Some(...) then the window contents are visible, so we need to draw them!
+            ui.text("Window contents!")
+        }
+
+        // Here we create a window with a specific size, and force it to always have a vertical scrollbar visible
+        ui.window("Big complex window")
+        .size([200.0, 100.0], imgui::Condition::FirstUseEver)
+        .always_vertical_scrollbar(true)
+        .build(||{
+            ui.text("Imagine something complicated here..");
+
+            // Note you can create windows inside other windows, however, they both appear as separate windows.
+            // For example, somewhere deep inside a complex window, we can quickly create a widget to display a
+            // variable, like a graphical "debug print"
+            ui.window("Confusion").build(||{
+                ui.text(format!("Some variable: {:?}", ui.io().mouse_pos))
+            })
+        });
+
+        // If you want to nest windows inside other windows, you can a "child window".
+        // This is essentially a scrollable area, with all the same properties as a regular window
+        ui.window("Parent window").build(||{
+            ui.child_window("Child window")
+            .size([100.0, 100.0])
+            .build(||{
+                for _ in 0..10 {
+                    ui.text("Lines and");
+                }
+            });
+            ui.child_window("Second child window")
+            .size([100.0, 100.0])
+            .build(||{
+                for _ in 0..10 {
+                    ui.text("More and");
+                }
+            });
+        });
+    });
+}
--- a/imgui-examples/examples/creating_windows.rs
+++ b/imgui-examples/examples/creating_windows.rs
@@ -12,8 +12,7 @@ fn main() {
         // imgui-rs has two main methods of creating windows (and these same approaches
         // apply to many other widgets). First, callback based:
 
-        ui.window("My window via callback")
-        .build(||{
+        ui.window("My window via callback").build(|| {
             ui.text("This content appears in a window");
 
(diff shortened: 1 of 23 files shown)


imgui-examples/examples/creating_windows.rs
```diff
@@ -0,0 +1,67 @@
+mod support;
```
sanbox-irl, line 1: should we show off using `imgui::Window::new`? I think probably no, yeah?

imgui-examples/examples/id_wrangling.rs
```diff
@@ -0,0 +1,45 @@
+mod support;
```
sanbox-irl, line 1: Hahah, love that you made an example out of this. Thanks so much!

imgui/src/window/child_window.rs
```diff
@@ -18,12 +18,13 @@ pub struct ChildWindow<'a> {
     border: bool,
 }
 
-impl<'a> ChildWindow<'a> {
```
sanbox-irl, line 21: Ah we gotta revert these changes here. For *some* reason, ChildWindows uniquely take `Id` for their `name` instead of `label`. This might be because child windows don't showcase that label anywhere, but is rather bizarre.

How about we leave the bounds on `new` as is (this is how most will consume the API), but let's do `new_with_id` that uses the old `Into<Id>`. Internally, we'll use the `Id` match that we did previously (wrapping the `new` impl AsRef<str> param into an `Id::Str`), and then match on build and go from there?

sanbox-irl (changes requested):
ah, just that one child window change with the id stack, and it's good to go! Sorry about all the annoying changes -- your meticulous examples and documentation just makes everything so perfect that I hate to request specific changes. I think if we lose that `Id` support, though, we'll be breaking some code that...somehow might rely on it.

I think I'll do an `Id` pass after this PR is accepted and remove all the `Into<Id>` bounds to just be `Id`, since `Id` is weird and asking users to make it themselves seems appropriate

imgui/src/window/child_window.rs
dbr: Ah, interesting - BeginChild takes either str or ID,

https://github.com/ocornut/imgui/blob/fc4988ffb048f7d3ebcf9cb96bf7727878ad9d0e/imgui.cpp#L5093-L5103

and both call `BeginChildEx` which explains

https://github.com/ocornut/imgui/blob/fc4988ffb048f7d3ebcf9cb96bf7727878ad9d0e/imgui.cpp#L5061

> If you need to append to a same child from multiple location in the ID stack, use BeginChild(ImGuiID id) with a stable value

..which I understand as meaning `ID` is essentially a global address to a child window, whereas the string option only works relative to the current window.

So.. makes sense, the `new_with_id` thing seems a good approach

imgui-examples/examples/creating_windows.rs
dbr: Might as well mention it as long as its available, even if just in a "probably don't use this" way

imgui-examples/examples/id_wrangling.rs
```diff
@@ -0,0 +1,45 @@
...
+                }
+            });
+
+        ui.window("Good Example")
+            .position([400.0, 0.0], imgui::Condition::FirstUseEver)
+            .size([390.0, 200.0], imgui::Condition::FirstUseEver)
+            .build(|| {
+                ui.text("Works!");
+                for it in &items {
+                    let _label_id = ui.push_id(it);
```
thomcc, line 33: Hm, this usage suggests to me we should consider making the result of push_id deref into `ui`, so that you can do `let ui = ui.push_id(...);` (which avoids the footguns of `let _ = ui.push_id(...)` or forcing you to come up with a dummy name).

imgui/src/window/child_window.rs
dbr: Uh this makes perfect sense but I have no idea how to do it.

Since ChildWindow would now stores `'&'ui Ui` and an `&'id Id` I end up with something like
```rust
    pub fn child_window<Label: 'ui + AsRef<str>>(&'ui self, name: Label) -> ChildWindow<'ui, '_> {
        ChildWindow::new(self, name)
    }

    pub fn child_window_with_id(&'ui self, id: Id<'id>) -> ChildWindow<'ui, 'id> {
        ChildWindow::new_with_id(self, id)
    }
```

..but in trying to implement these bits on ChildWindow I get baffling errors like

```
error[E0495]: cannot infer an appropriate lifetime for lifetime parameter `'id` due to conflicting requirements
  --> imgui/src/window/child_window.rs:40:43
   |
40 |         let mess: ChildWindow<'ui, 'id> = ChildWindow {
   |                                           ^^^^^^^^^^^
   |
note: first, the lifetime cannot outlive the lifetime `'id` as defined on the impl at 21:11...
  --> imgui/src/window/child_window.rs:21:11
   |
21 | impl<'ui, 'id> ChildWindow<'ui, 'id> {
   |           ^^^
note: ...so that the expression is assignable
  --> imgui/src/window/child_window.rs:42:13
   |
42 |             id,
   |             ^^
   = note: expected `Id<'_>`
              found `Id<'id>`
note: but, the lifetime must be valid for the lifetime `'ui` as defined on the impl at 21:6...
  --> imgui/src/window/child_window.rs:21:6
   |
21 | impl<'ui, 'id> ChildWindow<'ui, 'id> {
   |      ^^^
note: ...so that the expression is assignable
  --> imgui/src/window/child_window.rs:40:43
   |
40 |           let mess: ChildWindow<'ui, 'id> = ChildWindow {
   |  ___________________________________________^
41 | |             ui,
42 | |             id,
43 | |             flags: WindowFlags::empty(),
...  |
48 | |             border: false,
49 | |         };
   | |_________^
   = note: expected `child_window::ChildWindow<'ui, 'id>`
              found `child_window::ChildWindow<'_, '_>`
```

..which I think is https://github.com/rust-lang/rust/issues/65866 which ends with a not overly reassuring 2 year old comment "Every time I see this message it makes my head explode. So I think my journey with rust will end today" :grimacing:

sanbox-irl: hahahaha holy shit. okay, I'll accept it as it is now, and then I'll schwoop it and try to get more functionality. I can do it at work today

imgui-examples/examples/id_wrangling.rs
sanbox-irl: Not only do I agree with that, I have been working on EXACTLY this change to the entire library.

this is the sort of thing that would be required to have proper mutability support for `Ui` (ie, some methods will take `&mut self` and others will take `&self`). For example, it occured to me that if the drop token for the table's api was also where you access all of the table's methods, then we solve https://github.com/imgui-rs/imgui-rs/issues/524 at the type level. 
There are issues and i'm not sure we'll get that complex to start with, but I'm in total agreement here on that idea

PR #538, opened 2021-09-25
Docking branch as cargo feature

dbr:
[Yes](https://github.com/imgui-rs/imgui-rs/pull/207), [another](https://github.com/imgui-rs/imgui-rs/pull/285), [docking](https://github.com/imgui-rs/imgui-rs/pull/391), [PR](https://github.com/imgui-rs/imgui-rs/pull/476), but.. this time as a optional feature

1. There is both `imgui-sys/third-party/imgui-master/` and `imgui-docking`
2. bindgen is run for each (along with the existing variants for `wasm`)
3. Similar to the wasm handling, `imgui-sys/src/lib.rs` conditionally imports the appropriate one
4. There's a few struct members/flags/methods are switched based on `#[cfg(feature = "docking")]` but not actually too much difference
5. There's a rough binding to the basic parts of the docking API I need (pretty much just splitting/docking windows to create an initial default layout), although it needs updated to task `AsRef<str>` instead of `&ImStr` as it was copied directly from my old 0.7.0 docking fork

A few other random things:
1. I experimented with ditching the submodules, by instead using `git subtree`. I don't have any strong opinions either way; subtree seems nicer mostly just because the lack of `submodule init`, but remarkably the `git subtree` CLI seems even clunkier than `git submodule`. I'm not that sure either way has much benefit over just copying `imgui.cpp`/`imgui.h` etc into repo
2. We should probably lock down the versions of bindgen/cimgui in some way
3. This would need a bunch of cleanup and improvements before merging, if it seems like a sensible direction to go on.

Thoughts?


sanbox-irl:
@thomcc would you mind taking a look? Both of you know the imgui-sys backend better than I do

PR #540, opened 2021-09-28
Implement `imgui-sdl2-support` crate.

NightShade256:
This PR implements a SDL 2 support crate for imgui-rs, similar to the existing `imgui-winit-support` crate. This is mainly due to the fact that the existing SDL 2 support crate is severely outdated and a "official" crate is beneficial. This also checks off the last point from the list at the end of #444.

Currently most things except locked/rounded DPI are implemented, and I don't think there is a straightforward way to implement that. Please let me know if there is something that has to be changed, or can be improved.

### DIFF
--- a/imgui-sdl2-support/Cargo.toml
+++ b/imgui-sdl2-support/Cargo.toml
@@ -11,4 +11,9 @@ categories = ["gui"]
 
 [dependencies]
 imgui = { version = "0.8.1-alpha.0", path = "../imgui" }
-sdl2 = "0.34"
+sdl2 = "^0.34.5"
+
+[dev-dependencies]
+glow = "0.10.0"
+imgui-glow-renderer = { version = "0.8.1-alpha.0", path = "../imgui-glow-renderer" }
+sdl2 = { version = "^0.34.5", features = ["bundled", "static-link"] }
--- a/imgui-sdl2-support/Cargo.toml
+++ b/imgui-sdl2-support/Cargo.toml
@@ -11,3 +11,4 @@ categories = ["gui"]
 
 [dependencies]
 imgui = { version = "0.8.1-alpha.0", path = "../imgui" }
+sdl2 = "0.34"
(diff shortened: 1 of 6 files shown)


sanbox-irl:
hey thanks so much! this looks pretty great. I'm going out to a show tonight but will make sure to see it tomorrow night. i think it needs to get rebased from main (i just pushed some changes), but otherwise looks great just skimming it

imgui-sdl2-support/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+categories = ["gui"]
+
+[dependencies]
+imgui = { version = "0.8.1-alpha.0", path = "../imgui" }
+sdl2 = "^0.34.5"
+
+[dev-dependencies]
+glow = "0.10.0"
+imgui-glow-renderer = { version = "0.8.1-alpha.0", path = "../imgui-glow-renderer" }
+sdl2 = { version = "^0.34.5", features = ["bundled", "static-link"] }
```
thomcc, line 19: I'll take a more thorough look later (at work now) but we shouldn't be turning either of these features on for people (bundled is particualrly bad because it has UB on 64-bit windows ([for example](https://github.com/Rust-SDL2/rust-sdl2/blob/master/sdl2-sys/sdl_bindings.rs#L951) long is 32 bit on win64 but size_t is 64 bit)

Also `^` is the default so you don't need to specify it.

sanbox-irl: It's just for the examples

thomcc: Ah, right, i missed this was under dev-dependencies.

sanbox-irl: didn't know about that UB though!

NightShade256:
Opening a new PR since I forgot to checkout to a new branch before implementing the crate, sorry!

PR #490, opened 2021-05-28
Add double input and allow setting display format

EmbersArc:
Closes #489.

PR #544, opened 2021-10-02
Add `input_scalar` and allow setting float input precision

EmbersArc:
Continuation of https://github.com/imgui-rs/imgui-rs/pull/490

I haven't been able to test it yet since the examples fail to build on the main branch.

### DIFF
--- a/imgui/src/input_widget.rs
+++ b/imgui/src/input_widget.rs
@@ -1,7 +1,9 @@
 use bitflags::bitflags;
 use std::ops::Range;
 use std::os::raw::{c_char, c_int, c_void};
+use std::ptr;
 
+use crate::internal::DataTypeKind;
 use crate::math::*;
 use crate::sys;
 use crate::Ui;
@@ -514,11 +516,12 @@ impl<'ui, 'p, L: AsRef<str>> InputInt<'ui, 'p, L> {
 }
 
 #[must_use]
-pub struct InputFloat<'ui, 'p, L> {
+pub struct InputFloat<'ui, 'p, L, F = &'static str> {
     label: L,
     value: &'p mut f32,
     step: f32,
     step_fast: f32,
+    display_format: Option<F>,
     flags: InputTextFlags,
     ui: &'ui Ui,
 }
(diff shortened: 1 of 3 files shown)


EmbersArc:
Should we add `InputDouble` as well? Might be redundant when `InputScalar` is available. Maybe `InputFloat*` and `InputInt*` are not required anymore either.

imgui/src/input_widget.rs
```diff
@@ -530,19 +533,36 @@ impl<'ui, 'p, L: AsRef<str>> InputFloat<'ui, 'p, L> {
             value,
             step: 0.0,
             step_fast: 0.0,
+            display_format: None,
```
sanbox-irl, line 536: we should default this display format to `"%.3f"`, as it originally was. That also means that display_format can actually become `F` instead of `Option<F>`,  and start life with the type `&'static str`

imgui/src/input_widget.rs
```diff
@@ -572,22 +593,35 @@ macro_rules! impl_input_floatn {
...
+                    flags: self.flags,
+                    ui: self.ui,
+                }
+            }
+
             pub fn build(self) -> bool {
                 let value: $MINT_TARGET = (*self.value).into();
                 let mut value: [f32; $N] = value.into();
 
+                let (one, two) = self
```
sanbox-irl, line 619: same thing with the display format as above

imgui/src/lib.rs
```diff
@@ -551,6 +551,30 @@ impl<'ui> Ui {
     {
         InputInt4::new(self, label, value)
     }
+    #[doc(alias = "InputScalar")]
```
sanbox-irl, line 554: Can we add some doc comments here explaining what `InputScalar` is, and maybe in particular how it differs from the other `Input` types (something about basically "let's you do this with u64s and other data types than float or int").

sanbox-irl (commented):
Looks great! I think if you rebase and then run `cargo fmt --all` we should be basically good to go!

sanbox-irl:
> Should we add `InputDouble` as well? Might be redundant when `InputScalar` is available. Maybe `InputFloat*` and `InputInt*` are not required anymore either.

`InputDouble`, I feel is fine if we don't cover for now. f64s aren't super commonly used in `DearImGui`, but if you feel like putting in the effort...you're welcome to! I'd accept it, but until someone asks, I'm not too worried about it. `InputFloat` COULD be removed actually, but `InputInt`'s widget has the `+/-` buttons that this widget doesn't have, and so we'll want to keep that around.

Personally, as long as Dear ImGui has both, confusingly, I'm fine with us having both too. It's a bit odd, but since we can do this all with generics, our `InputScalar` is pretty good looking and nice to use, compared to Dear ImGui's input scalar. I think we could likely move towards `InputFloat` being a facade for `InputScalar`. `InputFloatX` though lets you use MintTypes, which we can't get `InputScalarN` to do, at least not in the meantime I think.

imgui/src/input_widget.rs
EmbersArc: Just for the float case I presume? For `InputScalar` I'd definitely stick with letting imgui pick the format.

EmbersArc:
> InputInt's widget has the +/- buttons that this widget doesn't have, and so we'll want to keep that around.

Are you sure about that? Maybe it still shows up when `step` is set explicitly. I'd try it but still can't compile it, it fails with the message

```
the trait `imgui_glium_renderer::glium::backend::Facade` is not implemented for `glium::Display`
```

EmbersArc:
Sounds good to me. Thanks again!

PR #551, opened 2021-10-12
menu rewrite

sanbox-irl:
Moved the Menu-Item constructors onto `ui`.

I'm pleased with how this API looks, so I'm going to write this out. If @dbr and @thomcc also agree with this kind of design, I'll add this to our docs and aim to keep this format for most of our overload-style code.

1. `ui.menu_item` creates a `menu_item` with just a label, and nothing else. This function returns a bool directly -- this means that users don't have to write `.build` at all. This allows for fast code to be written easily.
2. `ui.menu_item_x` are made for common overloads of a single value. So for `menu_item`, I bound all of them the various parameters, since they're common enough -- so we have a `ui.menu_item_shortcut`, `ui.menu_item_selected`, `ui.menu_item_enabled`, etc. These, like the main simple one, return *bool*, not the Builder struct, so users shouldn't have to type `.build` themselves.
3. `ui.menu_item_config` takes the same parameters as our basic `ui.menu_item`, but returns a Builder instead (ie, users need to type `.build()` at the end). For less common situations, we can use this builder to generate complex forms.

This should be generalizable to the vast majority of our builder structures, and in my code, made this as easy as it could be in a language without optional arguments. Optional arguments would be nice...

What do you think? imo, it would be nice if at least MOST of the library worked in a similar way

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -24,7 +24,7 @@
 
 - Added `add_polyline` method to `DrawListMut`, which binds to Dear ImGui's `AddPolyline` and `AddConvexPolyFilled`
 
-- BREAKING: `MenuItem::new` now takes `&ui`, but has been deprecated. Instead, use `ui.menu_item`.
+- BREAKING: `MenuItem::new` now takes `&ui`, but has been deprecated. Instead, use `ui.menu_item`. Additionally, a `ui.menu_item_config` has been created to access the builder pattern on `MenuItem`.
 
 ## [0.8.0] - 2021-09-17
 
--- a/imgui-examples/examples/test_window_impl.rs
+++ b/imgui-examples/examples/test_window_impl.rs
@@ -342,22 +342,22 @@ fn show_test_window(ui: &Ui, state: &mut State, opened: &mut bool) {
                 menu.end();
             }
             if let Some(_t) = ui.begin_menu("Examples") {
-                ui.menu_item_selected_ref("Main menu bar", &mut state.show_app_main_menu_bar);
-                ui.menu_item_selected_ref("Console", &mut state.show_app_console);
-                ui.menu_item_selected_ref("Log", &mut state.show_app_log);
-                ui.menu_item_selected_ref("Simple layout", &mut state.show_app_layout);
-                ui.menu_item_selected_ref("Property editor", &mut state.show_app_property_editor);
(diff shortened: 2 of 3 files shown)


dbr:
I do like that:
1. Menu built off `ui.` is much neater and easier to find from auto-complete
2. The simple case is simple - `if ui.menu_item("blah") { ... }` is good

Main thing I question is the value of having the multiple "shortcut" methods (menu_item_enabled/selected/selected_ref etc), because:

1. There is quite reasonable people would want to use combination of the two, like a "disable-able" menu item with a shortcut (and obviously then adding  `menu_item_enabled_and_shortcut` and every other combination becomes silly)
2. Say I'm using `menu_item("Thing")` then realize I need to make that menu disabled on `$condition`, or want to add a shortcut to it, etc

    If I have some code `if ui.menu_item_config("Thing").build()` - I think it is easier to discover I can do `ui.menu_item_config("Thing").shortcut("...").build() {` than if the code was `ui.menu_item("Thing") {` and I need to change it to `ui.menu_item_config("Thing").build() {` *then* add the shortcut builder argument.
3. Similarly, if there is a bunch of `menu_item_selected`, `menu_item_enabled` etc - as someone reading the API docs I wouldn't assume there was another way of making a menu item with both `selected` and `enabled`. There is some value in nudging people towards the builder for any "non default" usage

So maybe we just have `ui.menu_item("Label")` and the builder-based `ui.menu_item_config("...")`? Generalised this would mean:

1. Everything builds of `ui.`
2. If there is a clear "default usage", there is a simple construction method (like `ui.button("...")`, `ui.menu_item("...")`, `ui.same_line()` etc)
3. If there is more complex usages:

    a) If the alternative uses are all exclusively non-overlapping (e.g you can make a button with a default size, or a button with a specific size - nonsensical to want to do both at same time) there is only nonbuilder methods (`ui.button` and `ui.button_size`)
    b) Otherwise, there is a builder method (plus the "simple construction menu", if applicable)

sanbox-irl:
seems very reasonable to me! In my case, I often end up using those other options for menu items and wanted to avoid the `build` call, but perhaps I was just being annoying. Increasing library space for Rust differences from C++ seems like a not good idea, so I think I'll implement your idea

sanbox-irl:
going to merge when this goes green

PR #559, opened 2021-11-01
[WIP] D3D12 renderer for imgui-rs

curldivergence:
Hey, I've implemented a simple D3D12 renderer using my D3D12 bindings [rusty-d3d12](https://github.com/curldivergence/rusty-d3d12) and would like to share it. It is based on both [imgui-dx11-renderer](https://github.com/veykril/imgui-dx11-renderer) and C++ D3D12 renderer from [here](https://github.com/TheCherno/Hazel1HourGame/blob/master/Hazel/vendor/imgui/examples/imgui_impl_dx12.cpp). Unfortunately I'm not very familiar with ImGUI so some parts may be missing, but the bundled example seems to work without any issues.
Of course I'm not sure if such a change is welcome, so please share your thoughts :) Thanks!

### DIFF
--- a/imgui-d3d12-renderer/Cargo.toml
+++ b/imgui-d3d12-renderer/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "imgui-d3d12-renderer"
+version = "0.8.1-alpha.0"
+authors = ["Andrey Pushkar <andrey.s.pushkar@gmail.com>"]
+edition = "2021"
+description = "D3D12 renderer for the imgui crate"
+homepage = "https://github.com/curldivergence/imgui-d3d12-renderer"
+repository = "https://github.com/curldivergence/imgui-d3d12-renderer"
+license = "MIT/Apache-2.0"
+readme = "README.md"
+categories = ["gui", "rendering"]
+
+[dependencies]
+rusty-d3d12 = "0.2.4"
+imgui = "0.8"
+thiserror = "1.0"
+log = "0.4"
+memoffset = "0.5"
+
+[dev-dependencies]
+imgui = "0.8"
+imgui-winit-support = "0.8"
+winit = "0.25"
+clap = "2.33"
+simple_logger = "1.6"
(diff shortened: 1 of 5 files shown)


curldivergence:
Hi, just as an FYI, I've also tried running [imnodes-rs](https://github.com/benmkw/imnodes-rs) with this renderer, and it went smoothly, too :) Please let me know if you want to have a look or have any questions. Thanks!

sanbox-irl:
Hi!

I appreciate your work here, but I don't think we'll be accepting the PR right now. The core issue is that once we accept this PR, we (as in myself, dbr, and thomcc, the other two maintainers) have to maintain the new sub-crate renderer. Since `rusty-d3d12` is still a young crate and doesn't have wide usage yet, I'm against us taking on that maintenance burden, as we won't have the users to justify the time-sink we'll need to spend.

We actually only maintain our crate within our repo for the purposes of promising users that these backends are up to date, but otherwise, they don't have access to any private imgui functionality. Therefore, for now, would you consider publishing this as a separate crate in its own repository, and if rusty-d3d12 gets a bigger user-base over time, we can consider adding it to this repository then?

curldivergence:
Hi @sanbox-irl, thanks for your reply, I see your point :) However, may I kindly ask if you'd be open to adding a section named like `Non-official renderers` or `Experimental 3rd-party renderers` in the readme and mentioning my project and @Veykril's [imgui-dx11-renderer](https://github.com/Veykril/imgui-dx11-renderer) (or [it's recently updated fork by @Netdex](https://github.com/Netdex/imgui-dx11-renderer)) so that people at least would know about these options? :) Thanks!

sanbox-irl:
That's a great idea -- when I add that, I'll ping here too

curldivergence:
Thank you! I'll close the PR then.

curldivergence:
Hi @sanbox-irl, any chance this is still on the table? :) Thanks!

sanbox-irl:
closed with https://github.com/sanbox-irl/imgui-rs/commit/3166550958f144069c8489db5eda41a49d4d1b11

Thank you!

Issue #566, opened 2021-11-15
How to build freetype feature on Windows (in GH Actions)


dbr:
I've tried to build with the Freetype feature on Windows in GH Actions and got stuck, as I'm not sure how best to make library available

Upstream imgui [only tests freetype on Linux](https://github.com/ocornut/imgui/blob/d80a9123b741dc7d99cee76cbbb047cfb7d91292/.github/workflows/build.yml#L213-L258), so I can't use that as an example :cry:

We could just do the same, but it'd be nice to keep the testing consistent

Anyone have any ideas of how to achieve this?

ctrlcctrlv:
I tend to get the DLL's from here: https://github.com/ubawurinna/freetype-windows-binaries/tree/master/release%20dll/win64

It's linked on freetype.org: https://freetype.org/download.html

After that the process is identical to linking any DLL.

dbr:
Ah useful, thanks! Will try with these soon

ruifengx:
FYI, [vcpkg](https://github.com/microsoft/vcpkg) has a port for `freetype`, and there is [a Rust binding to `vcpkg`](https://docs.rs/vcpkg/latest/vcpkg/) (which looks relatively easy to use). I personally have been using vcpkg for package management on Windows for a while, and as I have just tested, it works without problems when instructed to install `freetype`.

Some advantages of using vcpkg:
+ all versions of the library are available: static/shared, debug/release, as well as static/shared CRT. This is configurable using target triplets.
+ support custom features for packages during installation. Below is part of the output from the command line, and available features are listed in square brackets:
    ```
    freetype                 2.11.0#2         A library to render fonts.
    freetype[brotli]                          Support decompression of WOFF2 streams
    freetype[bzip2]                           Support bzip2 compressed fonts.
    freetype[png]                             Support PNG compressed OpenType embedded bitmaps.
    freetype[zlib]                            Use zlib instead of internal library for DEFLATE
    ```
    For example, to install `freetype` with features `png` and `zlib` enabled, we use `vcpkg install freetype[png,zlib]`.

One possible drawbacks compared to prebuilt DLLs: it builds the library locally instead of using prebuilt versions, so installation time is expected to be slightly longer.

PR #567, opened 2021-11-15
Test docking feature in CI

dbr:
Continuing on #557 - build and run tests with the `docking` feature

I also want to add in a build using the `freetype` feature but got stuck, shall split that out to another issue, https://github.com/imgui-rs/imgui-rs/issues/566

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,8 +45,12 @@ jobs:
             ${{ runner.os }}-target-lint-stable-${{ hashFiles('**/Cargo.toml') }}
             ${{ runner.os }}-target-lint-stable-
             ${{ runner.os }}-target-lint-
+
+      # default
       - run: cargo clippy --workspace --all-targets
-      # supported winit versions
+      # with only docking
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -132,12 +137,12 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-target-test-${{ matrix.rust }}-${{ hashFiles('**/Cargo.toml') }}
             ${{ runner.os }}-target-test-${{ matrix.rust }}-
+
       - run: cargo test --workspace --all-targets
+      - run: cargo test --workspace --all-targets --features docking
```
thomcc, line 142: There's a thing in the xtask to do the tests like CI does, can we mirror this change there? (It's possible now that I'm not making many changes to this stuff that code isn't worth keeping, though. I wouldn't fight removing it if nobody uses it)

dbr: Ah that's a good point! I almost wonder if it's worth calling `cargo xtask test` for the CI, but obvious downside is each step wouldn't show up as nicely in the Actions UI :thinking:

Issue #565, opened 2021-11-13
`slider` example is broken on macos


dymk:
**Describe the bug**

When running the `slider` example, a crash is observed when selecting the "Example 1: Basic sliders" option. 

Here's a snippet of the stacktrace of the crash: https://gist.github.com/dymk/4a4dd569c2f035a8aebf7e2bb44c5bf8

The specific assertion failing is this:

```cpp
IM_ASSERT(*(const float*)p_min >= -FLT_MAX / 2.0f && *(const float*)p_max <= FLT_MAX / 2.0f);
```

The values of:
- `p_min`: `(const float) $0 = -3.40282347E+38`
- `p_max`: `(const float) $1 = 3.40282347E+38`
- `FLT_MAX` from `float.h`: `$2 = 3.40282347E+38`

In particular, the second clause of the above assertion is what's failing (cmp has been set to FLT_MAX):

```
p *(const float*)p_min >= -cmp / 2.0f
(bool) $8 = true
p *(const float*)p_max <= cmp / 2.0f
(bool) $9 = false
```

**To Reproduce**

Run slider example, select Basic sliders option

**Expected behavior**

It to not do that

**Screenshots**

**Please describe your environment**
 - imgui-rs version 0.8.0
 - macOS 11.3.1 (20E241)

**Other information**
Add related issues, suggestions about how to fix, any other context here.

dbr:
Same crash occurs on Linux, and doesn't happen in 0.7

    slider: ./third-party/imgui-master/imgui/imgui_widgets.cpp:2974: bool ImGui::SliderBehavior(const ImRect&, ImGuiID, ImGuiDataType, void*, const void*, const void*, const char*, ImGuiSliderFlags, ImRect*): Assertion `*(const float*)p_min >= -3.40282346638528859811704183484516925e+38F / 2.0f && *(const float*)p_max <= 3.40282346638528859811704183484516925e+38F / 2.0f' failed.

Pretty sure this is caused by the change to remove the (slightly insane) default ranges from `imgui::Slider`,

https://github.com/imgui-rs/imgui-rs/commit/7b644bf159#diff-9a0a75de1c60eb3869ca58aa756e5d200ff93fa919ef3c302144810e18f2550f

The library change itself is good I think, just the example code needs fixed, as

    Slider::new(im_str!("f32 value")).build(ui, &mut state.f32_value);

was changed to

[...]

PR #568, opened 2021-11-16
Update slider example #565

dbr:
I've updated the slider example to use the limited range, however the pointlessly-giant range slider doesn't seem to work correctly:

sliderbug

Not too sure what this might be caused by - any idea @sanbox-irl ?

It might be an upstream ~problem, as the "Dear ImGui Demo" `Widgets > Data types > Sliders` section doesn't include a "slider float full" example (only a "high" which goes from 1e-10 to 1e10). If so we should probably just remove this full-range-f32 from teh example as it's a bit academic

### DIFF
--- a/imgui-examples/examples/slider.rs
+++ b/imgui-examples/examples/slider.rs
@@ -49,7 +49,7 @@ fn example_1(ui: &Ui, state: &mut State) {
 
         // However for larger data-types, it's usually best to specify
         // a much smaller range. The following slider is hard to use.
-        Slider::new("Full range f32 value", -f32::MIN/2.0, f32::MAX/2.0)
+        Slider::new("Full range f32 value", f32::MIN/2.0, f32::MAX/2.0)
             .build(ui, &mut state.f32_value);
         // Note the `... / 2.0` - anything larger is not supported by
         // the upstream C++ library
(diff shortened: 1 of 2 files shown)


dbr:
I've tested the `-FLT_MAX/2.0 .. FLT_MAX/2.0` ranged slider in Dear ImGui itself and it works fine, so something is amiss here

dbr:
> I've tested the `-FLT_MAX/2.0 .. FLT_MAX/2.0` ranged slider in Dear ImGui itself and it works fine, so something is amiss here

Interestingly recreating the same basic example in imgui-rs also works completely fine:

```rust
use imgui::*;

mod support;

fn main() {
    let mut value: f32 = 0.0;
    system.main_loop(move |run, ui| {
        Slider::new("f32 value with range", -f32::MAX/2.0, f32::MAX/2.0)
            .build(ui, &mut value);
    });
}
```

Something about how the slider.rs example specifically is going wrong

dbr:
:facepalm:

I wrote `-f32::MIN/2.0` is "quite similar" to `f32::MAX/2.0` so the slider in example is acting completely correctly (as far as can be expected for a slider to and from the same value). I think I meant to write `-f32::MAX/2.0, f32::MAX/2.0`, and then completely missed the difference been `MIN` and `MAX` when comparing to the working example above!

Good news being everything is working as expected!

PR #570, opened 2021-11-20
Fix DPI scaling in examples

dbr:
Extending from #504,

1. Using `HiDpiMode::Default` is indeed a good default
2. As noted in comments (and in comment on the linked PR), imgui-winit-support + winit handles most of the scaling for us
3. However imgui's default bitmap font then looks a bit naff (most likely why `Rounded` was used in first place), so I've changed the default font to Roboto as that was already in the repo
4. I've enabled oversampling on this font as it helps with the default rasterizer
5. The font rendering still then doesn't look too great due to imgui-glium-renderer's lack of proper compositing, so I've kept the "multiplier" value (but nudged down a little as I think it looks "better")

Overall I think this makes the DPI scale handling about as close to correct as it can be - so for example if I force the DPI scale to 1, and 6, it just looks like a 6x scale:

scaling example

overlay

Point 5 does kind of hints we should either add some easy sRGB mode into glium-renderer, or switch the examples to using imgui-glow-renderer (which already does the correct thing by default)

### DIFF
--- a/imgui-examples/examples/support/mod.rs
+++ b/imgui-examples/examples/support/mod.rs
@@ -46,15 +46,41 @@ pub fn init(title: &str) -> System {
     {
         let gl_window = display.gl_window();
         let window = gl_window.window();
-        platform.attach_window(imgui.io_mut(), window, HiDpiMode::Rounded);
+
+        let dpi_mode = if let Ok(factor) = std::env::var("IMGUI_EXAMPLE_FORCE_DPI_FACTOR") {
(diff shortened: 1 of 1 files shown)


dbr:
Only thing I've not really tested is what happens when you drag a window between different DPI screens - winit emits an event for these, but I'm not sure if there is additional work needed

sanbox-irl:
yeah i'd say this is totally fine. Nice work

PR #571, opened 2021-11-23
Add a StyleVar for CellPadding

Sejsel:
This allows setting a table-related style var which was introduced in imgui 1.80.

I also added cell padding to some related tests when I was at it. There's a couple of other style variables that not consistently tested, but that's for another time.


dbr:
Thanks! Just needs the slightly tedious `cargo fmt` step fixed and I think this is good to merge

Sejsel:
I was sure I ran it... oh well. Should be fixed now, thanks!

PR #574, opened 2021-11-27
Add missing fields to memory layout tests

Sejsel:
Adds fields introduced since ImGui 1.80 to memory layout tests.

These tests were incredibly useful for me recently, but they seem to be easy to forget.


thomcc (approved):
LGTM, nice catch.

PR #577, opened 2021-11-29
Update xflags to 0.2.x

thomcc:
New `xflags` compiles faster, is a lot less janky, is easier to update (no more "copy this into the file"), and no longer causes r-a to suffer quite as much distress.

So there's really no downsides. Actually, I thought we had already done this.


dbr:
LGTM!

PR #578, opened 2021-12-03
imgui-winit-support: Add basic touch input

hrydgard:
Ignores the complexities of multitouch, but makes imgui usable directly through touch events.

If an app doesn't want this, it can easily filter out Touch events before passing their event stream to imgui-winit-support.

(EDIT: I also just realized that apps could simply convert touch events to mouse on their own before passing events to imgui-winit-support, so if I do that, this patch is no longer really necessary, making this low priority, but still might be useful to someone).

### DIFF
--- a/imgui-winit-support/src/lib.rs
+++ b/imgui-winit-support/src/lib.rs
@@ -909,6 +909,15 @@ impl WinitPlatform {
                     _ => (),
                 }
             }
+            WindowEvent::Touch(touch) => {
+                let position = self.scale_pos_from_winit(window, touch.location);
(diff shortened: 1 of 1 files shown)


sanbox-irl:
@hrydgard did you end up doing the above idea, and if so, is it better than accepting this PR?

hrydgard:
I did implement it (converting events) and it works for my purposes. Given that imgui is not particularly touch-friendly anyway, maybe it's not really motivated to have it in the repo.

sanbox-irl:
thank you! hopefully if someone searches around, they can find this

PR #579, opened 2021-12-03
Remove NO_ALPHA flag insertion in ColorPicker4::build

kazatsuyu:
`ColorEditFlags::NO_ALPHA` is used in `ColorPicker4::build` function, so there is no way to edit alpha value.  
It was probably copied from `ColorPicker3::build` and should be removed.

### DIFF
--- a/imgui/src/widget/color_editors.rs
+++ b/imgui/src/widget/color_editors.rs
@@ -886,8 +886,7 @@ where
     /// Builds the color picker.
     ///
     /// Returns true if the color value was changed.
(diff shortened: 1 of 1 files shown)


dbr:
Ahh I briefly confused the ColorEdit with the ColorPicker! `ColorPicker4` which is the "inline color swatch" widget, and it indeed is permanently missing the alpha without this fix:

no_alpha_bug

Looks good, shall merge once all the CI bits are happy! Thanks!

Issue #562, opened 2021-11-07
`freetype` feature does not compile on Windows: tries to call non-existent `pkg-config`


ctrlcctrlv:
```
error: failed to run custom build command for `imgui-sys v0.8.1-alpha.0 (https://github.com/imgui-rs/imgui-rs#6faa7f09)`

Caused by:
  process didn't exit successfully: `D:\a\glif\glif\target\debug\build\imgui-sys-98982d076e541193\build-script-build` (exit code: 101)
  --- stdout
  cargo:THIRD_PARTY=C:\Users\runneradmin\.cargo\git\checkouts\imgui-rs-61ccdc9370f94e57\6faa7f0\imgui-sys\third-party
  cargo:DEFINE_IMGUI_USE_WCHAR32=
  cargo:DEFINE_CIMGUI_NO_EXPORT=
  cargo:DEFINE_IMGUI_DISABLE_WIN32_FUNCTIONS=
  cargo:DEFINE_IMGUI_DISABLE_OSX_FUNCTIONS=
  cargo:rerun-if-env-changed=FREETYPE2_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=FREETYPE2_STATIC
  cargo:rerun-if-env-changed=FREETYPE2_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_pc_windows_msvc
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-pc-windows-msvc
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64_pc_windows_msvc
[...]

PR #582, opened 2021-12-07
add vcpkg-rs for locating freetype

ruifengx:
Should solve #562 and #566.

I used the method I proposed [here](https://github.com/imgui-rs/imgui-rs/issues/566#issuecomment-987711170) (using `vcpkg` in addition to `pkg-config`). It builds and links properly with my local settings, but I'm not entirely sure whether or not it runs correctly (the program looks correct, though).

This PR chose to base on the tag `v0.8.0` (instead of `HEAD`) on purpose, so that it is possible to test and compare the result with the latest published release on `crates.io`. If needed, I can rebase it onto `HEAD`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,17 +136,23 @@ jobs:
       - name: docking feature
         run: cargo test --workspace --all-targets --features docking
 
-      - name: freetype feature
+      - name: freetype feature (non-Windows, pkg-config)
         if: matrix.os != 'windows-latest'
         run: cargo test --workspace --all-targets --features freetype
 
-      - name: all features
+      - name: freetype and docking (non-Windows, pkg-config)
         if: matrix.os != 'windows-latest'
-        run: cargo test --workspace --all-targets --features docking,freetype
+        run: cargo test --workspace --all-targets --features freetype,docking
 
-      - name: doc tests
-        run: cargo test --workspace --doc
+      - name: freetype feature (Windows, vcpkg)
+        if: matrix.os == 'windows-latest'
+        run: cargo test --workspace --all-targets --features freetype,use-vcpkg
 
+      - name: freetype and docking (Windows, vcpkg)
+        if: matrix.os == 'windows-latest'
+        run: cargo test --workspace --all-targets --features freetype,docking,use-vcpkg
+
+      - run: cargo test --workspace --doc
       # run to check for lint problems
       - name: build documentation
         run: cargo doc
--- a/imgui-sys/Cargo.toml
+++ b/imgui-sys/Cargo.toml
@@ -39,4 +39,5 @@ vcpkg = { version="0.2.15", optional=true }
 default = []
 wasm = []
 docking = []
-freetype = ["pkg-config", "vcpkg"]
+freetype = ["pkg-config"]
+use-vcpkg = ["vcpkg"]
--- a/imgui-sys/Cargo.toml
+++ b/imgui-sys/Cargo.toml
@@ -33,9 +33,10 @@ cfg-if = "1"
 [build-dependencies]
 cc = "1.0"
 pkg-config = { version="0.3", optional=true }
+vcpkg = { version="0.2.15", optional=true }
 
 [features]
 default = []
 wasm = []
-freetype = ["pkg-config"]
 docking = []
+freetype = ["pkg-config", "vcpkg"]
--- a/imgui-sys/build.rs
+++ b/imgui-sys/build.rs
@@ -9,11 +9,34 @@ const DEFINES: &[(&str, Option<&str>)] = &[
     ("IMGUI_DISABLE_OSX_FUNCTIONS", None),
 ];
 
+// Output define args for compiler
 fn main() -> std::io::Result<()> {
     // Root of imgui-sys
     let manifest_dir = std::path::Path::new(env!("CARGO_MANIFEST_DIR"));
 
-    // Output define args for compiler
+    #[cfg(feature = "freetype")]
+    fn find_freetype() -> Vec<impl AsRef<std::path::Path>> {
+        let err_pkg_config;
+        let err_vcpkg;
+        match pkg_config::Config::new().find("freetype2") {
+            Ok(freetype) => return freetype.include_paths,
+            Err(err) => err_pkg_config = err,
+        }
+        match vcpkg::find_package("freetype") {
+            Ok(freetype) => return freetype.include_paths,
+            Err(err) => err_vcpkg = err,
+        }
+        panic!(
+            "cannot find freetype:\n\
+        - pkg-config failed with: {}\n\
+        - vcpkg failed with: {}",
+            err_pkg_config, err_vcpkg
+        );
+    }
+    println!(
+        "cargo:THIRD_PARTY={}",
+        manifest_dir.join("third-party").display()
+    );
     for (key, value) in DEFINES.iter() {
         println!("cargo:DEFINE_{}={}", key, value.unwrap_or(""));
     }
@@ -48,9 +71,7 @@ fn main() -> std::io::Result<()> {
         // Freetype font rasterizer feature
         #[cfg(feature = "freetype")]
         {
-            // Find library
-            let freetype = pkg_config::Config::new().find("freetype2").unwrap();
-            for include in freetype.include_paths.iter() {
+            for include in find_freetype() {
                 build.include(include);
             }
             // Set flag for dear imgui
--- a/imgui-sys/build.rs
+++ b/imgui-sys/build.rs
@@ -11,22 +11,16 @@ const DEFINES: &[(&str, Option<&str>)] = &[
 
 #[cfg(feature = "freetype")]
 fn find_freetype() -> Vec<impl AsRef<std::path::Path>> {
-    let err_pkg_config;
-    let err_vcpkg;
+    #[cfg(not(feature = "use-vcpkg"))]
     match pkg_config::Config::new().find("freetype2") {
         Ok(freetype) => return freetype.include_paths,
-        Err(err) => err_pkg_config = err,
+        Err(err) => panic!("cannot find freetype: {}", err),
     }
+    #[cfg(feature = "use-vcpkg")]
     match vcpkg::find_package("freetype") {
         Ok(freetype) => return freetype.include_paths,
-        Err(err) => err_vcpkg = err,
+        Err(err) => panic!("cannot find freetype: {}", err),
     }
-    panic!(
-        "cannot find freetype:\n\
-    - pkg-config failed with: {}\n\
-    - vcpkg failed with: {}",
-        err_pkg_config, err_vcpkg
-    );
 }
 
 // Output define args for compiler
--- a/imgui-sys/build.rs
+++ b/imgui-sys/build.rs
@@ -9,30 +9,31 @@ const DEFINES: &[(&str, Option<&str>)] = &[
     ("IMGUI_DISABLE_OSX_FUNCTIONS", None),
 ];
 
+#[cfg(feature = "freetype")]
+fn find_freetype() -> Vec<impl AsRef<std::path::Path>> {
+    let err_pkg_config;
+    let err_vcpkg;
+    match pkg_config::Config::new().find("freetype2") {
+        Ok(freetype) => return freetype.include_paths,
+        Err(err) => err_pkg_config = err,
+    }
+    match vcpkg::find_package("freetype") {
+        Ok(freetype) => return freetype.include_paths,
+        Err(err) => err_vcpkg = err,
+    }
+    panic!(
+        "cannot find freetype:\n\
+    - pkg-config failed with: {}\n\
+    - vcpkg failed with: {}",
+        err_pkg_config, err_vcpkg
+    );
+}
+
 // Output define args for compiler
 fn main() -> std::io::Result<()> {
     // Root of imgui-sys
     let manifest_dir = std::path::Path::new(env!("CARGO_MANIFEST_DIR"));
 
-    #[cfg(feature = "freetype")]
-    fn find_freetype() -> Vec<impl AsRef<std::path::Path>> {
-        let err_pkg_config;
-        let err_vcpkg;
-        match pkg_config::Config::new().find("freetype2") {
-            Ok(freetype) => return freetype.include_paths,
-            Err(err) => err_pkg_config = err,
-        }
-        match vcpkg::find_package("freetype") {
-            Ok(freetype) => return freetype.include_paths,
-            Err(err) => err_vcpkg = err,
-        }
-        panic!(
-            "cannot find freetype:\n\
-        - pkg-config failed with: {}\n\
-        - vcpkg failed with: {}",
-            err_pkg_config, err_vcpkg
-        );
-    }
     println!(
         "cargo:THIRD_PARTY={}",
         manifest_dir.join("third-party").display()
--- a/imgui-sys/build.rs
+++ b/imgui-sys/build.rs
@@ -13,12 +13,12 @@ const DEFINES: &[(&str, Option<&str>)] = &[
 fn find_freetype() -> Vec<impl AsRef<std::path::Path>> {
     #[cfg(not(feature = "use-vcpkg"))]
     match pkg_config::Config::new().find("freetype2") {
-        Ok(freetype) => return freetype.include_paths,
+        Ok(freetype) => freetype.include_paths,
         Err(err) => panic!("cannot find freetype: {}", err),
     }
     #[cfg(feature = "use-vcpkg")]
     match vcpkg::find_package("freetype") {
-        Ok(freetype) => return freetype.include_paths,
+        Ok(freetype) => freetype.include_paths,
         Err(err) => panic!("cannot find freetype: {}", err),
     }
 }
--- a/imgui-sys/third-party/imgui-docking/cimgui.cpp
+++ b/imgui-sys/third-party/imgui-docking/cimgui.cpp
@@ -4927,7 +4927,11 @@ CIMGUI_API void igDebugRenderViewportThumbnail(ImDrawList* draw_list,ImGuiViewpo
 }
 CIMGUI_API const ImFontBuilderIO* igImFontAtlasGetBuilderForStbTruetype()
 {
+#ifdef IMGUI_ENABLE_FREETYPE
+    return static_cast<const ImFontBuilderIO*>(0);
+#else
     return ImFontAtlasGetBuilderForStbTruetype();
+#endif
 }
 CIMGUI_API void igImFontAtlasBuildInit(ImFontAtlas* atlas)
 {


dbr:
This looks great, thanks! Will give it a proper test as soon as I can - so far I've only checked on Linux and it does as expected (happily falls back to pkgconfig)

ctrlcctrlv:
I really hope this can be merged soon, MFEKglif is still stuck on imgui-rs 0.7.0 and the branch has diverged a whole bunch yet again, I'd like to be able to upgrade it to 0.8.0 finally. (_Cf._ https://github.com/MFEK/glif/pull/128#issuecomment-997449670)

sanbox-irl:
hey @ruifengx this looks excellent, but I'm having some trouble in the CI here. You can see on my fork my current progress:
https://github.com/sanbox-irl/imgui-rs/runs/4611636419?check_suite_focus=true

sanbox-irl:
No time pressure, but this is probably the last thing we need merged in before I can do a 0.8.1 patch!

ruifengx:
@sanbox-irl `vcpkg` does not automatically install missing packages on build (especially when this build is triggered by Cargo as is the case here). You need a separate `vcpkg install freetype` in advance just like on Linux you need a `dnf install` (or `apt install`, etc.).

BTW, in the current design, we always try `pkg-config` first, and then `vcpkg`. So if both are available, the `freetype` available via `pkg-config` is always preferred. In such cases, to manually switch to the `vcpkg` version, `pkg-config` should be temporarily removed from `PATH`. I am not entirely sure whether or not this would be a problem.

sanbox-irl:
Hmm! That seems less useful to end users then, huh? I was thinking this would work like how sdl2's vcpkg works, where it DOES fetch it for you

sanbox-irl:
I think i'd confused this with their `bundled` feature! I'd definitely like to have that

sanbox-irl:
For now, i'm going to merge this in and fix up the error messages, change some of hte feature gates (I'd rather you opt into vcpkg *and* freetype, rather than automagically getting both, since vcpkg seems to be much more common on windows) and then merge it in. Seperately, I'll make an issue for a `bundled` feature for freetype, which looks trivial for us to do

sanbox-irl:
@ruifengx 
Actually, adding `vcpkg install freetype` didn't work! I'm going to leave my work on this here, but here's what this PR needs to be accepted:

A. Add some testing for this feature to our CI, specifically in our Windows builds. For that, see lines 133-150.
1. Uncomment lines 133-135.
2. Remove the if guard on lines 145 and 149.
3. Somehow edit lines 133-135 such that this correctly builds!

B. Separate this feature into a `use-vcpkg` feature. I want users to opt into this rather than have it silently happen to them -- I'm concerned about cascading errors which will confuse people. So if you wanted to use just `freetype` and `pkg-config`, you'd just enable the `freetype` feature -- if you want `freetype` and find it with `vcpkg`, then use `freetype,use-vcpkg`. In the future, we'll have a "bundled" feature, which will be XOR with `use-vcpkg`, and I want to have that clearly differentiated now for that time in the future.

Let me know when you've got those done, and I'll merge it it, or let me know if you'd rather drop your own work on this, and when I'm back home for the holidays and have a Windows machine available, I can investigate this myself! (I'm on a Mac till the new year)

ruifengx:
Hi @sanbox-irl, sorry for the delayed response. I see plenty of exclamation marks in your comments; please be assured that I get your frustration on a PR failing your expectation, and that the failure is due to complicated real-life situations rather than poor design.

> That seems less useful to end users then, huh? I was thinking this would work like how sdl2's `vcpkg` works, where it DOES fetch it for you

> I think i'd confused this with their bundled feature! I'd definitely like to have that

I am afraid this is not the case, and I also believe generally it is _not_ a wise decision to automatically install the dependency (in this case `freetype`) **system-wide** (however, I will definitely leave this decision to you maintainers of `imgui-rs`). To support a `bundled` feature, you will need a git submodule or something similar in your source tree (so that the dependency remain local, instead of conflicting with the global package installation on the target machine). In such circumstances, I believe the whole compilation process could be managed by [`cmake`](https://lib.rs/crates/cmake). Worth-mentioning though, using a bundled `freetype` rules out the possibility for custom compile-time options (to enable or disable various features in `freetype`), so I personally prefer a globally-installed customisable version for my specific use case.

> Actually, adding `vcpkg install freetype` didn't work!

I will take some time investigating this tonight. I did confirm that the approach taken here works correctly on my PC before submitting this PR, so I guess there is something misconfigured in the CI process.

> A. Add some testing for this feature to our CI, specifically in our Windows builds. For that, see lines 133-150.
>    1. Uncomment lines 133-135.
>    2. Remove the if guard on lines 145 and 149.
>    3. Somehow edit lines 133-135 such that this correctly builds!

OK, get it. I will try fixing this CI script and report my progress soon.

> B. Separate this feature into a `use-vcpkg` feature. I want users to opt into this rather than have it silently happen to them -- I'm concerned about cascading errors which will confuse people. So if you wanted to use just `freetype` and `pkg-config`, you'd just enable the `freetype` feature -- if you want `freetype` and find it with `vcpkg`, then use `freetype`, `use-vcpkg`. In the future, we'll have a "bundled" feature, which will be XOR with `use-vcpkg`, and I want to have that clearly differentiated now for that time in the future.

I did not feature-gate `use-vcpkg` because `pkg-config` was not feature-gated as well, and I had some misunderstandings on your opinions about this. I agree a `use-vcpkg` feature is clearly a better design, and I would like to go further and suggest also feature-gating `use-pkg-config`. However, feature-gating `use-pkg-config` would be a semver-breaking change (because this update would break packages depending on `imgui-rs` with default features disabled). In my opinion, we should have three mutually-exclusive features: `freetype-via-vcpkg`, `freetype-via-pkg-config`, and `freetype-bundled`. (Again, this is just a suggestion, and I leave the final decision to you.)

> Let me know when you've got those done, and I'll merge it it, or let me know if you'd rather drop your own work on this, and when I'm back home for the holidays and have a Windows machine available, I can investigate this myself! (I'm on a Mac till the new year)

I will try to get a properly-updated version in several days.

sanbox-irl:
Oh hey, no frustration on my end -- your work is great and I appreciate your contribution greatly! I'm just of the age where I can't help but put `!` on everything.

> I did not feature-gate use-vcpkg because pkg-config was not feature-gated as well, and I had some misunderstandings on your opinions about this. I agree a use-vcpkg feature is clearly a better design, and I would like to go further and suggest also feature-gating use-pkg-config. However, feature-gating use-pkg-config would be a semver-breaking change (because this update would break packages depending on imgui-rs with default features disabled). In my opinion, we should have three mutually-exclusive features: freetype-via-vcpkg, freetype-via-pkg-config, and freetype-bundled. (Again, this is just a suggestion, and I leave the final decision to you.)

I agree with the above -- but unfortunately, we're a bit locked in without making a breaking change. I think perhaps in v0.10.0 we can make a major goal to have better freetype management for users. For now, just getting something that works for the majority of users will be sufficient I suspect.

Thank you for the effort -- I don't know nearly enough about this to get it done on my own, so your input is essential

ctrlcctrlv:
I really really don't think it's a problem to have `vcpkg` required in this PR, then another PR to add _both_ the `use-pkg-config` and `use-vcpkg`. It won't matter on Linux that `vcpkg` will never be available because `pkg-config` will work. I agree with the version of this PR as submitted by @ruifengx.

sanbox-irl:
fair point, but I would like CI figured out before merge

ctrlcctrlv:
Oh, that's not really a major problem. This commit doesn't touch CI. You need to add something like `friendlyanon/setup-vcpkg@v1` (https://github.com/friendlyanon/setup-vcpkg) then add a command like this

```yaml
      - if: matrix.platform == 'windows'
        name: Setup dependencies (Windows)
        run: vcpkg install freetype
```

ctrlcctrlv:
Sorry, I know you are waiting or him to reply, I just want this fixed :-P

sanbox-irl:
I actually did try that! If you see my above posts, I did exactly that, but it failed in the CI on my fork.

I didn't use that linked action though since windows runners come with vcpkg pre installed. That looks like it mostly handles caching -- does it also do some other work that might fix the issues I ran into?

sanbox-irl:
And don't worry, any help is good help and appreciated @ctrlcctrlv

ctrlcctrlv:
Oh, I'm sorry, I didn't understand fully the CI problem.

I think you need this also: https://crates.io/crates/cargo-vcpkg, which claims to interoperate with `vcpkg`

Then in `Cargo.toml`:
```
[package.metadata.vcpkg.target]
x86_64-pc-windows-msvc = { triplet = "x64-windows-static", dependencies = ["freetype"] }
```

ctrlcctrlv:
Also I am doubting a little bit that the Windows GitHub CI comes with vcpkg.

The opencv-rust project, well maintained, installs it themselves:

https://github.com/twistedfall/opencv-rust/blob/master/.github/workflows/opencv-rust.yml
https://github.com/twistedfall/opencv-rust/blob/master/ci/install-windows-vcpkg.sh

ctrlcctrlv:
RustDesk project also installs own `vcpkg`. Why do we think `vcpkg` globally included?

https://github.com/rustdesk/rustdesk/blob/b3e3f6151d537a009dbbb8ef695c6fe13fd45527/.github/workflows/ci.yml#L86

Their `ci.yml` also mentions `$VCPKG_ROOT` envvar.

ctrlcctrlv:
Oh, I just found something interesting, even if you are right about vcpkg bundling with Windows GH CI, there is this in the `vcpkg` crate docs:

```rust
//! # Environment variables
//!
//! A number of environment variables are available to globally configure which
//! libraries are selected.
//!
//! * `VCPKG_ROOT` - Set the directory to look in for a vcpkg installation. If
//! it is not set, vcpkg will use the user-wide installation if one has been
//! set up with `vcpkg integrate install`, and check the crate source and target
//! to see if a vcpkg tree has been created by [cargo-vcpkg](https://crates.io/crates/cargo-vcpkg).
```

So you may have to run `vcpkg integrate install`, but really I think that it's more Rusty (is this a word?) to use `cargo-vcpkg` and not rely on any bundling, as who knows, they might drop it when M$ decides they have a new "solution" to problems solved on Linux almost 3 decades hence, as they tend to do every few years.

thomcc:
We should not use cargo-vcpkg in ci if we can avoid it, as cargo installing a binary in ci tends to be very slow

ctrlcctrlv:
That's not a good argument because you can just cache `%USERPROFILE%\.cargo\bin` (Win32 version of Linux `$HOME/.cargo/bin`).

thomcc:
Caching is pretty dodgy at best, snd installing a program to avoid running a single command is… iffy, imo

ctrlcctrlv:
There's nothing iffy about GH Actions caching, and if you don't trust it, you could upload a Win32 binary of cargo-vcpkg.exe somewhere, and fetch it on a cache miss. 

Given it is only some 5MB, probably would even fit in a GitHub gist.

```
$ ls -alh ~/.cargo/bin/cargo-vcpkg 
-rwxr-xr-x 1 fred fred 4.9M Dec 25 02:01 /home/fred/.cargo/bin/cargo-vcpkg
```

There are benefits to users of using `vcpkg` `Cargo.toml` section, because it means upstream project CI will work too. If something is done ad hoc just in your ci.yml, such as what opencv-rust is doing, then that'd need to be copied by all.

Regardless, current imgui 0.8 is very broken on Win32, so I'd take iffy over broken, given it will only affect Windows CI build.

ruifengx:
@sanbox-irl It seems I managed to make the CI pass: https://github.com/ruifengx/imgui-rs/runs/4631059708. Our first attempt was indeed in the correct direction, it is just Rust expects to be linked with 64bit static libraries with the MSVC runtime dynamically linked (thus the `x64-windows-static-md` triplet).

I will update the CI script first, and then start working on feature-gating `use-vcpkg`.

dbr:
> RustDesk project also installs own `vcpkg`. Why do we think `vcpkg` globally included?

The default software for the Windows runners is listed here, and does include vcpkg:
https://github.com/actions/virtual-environments/blob/7518d04a3a12a7fb369179b0e77d4d89c3db9b0f/images/win/Windows2022-Readme.md#package-management

Not sure when this was added, it might be a fairly recent thing thus RustDesk etc installed it manually?

> have three mutually-exclusive features: freetype-via-vcpkg, freetype-via-pkg-config, and freetype-bundled

I think the bundling flag is a good convenience (maybe to do later), but I'm not too sure it's worth the minor extra complexity of having the two `freetype-via-` flags - is there a clear use-case for this?

My thinking is;

1. vcpkg-rs and pkg-config-rs are both quite lightweight dependencies (both just small codebases with no dependencies, they just call the relevant commands as a subprocess), so shouldn't add any significant compile-time overhead
2. It seems very unlikely someone would a) have both installed, and b) have freetype available in both, and c) care specifically which one is used

ruifengx:
I am not necessarily for or against separate feature flags for `pkg-config` and `vcpkg`. I agree that dependency on `vcpkg` in build scripts should be lightweight enough.

I see several compile errors appearing. The lint error should be as easy to fix as removing the redundant return statement, but errors related to the `docking` feature might require some further investigation. I will report back as soon as I find out the cause.

ruifengx:
Weird enough, I cannot reproduce those errors locally on my machine.

The error mentions unresolved external symbol for `ImFontAtlasGetBuilderForStbTruetype`. After some investigations, it seems that this function is guarded by the macro `IMGUI_ENABLE_STB_TRUETYPE` in [`imgui_draw.cpp`](https://github.com/ocornut/imgui/blob/512c54bbc062c41c74f8a8bd8ff1fd6bebd1e6d0/imgui_draw.cpp#L2596-L2603). And according to [these lines](https://github.com/ocornut/imgui/blob/512c54bbc062c41c74f8a8bd8ff1fd6bebd1e6d0/imgui_internal.h#L104-L106), `IMGUI_ENABLE_STB_TRUETYPE` is only automatically defined if `IMGUI_ENABLE_FREETYPE` is not defined.

So, in theory, `ImFontAtlasGetBuilderForStbTruetype` becomes undefined after enabling the `freetype` feature here, but in `cimgui`, the binding to `ImFontAtlasGetBuilderForStbTruetype` (named `igImFontAtlasGetBuilderForStbTruetype`) is [unconditional](https://github.com/cimgui/cimgui/blob/afce852875e36e948951d6253963cea46101c17c/cimgui.cpp#L4579). Therefore I would expect link errors occur on every platform, but I cannot even reproduce the errors on my PC... (Maybe some link mechanism is taking effect here.)

There are several things I can come up with that might be different for my local configurations and the CI container: (1) I am using Visual Studio 2022 17.0.2, and (2) I might have a newer `vcpkg` and `freetype`.

ruifengx:
I believe it is safe to conclude that the problem is from upstream in `cimgui`. (Though I am still confused by the whole situation.) Have a look at [this CI run](https://github.com/ruifengx/imgui-rs/runs/4631866086), it turns out [this one line change](https://github.com/ruifengx/imgui-rs/commit/572873bf5734d22624e84658a83a0bf6b80227e7#diff-a82895948ba19f1abd1006f02066150b9c6a206d14a2d017ec50ef68b3d52469) fixes the Windows build: and all it does is remove the reference to `ImFontAtlasGetBuilderForStbTruetype`.

The proper way to fix this might be to guard function `igImFontAtlasGetBuilderForStbTruetype` behind a feature macro `IMGUI_ENABLE_STB_TRUETYPE` in `cimgui`.

sonoro1234:
> The proper way to fix this might be to guard function `igImFontAtlasGetBuilderForStbTruetype` behind a feature macro `IMGUI_ENABLE_STB_TRUETYPE` in `cimgui`.

The proper way to fix this might be to guard function `ImFontAtlasGetBuilderForStbTruetype` behind a feature macro `IMGUI_ENABLE_STB_TRUETYPE` in `imgui_internal.h`. Besides adding `IMGUI_ENABLE_FREETYPE` in generator.lua when `freetype` option is enabled.

ruifengx:
> The proper way to fix this might be to guard function `ImFontAtlasGetBuilderForStbTruetype` behind a feature macro `IMGUI_ENABLE_STB_TRUETYPE` in `imgui_internal.h`.

Seems this issue is gradually propagating through the dependency chain. :laughing:

So shall we open an issue ticket in `imgui`, or should we just wait for a response in [imgui/193](https://github.com/cimgui/cimgui/issues/193)?

sonoro1234:
> So shall we open an issue ticket in `imgui`, or should we just wait for a response in [cimgui/193](https://github.com/cimgui/cimgui/issues/193)?

Better wait as @ocornut has been notified there.

ocornut:
Pushed a fix on my side to not mislead the generator:
https://github.com/ocornut/imgui/commit/980deb4c9eaf0702e97be1309d9cd10aaf8467af

sanbox-irl:
if i'm understanding this right then, we'll need to wait for hte upstream imgui change to trickle down before our CI can work with both features enabled?

ctrlcctrlv:
No. As of https://github.com/imgui-rs/imgui-rs/commit/b1150d6c3b8f4df2977e538fee765308662684a6 by @dbr imgui-rs copies imgui upstream into it, so that just needs to be updated to post ocornut/imgui@980deb4 because imgui isn't dynamically linked by imgui-sys

sonoro1234:
Imgui change is done in máster
Do you need docking?

sanbox-irl:
We do yeah. @dbr can you jumó in on this? And @ctrlcctrlv we do that but we don't grab random commits -- we're grabbing releases.

sonoro1234:
Meanwhile you can add to cimgui_manuals in generator.lua the function to be skipped in generation

dbr:
> [...] we do that but we don't grab random commits -- we're grabbing releases.

For imgui-rs 0.9, I reckon we can ~manually backport the https://github.com/ocornut/imgui/commit/980deb4 fix to the vendored copies of imgui

However in 0.8 it's a bit trickier as that version still uses the git-submodule, but if tweaking the cimgui generator would work, then that sounds like a good solution here

ctrlcctrlv:
@sanbox-irl 

how is this still open

should i just port my code to egui and be done with it because "we don't use random commits"

imgui-rs 0.8 has significant unfixed breakage and really i'm starting to doubt this project's priorities…

another option i could take is to PR a way to replace your entire imgui-sys with a link to `libimgui.so`. @ocornut feels that dynamic links to imgui are almost always overengineering (cf. https://github.com/ocornut/imgui/issues/980#issuecomment-272821010) — incredibly we may have found the exception

ocornut:
My post suggested that i encourage people (expectially c++ users) to compile themselves as it allows configuring stuff like compatible math types. Binaries built by others won’t do that. Do whatever works for you.

A friend of mine started rewriting cimgui because it appears like in its current iteration it became a bit of a bottleneck for progress for users of non C++ languages, if you are interested we can share the wip repo. (Our version currently output enough metadata to output headers with full comments etc)

ctrlcctrlv:
yes, very interested, as I experimented myself already with compiling libcimgui to .so

i believe it works

image

i'm frustrated with lack of progress in this project, so if i could PR a way to tell imgui-sys where to find a .so and just call `cargo:link` then all this back and forth over imgui version becomes something i can do internally at MFEK

thomcc:
> so if i could PR a way to tell imgui-sys where to find a .so and just call cargo:link then all this back and forth over imgui version becomes something i can do internally at MFEK

The better way to do this (which doesn't require you wait for a release even) is to just override it in the `[target.$target.imgui]` section of a `.cargo/config.toml` (note: not cargo.toml), as documented here. https://doc.rust-lang.org/cargo/reference/config.html#targettriplelinks. The key to use for the links part is from https://github.com/imgui-rs/imgui-rs/blob/main/imgui-sys/Cargo.toml#L12

ctrlcctrlv:
it's not something you want to expose at all, or not something you want to expose by default?

ctrlcctrlv:
oh i see what you're saying. with that `[links]` i could link a dylib libcimgui.so with few lines in toml, no need for PR here. i get it

ruifengx:
Glad to hear this! I could rebase this PR onto the latest `main` to resolve the conflicts. Please inform me when it is appropriate to do so. (Presumably after Steps 1 and 2 are done?)

dbr:
@ruifengx Sorry for the protracted delay, but.. I think this is now ready to rebase (v0.9 is out the door), if you have the time!

ruifengx:
I have just rebased my fork to `main`. I have confirmed on my Windows 11 PC that it now passes all the tests with `cargo test --features docking,freetype,use-vcpkg`.

**Note:** To link properly with MSVC for `docking` + `freetype`, I had to modify `cimgui.cpp`. I need to stub `igImFontAtlasGetBuilderForStbTruetype` if `IMGUI_ENABLE_FREETYPE` is defined, because otherwise the MSVC linker will complain about non-existing symbol. If my understanding is correct, the fix upstream in `cimgui` (cimgui/cimgui#193) requires regenerating the C binding every time the feature combination is changed, which I believe is not compatible with the settings here. Please do confirm [this change](https://github.com/ruifengx/imgui-rs/blob/7eec3e96d49cbc09f580843413aff49932b0f548/imgui-sys/third-party/imgui-docking/cimgui.cpp#L4928-L4935) is acceptable before merging.

dbr:
Ah I think that manual edit to `cimgui.cpp` will no longer be necessary once #688 is merged (the upgrade to 1.89.1) as the upstream fix https://github.com/ocornut/imgui/commit/980deb4c9eaf0702e97be1309d9cd10aaf8467af is persent in this version

The upgrade PR is 99% done, so should be merged in a few days (barring some delays for the Christmas holidays) - then I think we can finally merge this - thanks for all the work!

ctrlcctrlv:
Good news for users of this library but I invested in switching [https://github.com/MFEK/glif](MFEKglif) off of `imgui-rs`, and onto `egui`, primarily due to this bug and now that that's set in motion won't claw it back. 

- lucasmerlin/egui_skia#3
- [**Add OpenGL/Vulkan SDL2 support**](https://github.com/lucasmerlin/egui_skia/compare/2787071c6433881f17e29af20fda6ae818fae359..70ffaebb26bd0cd8cd7fe35661d66f55711379ec)  lucasmerlin/egui_skia@`2787071`..lucasmerlin/egui_skia@`70ffaeb` 

Not mad about it, just saying that maintenance decisions have consequences downstream :-) perhaps this one is even fine / worth it on both sides.

dbr:
Thanks @ruifengx - merged this with the in-place edit, which will be effectively reverted via imminent #699

PR #581, opened 2021-12-07
Backport text-input bug fixes for 0.8.1

dbr:
I didn't quite realize there was over 100 commits since v0.8.0 (:partying_face:), so I might have missed something but I think this covers the bugfixes relating to the text input bugs

Backports the following changes:

- 5cf67c22d423559b6d0f0e67e92083db39c27ed4
- 6faa7f090a34ef16b5c3880d988caa95babb1aba
- daea06108f5d7c60742b8d71b4a0977585ce5a76

Relates to #564 #546 #553

Also there didn't seem to be any existing conventions for release-maintenance branch naming (I would normally use `maint-0.8` but interferes in minor way with tab-complete for `main`)

PR #583, opened 2021-12-17
Support winit-0.26

cwfitzgerald:
This adds winit 0.26 support to imgui-winit-support as non-default. This is a non-breaking change and I've tested all versions of winit with the crate.

I have bumped the version number, as it would be greatly appreciated a patch version could be released. This is currently the last blocking piece of updating my project to winit 0.26.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,7 @@ jobs:
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-23/default --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-24/default --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-25/default --all-targets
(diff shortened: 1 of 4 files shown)


sanbox-irl:
Thanks -- I need to release a patch anyway. I'll try to get that done on Sunday night

cwfitzgerald:
Great, thanks!

thomcc:
We should update these places too:
- https://github.com/imgui-rs/imgui-rs/blob/main/.github/workflows/ci.yml#L60
- https://github.com/imgui-rs/imgui-rs/blob/main/xtask/src/main.rs#L80

dbr:
Along with the additions Thom mentions - if the version number is left untouched it'd make it easier to merge in, then we can cherry-pick the change into the branch referenced in #581 (which includes my attempts at backporting the text-input bugfixs)

cwfitzgerald:
All three things changed!

sanbox-irl:
going to merge this in once it goes green

sanbox-irl:
thanks @cwfitzgerald !

thomcc:
We should remember to move from 0.25-by-default to 0.26-by-default on the next semver-major change of imgui-winit-support

PR #591, opened 2021-12-28
Minor fixes to README

LoipesMas:
I assume that's what you meant.


dbr:
`imgui-glow-renderer` does technically use the `imgui` crate, but... yep that definitely looks like a mistake - thanks!

PR #595, opened 2022-01-05
0.8 patch: Ensure misc/freetype/* is included in packaged crate

### LINKED ISSUE #589: 0.8.1 docs build failing
https://docs.rs/crate/imgui/0.8.1/builds/482032

```
[INFO] [stderr] error: Package `imgui v0.8.1 (/opt/rustwide/workdir)` does not have the feature `docking`
```

Took me a while to realize why: 0.8 doesn't have the docking feature (that will be part of 0.9)

### LINKED ISSUE #594: imgui-rs 0.8.2 docs failing to build
Hey there,

I was getting ready to read the docs for imgui 0.8.2 but noticed that the build was failing.

I looked at the build logs and it seems to be getting stuck on specifically imgui-sys 0.8.2, with a fatal error:
`fatal error: misc/freetype/imgui_freetype.h: No such file or directory`

Does imgui still use that header? Or perhaps the pathing is incorrect?

Anyways, hopefully, this helps solve the build. I may take a look through the repo and submit a pull request if I anticipate being able to fix it myself :)

dbr:
Required for "--features freetype" to work when using via crates.io

Closes #594
Closes #589

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,6 @@ jobs:
             ${{ runner.os }}-target-lint-
       - run: cargo clippy --workspace --all-targets
       # supported winit versions
-      - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --all-features --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-19 --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-20 --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-22 --all-targets
--- a/imgui-sys/Cargo.toml
+++ b/imgui-sys/Cargo.toml
@@ -11,7 +11,18 @@ categories = ["gui", "external-ffi-bindings"]
 build = "build.rs"
 links = "imgui"
 # exclude json, lua, and the imgui subdirs (imgui/examples, imgui/docs, etc)
-exclude = ["third-party/*.json", "third-party/*.lua", "third-party/imgui/*/"]
+# ..but we need imgui/misc/freetype/ for the freetype feature
+exclude = [
+    "third-party/*.json",
+    "third-party/*.lua",
+    "third-party/imgui/backends/",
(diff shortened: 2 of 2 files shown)


zachary-moote-personal:
Hi guys, 

I'm trying to investigate the cause behind the failing linter from
22 days ago (linter 1.54)
https://github.com/imgui-rs/imgui-rs/runs/4710306933?check_suite_focus=true

It seems that the linter is throwing errors with the public enum
template types(?) within winit-0.26.0, ie:

EventLoop<T: 'static>
EventLoopWindowTarget<T: 'static>
EventLoopProxy<T: 'static>
EventLoopWindowTarget<T>

Which are located in
/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.26.0/src/platform_impl/linux/mod.rs
and 
/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.26.0/src/event_loop.rs

It seems the error is due to unused parameters.

Is this something that could be solved on imgui-rs's end, 
or something I should bring up with the winit team?

Thank you!

ZM

dbr:
Thanks for looking into that! I only briefly glanced at the lint failures, didn't realise it was coming from winit, interesting..

I wonder if the "error: Please select a feature to build for unix: `x11`, `wayland`" error is the main one to worry about, and the subsequent unsed-parameters errors are side-effects of that as the mentioned parameter appear to be used, e.g:

https://github.com/rust-windowing/winit/blob/v0.26.0/src/event_loop.rs#L32-L33

dbr:
Hmmm, actually, the solution might be even simpler.. The failing clippy command is

    cargo clippy --manifest-path imgui-winit-support/Cargo.toml --all-features --all-targets

from here:

https://github.com/imgui-rs/imgui-rs/blob/6456066aca8c496297cf8844164f74703ad05cd7/.github/workflows/ci.yml#L48-L57

..which just isn't present in current main branch
https://github.com/imgui-rs/imgui-rs/blob/18ded1a58aafde5674f92a3124456d1123ce47d3/.github/workflows/ci.yml#L70-L78

So we should probably just remove (as done [here](https://github.com/imgui-rs/imgui-rs/commit/ce22eb3da59334781097c6ea48edc3a3777d5ccc))

Issue #441, opened 2021-02-04
Display size is wrong at startup


Hperigo:
**Describe the bug**
Hello, I´m trying to implement a glow backend for imgui-rs and I ran with an issue with the winit implementation that can be reproduced with the examples... 

The initial window size that is passed to the imgui platform is wrong, once there's a resize event, that gets corrected. But initially the projection matrix and clipping settings look wrong 

calling the code below before and during the infinite loop returns different results ( without changing the window size ): 
```
        let logical_size = windowed_context.window().inner_size().to_logical(scale_factor.into());
        let logical_size = platform.scale_size_from_winit(windowed_context.window(), logical_size);
        println!("logical size: {:?}", logical_size);
```

this is the initial value (same when creating the window): 
```LogicalSize { width: 1024.3019236547334, height: 768.0000276385623 }```

And in the update loop:
```LogicalSize { width: 1131.0000407021016, height: 848.0000305175793 }```

I did a quick test on mac and it looks like the opposite, the matrices are correct at startup, but once the window resizes it gets buggy.

**To Reproduce**
run the `test_window` example on windows, the imgui window should be with the wrong size and clipping, but once you resize the app window its corrected

**Expected behavior**
the Logical size should be consistent from the start so that the `draw_data.display_size` passed to the projection matrix is correct.

**Screenshots**
https://user-images.githubusercontent.com/3333460/106895109-95f3a180-66ce-11eb-8d7b-e538f6796bb4.mp4

**Please describe your environment**
 - imgui-rs version: Latest from master
 - Windows 10
 
**Other information**
Add related issues, suggestions about how to fix, any other context here.

sanbox-irl:
Hmm, can you give us some more information? I ran the examples on latest master and on some older builds (a bit after 0.6.1, we accidentally flubbed that release) and I'm not seeing the behavior you're describing. running on windows 10.

Hperigo:
hello,
It was kinda hard to explain with the screen capture so I did a movie with my cellphone.. 
https://user-images.githubusercontent.com/3333460/106971015-cd913680-672c-11eb-9010-453c44d3e549.MOV

You can see at the beginning of the video that the imgui window is moved to the top a bit, once I resize the window it snaps to the correct position and the mouse pointer interaction works correctly. 

But i think I found the solution, swapping `HiDpiMode::Rounded` to `HiDpiMode::Default` in `platform.attach_window` solved the problem... 

Let me know if you need more information, 
thanks

[...]

PR #608, opened 2022-02-01
winit-support: fix dpi handling in attach_window() when not in defaul…

filnet:
…t mode

attach_window() was using the imgui scale factor when converting
winit physical size to logical size instead of the winit scale factor.

This would cause the imgui display size to be incorrect when not using the
HiDpiMode::Default mode (in default mode the imgui and winit scale factors
are the same and the issue goes unnoticed).

Note that handle_window_event() has similar (and correct) logic when handling WindowEvent::Resized events.
This also explain why the issue went mostly undiscovered as winit tends to
generate quite a bunch of resize events that would hide the miscalculation
done in attach_window().

A similar issue was also fixed in the WindowEvent::ScaleFactorChanged handling code.

should fix https://github.com/imgui-rs/imgui-rs/issues/441


filnet:
Any chance to get this reviewed and merged ?

sanbox-irl:
hey @filnet sorry about that! I got this confused with my own winit related PR in my notifications.

thanks for the work!

PR #618, opened 2022-02-27
Remove im_str! usage from examples in docs

### LINKED ISSUE #617: Slider docs still reference im_str!
There are 3 examples in `slider.rs` which still show `im_str!` usage. Those should probably get updated.

Sejsel:
This removes leftover `im_str!` calls within examples in docs. I believe I got them all, the only remaining mentions of `im_str` are in the changelog and the macro definition itself.

Fixes #617.

### DIFF
--- a/imgui/src/draw_list.rs
+++ b/imgui/src/draw_list.rs
@@ -426,7 +426,7 @@ impl<'ui> DrawListMut<'ui> {
     /// # use imgui::*;
     /// fn custom_button(ui: &Ui, img_id: TextureId) {
     ///     // Invisible button is good widget to customise with image
-    ///     ui.invisible_button(im_str!("custom_button"), [100.0, 20.0]);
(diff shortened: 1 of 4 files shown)


Sejsel:
The failed check is not caused by these changes, it's some kind of cmake failure.

sanbox-irl:
Thanks so much! And yeah, this looks related to github updated windows-latest so that it used VS2022

dbr:
> And yeah, this looks related to github updated windows-latest so that it used VS2022

More specifically that the cache for the old `windows-latest` is reused for the new version, since the cache key is based on `runner.os` which is just "Windows". Surprisingly there doesn't seem to be a way to get a more specific OS-version string:

https://github.community/t/possible-to-get-specific-os-version-for-cache-keys/162282
https://github.com/actions/cache/issues/543

Only workaround is to specify the OS versions explicitly in the `matrix:` which isn't the worst thing, just something we'd need to occasionally remember to update

Issue #299, opened 2020-02-13
Windows 10: Hide encompassing window


pyroxymat:
So I running the example in Windows 10 (built myself). But I get my Windows screen wherein the IMGUI window is located. Is it possible to only get the IMGUI screen without the windows screen around it? Here is a picture of what I see:

![image](https://user-images.githubusercontent.com/3037921/74447286-67efce80-4e79-11ea-85fd-c24d8bff06b1.png)

sanbox-irl:
Yes. Sort of.

Assuming you're using Winit, maximize the size of your ImGui window to the size of your winit window. When you create your winit window, pass in `.with_decorations(false)` to the builder, and it will not show you the outer frame.

The outer frame *still exists* of course -- just you won't see it.

It's worth pointing out here though that ImGui isn't really for end users -- it's meant for developers. I wouldn't ship it as your UI

naiba:
![image](https://user-images.githubusercontent.com/29243953/159122783-d83a5419-d4d7-4e0d-9a04-0630e64fe896.png)

![image](https://user-images.githubusercontent.com/29243953/159122794-28dc41d6-6467-4194-bc01-ff7796d6eb40.png)

@sanbox-irl Can we hide the white background?

PR #619, opened 2022-03-04
Expose Viewport functionality

Rob2309:
This PR intends to expose most of the Viewport functionality of ImGui.

- Exposes the essential structs like PlatformIo, Viewport, Monitor, etc.
- Adds support for PlatformViewportBackend and RendererViewportBackend.
- imgui-winit-support mostly supports multiple Viewports, without breaking code changes to code that does not support Viewports.

The backend functions are implemented via two traits called `PlatformViewportBackend` and `RendererViewportBackend`.

I'm very open for suggestions for improvement, as I am not sure if everything I did uses the best approach.


imgui-examples/examples/hello_world.rs
```diff
@@ -4,26 +4,32 @@ mod support;
...
-        ui.window("Hello world")
-            .size([300.0, 110.0], Condition::FirstUseEver)
-            .build(|| {
-                ui.text_wrapped("Hello world!");
-                ui.text_wrapped("こんにちは世界！");
-                if ui.button(choices[value]) {
-                    value += 1;
-                    value %= 2;
-                }
+        if open {
```
dbr, line 13: Minor thing but should possibly put this in a different example as would be good to keep `hello_world` example simple

imgui-examples/examples/support/mod.rs
```diff
@@ -111,6 +113,46 @@ pub fn init(title: &str) -> System {
...
+struct ViewportStorage<'a, T: 'static> {
+    event_loop: &'a glium::glutin::event_loop::EventLoopWindowTarget<T>,
+    viewports: &'a mut HashMap<imgui::Id, glium::glutin::window::Window>,
+}
+
+impl<'a, T> imgui_winit_support::WinitPlatformViewportStorage for ViewportStorage<'a, T> {
+    fn create_window(&mut self, id: imgui::Id, flags: imgui::ViewportFlags) {
+        let builder = WindowBuilder::new()
+            .with_always_on_top(flags.contains(ViewportFlags::TOP_MOST))
+            // .with_decorations(!flags.contains(ViewportFlags::NO_DECORATION))
```
dbr, line 125: Seems like this should be uncommented

dbr (commented):
This is excellent, I'm excited to have viewport support :partying_face: 

Possibly silly question, but: I assume viewport support needs added to the renderers for this to do much?

My only concern is the changes to imgui-examples `support/mod.rs` - I think people starting with imgui-rs often copy that module and start tweaking it for their own purposes, and these changes make it more complex. However I think this is a completely separate issue, and I'll split this into a different issue!

Rob2309:
Yes the renderers and platform integrations will need to be extended in order to support viewports.

I would have liked to add viewport support to the examples directly, but at least with glium it is currently impossible to create new windows at runtime, although it seems like this will be fixed soon :)

I also thought about reverting all the changes to the examples and add a single example with viewport support, as this might be easier to understand. Consider the changes to the examples as temporary testing code that i forgot to remove before opening the PR...

imgui-winit-support/src/lib.rs
```diff
@@ -491,6 +609,109 @@ impl WinitPlatform {
...
+        imgui.set_platform_backend(ViewportBackend {});
+
+        let mut monitors = Vec::new();
+        for monitor in event_loop.available_monitors() {
+            monitors.push(imgui::PlatformMonitor {
+                main_pos: [monitor.position().x as f32, monitor.position().y as f32],
+                main_size: [monitor.size().width as f32, monitor.size().height as f32],
+                work_pos: [monitor.position().x as f32, monitor.position().y as f32],
+                work_size: [monitor.size().width as f32, monitor.size().height as f32],
+                dpi_scale: 1.0,
```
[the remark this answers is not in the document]
Rob2309: I am not quite sure what you mean by pulling?

imgui-examples/examples/viewports.rs
```diff
@@ -0,0 +1 @@
+fn main() {}
```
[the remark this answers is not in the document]
Rob2309: Yep, that was supposed to become the example with viewport support. Might add this when i have time next week.

imgui-examples/examples/hello_world.rs
Rob2309: True! Should probably be reverted

imgui-examples/examples/support/mod.rs
Rob2309: Yep, i just commented it out as without any rendering code, the undecorated windows would be invisible and it would have been hard to test...

Rob2309:
I was finally able to work on this a bit more :)

The examples should now be the same as before my PR again. The viewport example now implements a working viewport example using glow (since it still seems impossible to create glium windows in the event_loop).

qthree:
Last time I checked (year or two ago), glium had troubles with sharing of textures between GL contexts. Dropped texture is cleaned up only from parent context, but left dangling in shared contexts. Maybe there are some other bugs related to multiple windows.
wgpu worked fine with viewports, except that it leaked surface&swapchain on window destruction (pool of reusable windows could work workaround).

Rob2309:
> Last time I checked (year or two ago), glium had troubles with sharing of textures between GL contexts. Dropped texture is cleaned up only from parent context, but left dangling in shared contexts (which messes up internal texture indices). Maybe there are some other bugs related to multiple windows. wgpu [worked fine](https://github.com/qthree/viewports) with viewports, except that it leaked surface&swapchain on window destruction (pool of reusable windows could work as workaround).

Those issues just add onto the problems with Glium i guess. It is just impossible to create windows at runtime as Glium relies on EventLoop rather than EventLoopWindowTarget, but this might be fixed soon on crates.io as in https://github.com/glium/glium/pull/1997.

The reusable window pool idea might work but that is a very dirty hack in my opinion :(

Rob2309:
I will try to implement context sharing in the viewport example as soon as possible

dbr:
Side note: I put this under the v0.10 milestone, but no pressure whatsoever if it's not done by then or if you stop working on the PR!

Plus we have a very lose schedule and are still working on 0.9, so who knows when 0.10 will even be a thing (I just put it in the milestone to keep track of it)

naiba:
Hi there, BTW Can this PR achieve this goal? https://github.com/inkyblackness/imgui-go-examples/issues/1

Rob2309:
> Hi there, BTW Can this PR achieve this goal? https://github.com/imgui-rs/imgui-rs/issues/299 also https://github.com/inkyblackness/imgui-go-examples/issues/1
> 
> 

You can definitely make all extra viewports undecorated, but the user is responsible for handling the main window himself...

Rob2309:
> Side note: I put this under the v0.10 milestone, but no pressure whatsoever if it's not done by then or if you stop working on the PR!
> 
> Plus we have a very lose schedule and are still working on 0.9, so who knows when 0.10 will even be a thing (I just put it in the milestone to keep track of it)

We might need to do some more work before releasing this feature into the wild:
- The `platform_user_data` and `renderer_user_data` fields of viewports are currently raw pointers, there might be a more rust-like approach
- The imgui-winit-support crate currently assumes winit-26 when viewports are enabled

dbr:
> Hi there, BTW Can this PR achieve this goal? https://github.com/imgui-rs/imgui-rs/issues/299 also https://github.com/inkyblackness/imgui-go-examples/issues/1

I would assume the solution described in the imgui-go link should work here - basically make your "main imgui window" occupy the full (OS) window by doing something like

```rust
ui.window("Main")
    .flags(flags)
    .position([0.0, 0.0], imgui::Condition::Always)
    .size(ui.io().display_size, imgui::Condition::Always)
    .build(...)
```

Then hide the decoration on the OS-window via winit. Then make your other windows new viewports and it should do what you describe - just be a bunch of floating widnows

..at least, in theory!

> The imgui-winit-support crate currently assumes winit-26 when viewports are enabled

I think this is fine - since the viewport handling would be a new thing, I think it's reasonable it would require a recent winit version. We could just make it error if you try adn use old winit + viewport feature, but.. maybe even simpler,  I've made #627 to discuss if we need to support so many winit versions at all

naiba:
Yay, That works pretty well. But I can only resize the window, not move it. How can I make the window move freely?

As I understand it, imgui's window is restricted to the main window, which now cannot be moved. after this PR merge, the window will be free to move, and then it will be awesome. Is this right?

image

Rob2309:
> Yay, That works pretty well. But I can only resize the window, not move it. How can I make the window move freely?
> 
> 
> 
> As I understand it, imgui's window is restricted to the main window, which now cannot be moved. after this PR merge, the window will be free to move, and then it will be awesome. Is this right?
> 
> 
> 
> image
> 
> 
> 
> 
> 
> 

You might be able to just set the winit window position to your imgui window position every frame.  Not sure if this works very well though...

naiba:
> You might be able to just set the winit window position to your imgui window position every frame. Not sure if this works very well though...


```rust
// return window pos offset of "Hello world" window
let window_pos = run_ui(&mut run, ui);
if !run {
    *control_flow = ControlFlow::Exit;
}

let gl_window = display.gl_window();
let mut target = display.draw();
target.clear_color_srgb(0.0, 0.0, 0.0, 0.0);
platform.prepare_render(&ui, gl_window.window());

println!("window_pos {:?}", window_pos);

let draw_data = imgui.render();

let gl_window_pos = gl_window.window().outer_position().unwrap();
println!("gl_window_pos {:?}", gl_window_pos);

// set winit window position
gl_window.window().set_outer_position(PhysicalPosition {
    x: window_pos[0] + gl_window_pos.x as f32,
    y: window_pos[1] + gl_window_pos.y as f32,
});
```

```rust
// code of run_ui
 ui.window("Hello world")
    .flags(WindowFlags::NO_COLLAPSE)
    .position([0.0, 0.0], Condition::Once)
   .build(|| {
      // blablabla
      pos = ui.window_pos();
  });
pos
```

I have implemented the winit window movement according to your suggestion, but it keeps moving and I didn't find a way to reset the position of the imgui window after moving the winit window.  🌚

Rob2309:
When viewports are enabled, the imgui window position will be in desktop corrdinates, so you need to set the winit window position to the imgui window position as is, without adding winit position as offset :)

Rob2309:
It might be a good idea to have separate winit-support modules, one for non-viewport support and one with viewports enabled. Since the coordinate system is completely different depending on whether viewports are enabled, it might make the code more readable...

Rob2309:
I will try to work on a combined winit + glow viewports renderer that could serve as a complete example on how to use viewports :)

dbr:
Combining the winit+glow into one crate for the viewport sounds like a good idea to me!

With the "combined" crate, I would suggest only supporting the latest version of winit. The `imgui-winit-support` supports a confusingly large range of versions, which I don't think is necessary for the viewport handling (it seems reasonable to require someone using new viewport feature to use a recent version of winit), and we may drop the handling of so many versions in a future release (https://github.com/imgui-rs/imgui-rs/issues/627#issuecomment-1073143077)

imgui-winit-support/src/lib.rs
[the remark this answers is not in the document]
dbr: Oh I just mean that `dpi_scale` was hardcoded to `1.0` - should this number be queried in some way (`monitor.dpi_scale()` or something I would guess), or is fixing it to 1.0 the correct thing to do (if so a comment explaining why would be worthwhile)

Rob2309:
Thought so too, latest winit should be enough.

naiba:
Hi, long time no see, any news?

naiba:
Hi, long time no see, any news?

Rob2309:
Hey, I must admit I haven't done much on this lately, although I plan to continue working on this soon.

Rob2309:
So the last view days I worked on this again. I will probably have a working renderer soon.
The only big problem that is left to solve is proper DPI handling with multiple monitors. It seems like ImGui itself has not yet reached a point were proper DPI handling is entirely possible.

Rob2309:
Ok so the basic viewport rendering now works with a specialized winit-glow backend. It would be nice if anyone could confirm that this works on other PCs as well. Note though, that this backend is not very polished yet.
Most notably, there is no DPI handling at all, since there seems to be no viable way to handle multiple monitors with different DPIs in ImGui yet.

Latias94:
`imgui-winit-glow-renderer-viewports`'s `basic` example work well in my Win11 22H2 PC.

naiba:
Doesn't work properly on my macOS Ventura 13.1(intel). This is the error message: `thread 'main' panicked at 'Vertex Shader contains error'`

Rob2309:
> Doesn't work properly on my macOS Ventura 13.1(intel). This is the error message: `thread 'main' panicked at 'Vertex Shader contains error'`

Oh it seems like macOS does not support OpenGL 4.5. Will try with a lower version....

Rob2309:
It could work now on macOS, I removed everything that might require a new OpenGL version...

naiba:
> It could work now on macOS, I removed everything that might require a new OpenGL version...

Still not working properly
image

I modified cargo.toml to make it work
image

Rob2309:
> > It could work now on macOS, I removed everything that might require a new OpenGL version...
> 
> Still not working properly image
> 
> I modified cargo.toml to make it work image

Not quite sure how that fixes an error in the Vertex Shader :D

dbr:
I'm using Linux (Debian 11 with X11 and XFCE), and the "[Removed explicit version from shaders and not using new glsl features](https://github.com/imgui-rs/imgui-rs/pull/619/commits/070f56f1d1ed6a86b0f63b97ccfa8128f306e77b)" commit stops the example working for me, erroring at launch

```
thread 'main' panicked at 'Vertex Shader contains error', imgui-winit-glow-renderer-viewports/src/lib.rs:92:13
```

If I revert that commit, then it launches and works (! :partying_face: ), but has a few bugs:

a) Window resize events aren't handled properly, as in:

- Resizing using the imgui window controls works perfectly
- If I do something like right-click the window in my window list and do "maximize", the contents becomes distorted/streched (same with alt+rightmouse resizing, basically resizing any way except via imgui controls)

imgui-viewport-resize-bug


b) When I dock a window (from any window to any window), it will always crash with

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: XError { description: "GLXBadWindow", error_code: 170, request_code: 152, minor_code: 32 }', /home/dbr/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.27.5/src/platform_impl/linux/x11/window.rs:1086:14
```

c) In the example, the "Dear Imgui Demo" window appears to start in a slightly broken state in regards to mouse input.

I need to move the "Viewports window" out the way to interact with it at all. When done, the mouse is significantly offset. Once I move the window at all, it acts correctly. The Style Editor window acts correctly always

imgui-viewport-mouse-offset

d) Not necessarily a bug (maybe up to each application to handle?), but if I close one of the floating windows (e.g from widnow manager context menu), it is ignored and remains open

imgui-viewport-close

Those aside, this is really exciting and is great work, thanks!

Rob2309:
I will try to fix those things as soon as possible. Could you give me the error that happens by calling glow.get_shader_info_log?

Rob2309:
> 
> b) When I dock a window (from any window to any window), it will always crash with
> 
> ```
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: XError { description: "GLXBadWindow", error_code: 170, request_code: 152, minor_code: 32 }', /home/dbr/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.27.5/src/platform_impl/linux/x11/window.rs:1086:14
> ```
> 

I seem to be getting the same error when running with X11 on WSL. I currently have no idea where this is coming from, but the code around that panic in winit says:

> // This should be okay to unwrap since the only error XGetGeometry can return
> // is BadWindow, and if the window handle is bad we have bigger problems.

which seems rather odd....

dbr:
Yep shader error and docking crash both fixed for me :partying_face: 

> Window resize events aren't handled properly

Played around with the upstream examples again and they avoid this by just making the (OS) windows not resizable :grin: (i.e maximize etc in the context menu is disabled and so on)

Other thing I noticed: in the rust impl, the docking viewport overlay graphic are a little laggy in the floating window:

imgui-viewport-latency

The upstream examples are synced better, *but* instead the whole window content appears to lag (e.g when moving the widnow quickly, there is a noticable jerkyness to the whole window content)

I suspect there's no a perfect solution to this, in which case the current Rust implementation seems preferable to me

Rob2309:
The new commits should fix the moving/resizing issue. I simply forgot informing ImGui about the position/size changes :)

As for the lagging, I am not sure that I can do anything about that...

naiba:
Does macOS still have a chance to work?

Rob2309:
It should work, yes :)

naiba:
> It should work, yes :)

Yes, it works great on my MacBook!

Rob2309:
That is great to hear 👍

Rob2309:
Is there anything that should be done here?

dbr:
Hmm, this is looking pretty close :partying_face: 

I think the main reminaing things are:

1. I'll have a quick look over the changes again early this week and see if there's anything that stands out
2. Has this been tested on Wayland? My experience has been it can be tricky to get working right (if it works, great; if just doesn't compile, that's fine; but if it's glitchy and broken we should either fix it or make it a compile-time error). I can test this at some point during the week
3. The viewports example uses glow 0.11 and glutin 0.30 but the glow-renderer uses 0.10. I had thought this was blocked by the issue linked in #686 - did you encouter any similar issues with the viewport-renderer?
4. We will need to bump the MSRV to 1.60 for new glutin (which is perfectly fine - just change the `.github/workflows/ci.yml` and the number in the README)
5. Oh it needs rebased atop `main` (conflict looks fairly trivial)

Thanks for all your work on this!

Rob2309:
It seems to be impossible to support wayland with viewports and winit, as controlling the window position is not possible. I could just disable viewports on Wayland if that is a good solution.

I might be able to work on the other points later today...

Rob2309:
> 3\. The viewports example uses glow 0.11 and glutin 0.30 but the glow-renderer uses 0.10. I had thought this was blocked by the issue linked in [Update glow beyond v0.10 #686](https://github.com/imgui-rs/imgui-rs/issues/686) - did you encouter any similar issues with the viewport-renderer?

This actually reminded me that I currently do not backup any GL state. That should probably be done before release.

Rob2309:
> 4\. We will need to bump the MSRV to 1.60 for new glutin (which is perfectly fine - just change the `.github/workflows/ci.yml` and the number in the README)

Actually we would only need to bump the MSRV for the renderer right? so should we stick to a lower one for all the other crates?

Rob2309:
Ok so I plan on doing a few things before I feel comfortable merging this:

- Make the renderer work with viewports disabled
- Automatically disable viewports on wayland (until a better solution is found if there is one)
- Make the renderer properly back up the required GL state
- Check the MSRV, should we bump every crate to 1.60 or just the new renderer that actually requires it (or all crates using glutin)?

dbr:
> Actually we would only need to bump the MSRV for the renderer right? so should we stick to a lower one for all the other crates?

Technically we could, but practically the MSRV is just "version of Rust we check in CI for all crates", I'm happy with having it higher than strictly necessary for all crates

Particularly since our MSRV is basically just a CI thing (i.e it's not set using the `rust-version = "..."` attribute in Cargo which would prevent someone using, say, imgui-sys with otherwise working older Rust version)

---

These changes all look good to me! I pushed a few minor fixes, and once the CI checks finish I think this is good to merge :partying_face:

Rob2309:
We could try to implement the viewport functionality into the existing backends, although that could also be done after merging this in an extra PR...

dbr:
:partying_face: 

> We could try to implement the viewport functionality into the existing backends

Hmm, what do you mean by this - e.g adding a winit+glium viewport example etc, or to integrate the viewport functionality into `imgui-winit-support` more directly so they are composable, or something else?

dbr:
Oh two semi-related things:

- I noticed a problem with imgui-winit-support with how modifier keys were sent - same problem probably applies to the viewport renderer, but I'll copy the fix over in #710
- if you have the time, it would be greatly appreciated if you could look into #686 - I have a reasonable idea of the problem now and could fumble my way to a fix, but since you've effectively fixed the same problem here, it might be a conceptual-copypaste for you :grin:

Rob2309:
> Hmm, what do you mean by this - e.g adding a winit+glium viewport example etc, or to integrate the viewport functionality into `imgui-winit-support` more directly so they are composable, or something else?

I mean making it more composable, there might be a way to do this in more or less beautiful rust.

PR #622, opened 2022-03-12
Fix error in clipboard.rs

BeLeap:
There is no `new` in `ClipboardContext`.
`ClipboardProvider` is not used, and there is `new`, thus `ClipboardProvider` seems to be correct.

### DIFF
--- a/imgui-examples/examples/support/clipboard.rs
+++ b/imgui-examples/examples/support/clipboard.rs
@@ -4,7 +4,7 @@ use imgui::ClipboardBackend;
 pub struct ClipboardSupport(pub ClipboardContext);
 
 pub fn init() -> Option<ClipboardSupport> {
-    ClipboardContext::new().ok().map(ClipboardSupport)
+    ClipboardProvider::new().ok().map(ClipboardSupport)
 }
 
 impl ClipboardBackend for ClipboardSupport {


dbr:
Hmm what this fixing a specific error? The original code seems correct as far as I can tell,

- [`ClipboardContext`](https://github.com/aweinstock314/rust-clipboard/blob/07d080be58a361a5bbdb548fafe9449843d968be/src/lib.rs#L51) is alised to platform specific thing like `pub type ClipboardContext = x11_clipboard::X11ClipboardContext;`
- `X11ClipboardContext` does have a `new` method via the `ClipboardProvider` impl, https://github.com/aweinstock314/rust-clipboard/blob/07d080be58a361a5bbdb548fafe9449843d968be/src/x11_clipboard.rs#L49-L55

dbr:
Ahh I see - thanks for the explaination!

It's not very obvious, and kind of a downside to the design of Rust's trait system, but the `new` method would come from the `impl <S> ClipboardProvider for X11ClipboardContext` mentioned in the docs (in other words, X11ClipboardProvider etc do have a new method, you just need to make sure the `ClipboardProvider` is in-scope)

If you check the full compiler output (e.g running `cargo build`), the full error would explain the solution:

```
error[E0599]: no function or associated item named `new` found for struct `X11ClipboardContext` in the current scope
 --> src/main.rs:5:23
  |
5 |     ClipboardContext::new();
  |                       ^^^ function or associated item not found in `X11ClipboardContext`
  |
  = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
  |
1 | use clipboard::ClipboardProvider;
  |

For more information about this error, try `rustc --explain E0599`.
```

..which is to add `use clipboard::ClipboardProvider;` to the top of your file

I think Rust Analyzer also makes things confusing by just showing the one line of the error - but I think it would show the "perhaps add a `use` for it" part as one of the auto-fix suggestions?

So in [imgui-examples/examples/support/clipboard.rs](https://github.com/imgui-rs/imgui-rs/pull/622/files#diff-8023d381478e3ad234ff913734ea850301d5e7d1ee19c4f8e9863683e03f6e5b) we do actually want to call `ClipboardContext::new()` (calling the method on the `X11ClipboardProvider` object which implements the required trait)

Hope that makes sense!

PR #625, opened 2022-03-15
imgui-winit-support: Build-test documentation example and fix compiler errors

MarijnS95:
This module-level documentation example fails to compile making it unnecessarily more complicated for doc readers to transplant to their own code, whilst it can be easily tested in the CI if a corresponding Winit version is imported.

- imgui-winit-support: Compile doctest on winit 20+
- imgui-winit-support: Fix use of update_delta_time in example doc
- imgui-winit-support: Get `DrawData` from `Context` instead of `Ui`

### DIFF
--- a/imgui-winit-support/src/lib.rs
+++ b/imgui-winit-support/src/lib.rs
@@ -15,8 +15,30 @@
 //!
 //! ## Complete example for winit 0.20+ (without a renderer)
 //!
-//! ```rust,no_run,ignore
-//! # // TODO: Remove ignore when only one winit version is used
+//! ```no_run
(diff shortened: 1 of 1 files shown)


imgui-winit-support/src/lib.rs
```diff
@@ -15,8 +15,30 @@
 //!
 //! ## Complete example for winit 0.20+ (without a renderer)
 //!
-//! ```rust,no_run,ignore
-//! # // TODO: Remove ignore when only one winit version is used
```
dbr, line 19: I admire the optimism in this comment :sweat_smile:

dbr (approved):
That's excellent, thanks!

PR #624, opened 2022-03-15
imgui: Upgrade parking_lot to 0.12 with windows-rs bindings

MarijnS95:
`parking_lot` now uses windows-rs's `windows-sys` crate, officially developed and vetted by Microsoft - the bindings are autogenerated from metadata and faster to compile too.

See also https://github.com/Amanieu/parking_lot/pull/311.

PR #629, opened 2022-03-19
Proposed v0.8.3 [don't merge PR]

dbr:
(note: I've made this new branch, named `patch-v0.8`, which holds all the tags/patch commits from `v0.8.x`. Also this PR just exists for visiblity of the commits - would be a terrible idea to actually merge, thus the draft status)

Compared to the 0.8.2 release, this contains:

- Bugfix to the `.crate` archive so it works with the freetype stuff
- This should fix the docs.rs build
- Updates parking_lot dep to 0.12 as per #624

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,8 @@ jobs:
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-23/default --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-24/default --all-targets
       - run: cargo clippy --manifest-path imgui-winit-support/Cargo.toml --no-default-features --features winit-25/default --all-targets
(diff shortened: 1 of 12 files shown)


dbr:
Oh I breifly thought parking_lot might be in the public API so would arguably be a breaking change,

https://github.com/imgui-rs/imgui-rs/blob/patch-v0.8/imgui/src/test.rs#L6

..but that's only enabled under `#[cfg(test)]` and otherwise only used internally, so I think this is fine :thinking: 

https://github.com/imgui-rs/imgui-rs/blob/patch-v0.8/imgui/src/lib.rs#L84-L85

MarijnS95:
Even though this probably doesn't matter as it "is superseded" by #645, can you update the target branch to https://github.com/imgui-rs/imgui-rs/compare/v0.8.2-patch, which this branch is based on top of (it looks like this was the base branch for a split-second)? That makes the diff much more obvious to read/understand.

(Perhaps that branch should be renamed though, https://github.com/imgui-rs/imgui-rs/pull/645#issuecomment-1152586314)

dbr:
Closing this in favour of releasing v0.9

Issue #531, opened 2021-09-17
Tracking Issue for v0.9


sanbox-irl:
Hello!

I am titling this `v0.8.1`, but I expect this will actually end up describing `v0.9.0` (we love to make breaking changes around here).

Here are my personal two goals for this release (which should happen *much* faster than `v0.8.0`) and one stretch goal.

1. Documentation. Documentation EVERYWHERE.
Eagle-eyed repo watchers may have notied that `#![deny(missing_docs)]` was added to the repo, commented out. I would like to release the next version of this crate with that lint enabled. Currently, we have an even 300 missing pieces of documentation. That's a lot, and to a degree, we need to solve that by...writing docs. However, on a deeper level, we have kinds of code which is almost purpose built to be annoying to document. Many of our builders attempt to hide bitflags with `bit_flag_name(&mut self, bool)` kinds of methods (for example, `collapsible` on the `WindowBuilder` struct). These create tons of functions and bad namespacing -- instead, I wish we just had a `.flags` method that users provided bitflags to.
Additionally, I'd like to have more examples for the most used widgets. We have a lot of that covered already, but I'll make a list soon of those that are lacking.
This might not be done fully in time for release, but I'd like us to make the effort. However, unless you're **sure** that the code you're going to document won't be edited in this release (if it's on `Ui` for example, it's probably okay), then let's hold off on this till the other points are addressed.
2. Make alternatives to the Builder pattern.
This one is a big complex because we have several approaches we can take. First, the problem itself: builders are annoying to use because the builder struct itself needs to be pulled into scope, and it makes it so that many basic widgets (buttons, `input_text`, etc) are on the `Ui` object, but more interesting widgets (like `Window`) are in some builder that you need to find. Ideally, *everything which Ui can build, Ui should have a method to build it*.
I see us approaching this with a few methods:
    1. `Ui` can create the requisite `Builder` with a method, and we maintain the builder syntax. So to create a window, for example, you could do `ui.begin_window("name", REQUIRED_PARAM).build();`. I think this is a fine idea, and involves probably the least code deletion.
    2. `Ui` uses `with_x` style "overloads" to allow for extra parameters to be passed in. This works, and is quite nice for many, perhaps even the majority, of functions, but really has a tough time for something like `Window`, where there are a LOT of things that can be given to `Window`, and the russian doll of `with_x` gets difficult.
[...]

PR #630, opened 2022-03-19
Update README with a more up-to-date code sample

Jasper-Bekkers:


dbr:
Hmm, I think the current example is correct - the general approach (mentioned in [the 0.9 tracking issue](https://github.com/imgui-rs/imgui-rs/issues/531)) is that everything gets built by methods on `imgui::Ui`

So `imgui::Window::new` is deprecated, and `ui.window("Hello World")` is the way : https://github.com/imgui-rs/imgui-rs/blob/3e531edfed48a12c7492d85262f38309893d1116/imgui/src/window/mod.rs#L187-L188

Jasper-Bekkers:
I see - makes sense why it doesn't work with the latest stable release then ;-)

dbr:
Ahhh, right! Yes the readme and examples will be a little confusing until the next release :grimacing: Need to make sure to view them via the previous tag, https://github.com/imgui-rs/imgui-rs/tree/v0.8.2

PR #634, opened 2022-04-06
Use nohash_hasher for the texture map

wildbook:
Contrary to the existing comment, this does not yet allow `Textures::new` to become a `const fn`. `hashbrown` has  `const fn with_hasher`, but it's not yet exposed from `std` and I wanted to avoid adding a dependency on `hashbrown` without at least asking first.

In one of the projects I'm working on I'm not able/willing to use the default hasher, so I've had to be somewhat selective with dependencies. In imgui-rs' case I'd love to see the hasher replaced with `nohash_hasher` or `fxhash` since that'd mean I'm not required to maintain my own fork of it, but I understand if that isn't something that others also care about.

I'm opening this PR mostly so that if you have nothing against merging it I'm not stuck with an unnecessary fork. As a bonus it also makes the transition to `const` one step smaller once std marks `with_hasher` as `const`.

### DIFF
--- a/imgui/Cargo.toml
+++ b/imgui/Cargo.toml
@@ -21,6 +21,7 @@ imgui-sys = { version = "0.8.1-alpha.0", path = "../imgui-sys" }
 mint = "0.5.6"
 parking_lot = "0.12"
 cfg-if = "1"
+nohash-hasher = "0.2.0"
 
 [features]
 wasm = ["imgui-sys/wasm"]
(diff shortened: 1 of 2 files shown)


thomcc (commented):
Given that a hasher that doesn't hash and just stores the integer is very little code, I think adding a dep for it might be overkill?

wildbook:
I opted to add the dependency since the `Hasher` trait requires quite a few functions to be implemented and the `nohash_hasher` crate also explains its limits pretty well.

I also didn't really feel comfortable adding 50+ lines of code to imgui-rs itself when said code was better documented and readily available in a crate with no dependencies.

If you'd rather have it be copied into imgui-rs feel free to do that instead!

PR #635, opened 2022-04-10
TextCallbackData::selection causes problems when selecting from right to left

wcassels:
Currently, when working with callbacks on an `InputText`, it's possible for `TextCallbackData::selection()` to return a range with `start` > `end`, eg. when selecting text by dragging from right to left. One resulting problem is that when you do this, `TextCallbackData::has_selection()` will return false, because the implementation is simply 
```rust 
pub fn has_selection(&self) -> bool {
    !self.selection().is_empty()
}
```
with stdlib's implementation of `Range::is_empty()` being
```rust
pub fn is_empty(&self) -> bool {
    !(self.start < self.end)
}
```
A bigger problem is that calling `TextCallbackData::selected()` with the selection in this state will cause a guaranteed panic, because the selection range is used to index the input buffer.

This MR provides one possible solution by ensuring the `Range` returned by `TextCallbackData::selection()` always satisfies `start` <= `end`.

### DIFF
--- a/imgui/src/input_widget.rs
+++ b/imgui/src/input_widget.rs
@@ -1051,7 +1051,20 @@ impl TextCallbackData {
     /// This Range is given in `usize` so that it might be used in indexing
     /// operations more easily. To quickly grab the selected text, use [selected](Self::selected).
     pub fn selection(&self) -> Range<usize> {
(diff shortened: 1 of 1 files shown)


sanbox-irl:
Nice catch! Thank ya for the work

PR #636, opened 2022-04-13
Corrected "Hello World" example

const-volatile:


const-volatile:
I saw the history of README.markdown just now and noticed that it was already modified earlier due to an API change in the newest version (for which docs.rs build is failing). Current "Hello World" example is correct, Sorry.

Issue #627, opened 2022-03-19
Simplify (or remove) handling of multiple winit versions


dbr:
I find the imgui-winit-support module extremely hard to follow, as there are methods duplicated several times, each with differing `#[cfg(...)]`

I propose we basically have one file per version of winit. E.g we have `winit_25.rs`, `winit_26.rs` etc - and `lib.rs` just brings in the right one based on the same `#[cfg(...)]` checks

The reasoning being:

1. It's makes it easier to diff two files to find what changes between each version (e.g `diff winit_25.rs winit_26.rs`)
2. While it increases the amount of duplication, it's not a lot - a majority of the methods are already duplicated (basically just `WinitPlatform::init`, `WinitPlatform::copy_mouse_to_io` and `WinitPlatform::prepare_render` are not duplicated)
3. It contains the `#[cfg(...)]` to a smaller area, which might make adding new versions simpler and less accident prone, same with dropping support for older versions

@thomcc @sanbox-irl - thoughts?

dbr:
Releatedly: what versions of winit do we still practically need to handle?

I realized I don't actually understand why we have support for so many winit versions - poking around it seems to originate from #228 which added support for 0.20 which was a big breaking change from 0.19.. but the description accompanying it was interesting:

> The main reason for supporting both versions is that gfx-rs is still on 0.19, so we can't just upgrade to 0.20. I've got a hunch we'll need to drop gfx-rs support soon completely because it's not maintained actively...

..and we have long since dropped gfx-rs as the project is abandonded

I wonder if a more tenable policy is:

1. Support the current primary version of winit at time of release
2. Also support the previous primary version of winit from last

So for imgui-rs 0.9 we suport winit-0.25 (default in imgui-rs 0.8) plus winit-0.26 (current release). For 0.10 we would support only 0.26 (assuming no major winit release before then)

That way we provide a smoother upgrade path (user can update imgui, *then* winit - rather than requiring both at once), while keeping the maintenance burdent reasonable (i.e not supporting 2019 release of winit)

However I wonder how necessary even this is - I'm not aware of (m)any crates which support multiple versions of their dependencies as a general policy

dbr:
I made #638 to see how things look if we just delete the old-winit-feature-flags

Given:

1. the fairly minor changes in `imgui-winit-support` (I think there was code differences only for winit-0.19 and 0.20, all subsequent versions were about the same)
[...]

PR #638, opened 2022-04-23
Removing support for old winit versions

dbr:
Experiment to see how things look if we just just support the "current" winit version (current v0.26), as discussed(?) in #627


sanbox-irl:
Yup this is going to make life better for us. We might end up needing to revert but I think we're good to go

PR #644, opened 2022-05-28
glow: move vertex creation to same method as deletion

### LINKED ISSUE #643: ResetRenderState memory leak in glow-renderer
When resetting the renderer, it doesn't clean up the allocated VAO. To fix this, simply delete the allocated VAO when setting up render state. Not sure if this has side effects, you might want to change it to an Option and only delete it when it is Some.

```rust
#[cfg(feature = "bind_vertex_array_support")]
if self.gl_version.bind_vertex_array_support() {
    unsafe {
        gl.delete_vertex_array(self.vertex_array_object);    // add this line
        self.vertex_array_object = gl
            .create_vertex_array()
            .map_err(|err| format!("Error creating vertex array object: {}", err))?;
        gl.bind_vertex_array(Some(self.vertex_array_object));
    }
}

```

https://github.com/imgui-rs/imgui-rs/blob/main/imgui-glow-renderer/src/lib.rs#L402

dbr:
Closes #643

(cc @Selicre @jmaargh )

### DIFF
--- a/imgui-glow-renderer/src/lib.rs
+++ b/imgui-glow-renderer/src/lib.rs
@@ -3,6 +3,11 @@
 //! This is heavily influenced by the
 //! [example from upstream](https://github.com/ocornut/imgui/blob/fe245914114588f272b0924538fdd43f6c127a26/backends/imgui_impl_opengl3.cpp).
 //!
+//! It is important to note this renderer's API is not foolproof. It was designed
+//! more for simplicity and control (allowing the user to manually have some control
+//! over the GL state) than as a production-ready fully-general API. This control is
+//! why so many things are `pub`.
+//!
 //! # Basic usage
 //!
 //! A few code [examples] are provided in the source.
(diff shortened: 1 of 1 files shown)


imgui-glow-renderer/src/lib.rs
```diff
@@ -3,6 +3,11 @@
 //! This is heavily influenced by the
 //! [example from upstream](https://github.com/ocornut/imgui/blob/fe245914114588f272b0924538fdd43f6c127a26/backends/imgui_impl_opengl3.cpp).
 //!
+//! It is important to note this renderer's API is not foolproof. It was designed
+//! more for simplicity and control (allowing the user to manually have some control
+//! over the GL state) than as a production-ready fully-general API. This control is
+//! why so many things are `pub`.
+//!
```
jmaargh, line 10: I'd suggest removing this and instead extending the "Scope" section of the docs to something like this:

```
//! Consider this an example renderer. It is intended to be sufficent for simple
//! applications running imgui-rs as the final rendering step. If your application
//! has more specific needs, it's probably best to write your own renderer, in
//! which case this can be a useful starting point. This renderer is also not
//! foolproof (largely due to the global nature of the OpenGL state). For example,
//! a few "internal" funcitons are marked `pub` to allow the user more
//! fine-grained control at the cost of potential rendering errors.
```

jmaargh (commented):
Added a suggested change to the docs, but otherwise LGTM

thomcc (approved):
This looks good to me. It's weird to create/delete the VAO every frame but it makes sense for debug UI like imgui to avoid relying on too much external state.

jmaargh:
@thomcc As well as simplifying shared-state management, the [upstream example](https://github.com/ocornut/imgui/blob/fe245914114588f272b0924538fdd43f6c127a26/backends/imgui_impl_opengl3.cpp#L384) notes that this allows the renderer to work across multiple OpenGL contexts

PR #646, opened 2022-06-02
Run 'cargo package' in CI

dbr:
Should verify the .crate file sent to crates.io can at least be built, which should catch things like the docs.rs build failure, as discussed in #645 and #631

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,13 +153,13 @@ jobs:
 
       # Test the .crate sent to crates.io can be built
       - name: build .crate package (default)
-        run: cargo package -p imgui-sys && cargo package -p imgui
+        run: cargo package --no-verify -p imgui-sys && cargo package --no-verify -p imgui
       - name: build .crate package (docking)
-        run: cargo package -p imgui-sys --features docking && cargo package -p imgui --features docking
+        run: cargo package --no-verify -p imgui-sys --features docking && cargo package --no-verify -p imgui --features docking
       - name: build .crate package (freetype)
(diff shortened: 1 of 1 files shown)


thomcc:
Would this actually catch the docs.rs failure? I'm surprised. Cargo does `cargo package` as part of publishing, so it doesn't make sense to me that it would, when it (presumably) wasn't caught as part of the publish.

dbr:
Ah, sorry, I should have explained better: it would catch the specific failure that caused #631

The problem was some files needed for the `freetype` weren't included in the .crate package

Then given we [build the docs with various features enabled](https://github.com/imgui-rs/imgui-rs/blob/1de842a402af309adb1332c2b452ba7857fdc5b1/imgui/Cargo.toml#L15-L16), the first thing to notice the problem was the docs.rs build (since that pulls from crates.io, not the git repo)

However the problem was also happen if someone used imgui like `imgui = {version = "0.8.2", features = ["freetype"]}`

> Cargo does cargo package as part of publishing, so it doesn't make sense to me that it would, when it (presumably) wasn't caught as part of the publish

This is correct, but only if there are no features :cry: (i.e publishing only checks `cargo build`, not `cargo build --features $x`)

dbr:
It is not easy to do `cargo package -p imgui` without publishing the dependant `imgui-sys` to crates.io (adding the `--no-verify` flag defeats the purpose of this since it skips building the crate)

There's an existing issue at on https://github.com/rust-lang/cargo/issues/9260 which I'll add a note to

There's probably workarounds (like setting up a temporary private package registry or something), but not worthwhile given the complexity versus the minor severity of the problem it might catch

PR #647, opened 2022-06-07
Removing note in README about being under new management.

robertkeizer:
Noticed that a line in the README stating that the project was under new maintenance had been there a while.

Checked the git blame showed it had been there for ~1.5 years, figured a PR to change it wouldn't be out of line.


dbr:
Good point - thanks!

Issue #649, opened 2022-06-13
Finish TextFilter


tetenpapier:
I would like to use the TextFilter on a list but I don't find anything about it in the documentation.

The binding is already written (#313 & [in code](https://github.com/imgui-rs/imgui-rs/blob/f523dd2501daa683f311359cae8eed5444e0b594/imgui-sys/src/bindings.rs#L1202)), but not the API.

dbr:
Are you referring to this https://github.com/ocornut/imgui/blob/v1.86/imgui_demo.cpp#L1357-L1376 ?

If so I think they are available in API:

1. The flags like `ImGuiInputTextFlags_CharsUppercase` are available as `InputText::chars_uppercase` https://docs.rs/imgui/0.8.0/imgui/struct.InputText.html#method.chars_uppercase etc
2. From #516, the callback based filters are usable via https://docs.rs/imgui/0.8.0/imgui/struct.InputText.html#method.callback and the trait mentioned there. I suspect it might be quite tricky to implement purely based on the docs, so there is an example which is probably clearer here: https://github.com/imgui-rs/imgui-rs/blob/v0.8.2/imgui-examples/examples/text_callbacks.rs

teteenpapier:
No, I'm referring to https://github.com/ocornut/imgui/blob/v1.86/imgui_demo.cpp#L5642-L5660.

I've searched `Filter` in the Rust documentation but only found the filters for `InputText`. I can make my own filter but I prefer to use the native features of ImGUI.

dbr:
Ahh, I see! That is indeed not yet exposed in the API yet. I don't have time to look into this currently, but a PR would be welcomed - it should be relatively simple to make the bindings (although anything dealing with strings can be a bit tricky because of the Rust String vs C++'s null-terminated strings)

dbr:
Closing - was done in #658 and will be released with v0.9 :partying_face:

PR #658, opened 2022-08-10
add TextFilter

tetenpapier:
I opened the issue #649 to add TextFilter. I did it but I think it could be better. I'ven't found an example in the library so I tried something. I also added it in the `test_window_impl` to test.

### DIFF
--- a/imgui-examples/examples/test_window_impl.rs
+++ b/imgui-examples/examples/test_window_impl.rs
@@ -745,6 +745,12 @@ CTRL+click on individual component to input value.\n",
 
             state.filter.draw();
             let lines = vec!["aaa1.c", "bbb1.c", "ccc1.c", "aaa2.cpp", "bbb2.cpp", "ccc2.cpp", "abc.h", "hello, world!"];
+
+            ui.same_line();
+            if ui.button("Clear##clear_filter") {
+                state.filter.clear();
+            }
+
             for i in lines.iter() {
                 if state.filter.pass_filter(String::from(*i)) {
                     ui.bullet_text(i);
--- a/imgui-examples/examples/test_window_impl.rs
+++ b/imgui-examples/examples/test_window_impl.rs
@@ -752,7 +752,7 @@ CTRL+click on individual component to input value.\n",
             }
 
             for i in lines.iter() {
-                if state.filter.pass_filter(String::from(*i)) {
+                if state.filter.pass_filter(i) {
                     ui.bullet_text(i);
                 }
             }
--- a/imgui-examples/examples/test_window_impl.rs
+++ b/imgui-examples/examples/test_window_impl.rs
@@ -53,6 +53,7 @@ struct State {
     stacked_modals_item: usize,
     stacked_modals_color: [f32; 4],
     app_log: Vec<String>,
+    filter: imgui::TextFilter,
 
     tabs: TabState,
 }
@@ -117,6 +118,7 @@ impl Default for State {
             stacked_modals_item: 0,
             stacked_modals_color: [0.4, 0.7, 0.0, 0.5],
             app_log: Vec::new(),
+            filter: TextFilter::new(String::from("Test")),
             tabs: TabState::default(),
         }
     }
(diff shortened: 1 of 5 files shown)


tetenpapier:
I'm not sure about the path of the source file. I don't think it's a widget but idk where to put it.

dbr:
Thanks for this! :partying_face:

> I'm not sure about the path of the source file. I don't think it's a widget but idk where to put it.

I think probably just `imgui/src/text_filter.rs` seems best - it's similar to things like `list_clipper.rs`

imgui/src/widget/text_filter.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// Returns true if the filter is not empty (`""`).
+    pub fn is_active(&self) -> bool {
+        unsafe { sys::ImGuiTextFilter_IsActive(self.raw) }
+    }
+
+    /// Returns true if the buffer matches the filter.
+    ///
+    /// [`draw()`](Self::draw) or [`build()`](Self::build) mut be called **before** this function.
+    pub fn pass_filter(&self, mut buf: String) -> bool {
```
dbr, line 66: Might be worth having this take `&str` instead, as otherwise it forces on the user to do something like

```rust
State {
   options: Vec<String>,
   filter: TextFilter::new(...),
}

fn render(ui: &imgui::Ui) {
    for x in &self.options {
        // currently would need:
        filter.pass_filter(x.clone());
    }
}
```

By taking `&str` I think it better describes what the argument actually is (a read-only string used to check if it matches)

It would mean the `pass_filter` method internally allocates a new String each time, but, we can probably optimize this away with the `imgui::string::UiBuffer` later (without changing the API)

imgui/src/widget/text_filter.rs
```diff
@@ -0,0 +1,104 @@
...
+        buf.push('\0');
+        let ptr = buf.as_mut_ptr();
+        unsafe {
+            sys::ImGuiTextFilter_PassFilter(self.raw, ptr as *mut sys::cty::c_char, ptr::null())
+        }
+    }
+
+    pub fn pass_filter_with_end(&self, mut start: String, mut end: String) -> bool {
```
dbr, line 74: What is this method used for? It's not immediatly clear (..even looking at imgui.cpp!)

imgui/src/widget/text_filter.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /// Returns true if the filter is not empty (`""`).
+    pub fn is_active(&self) -> bool {
+        unsafe { sys::ImGuiTextFilter_IsActive(self.raw) }
+    }
+
+    /// Returns true if the buffer matches the filter.
+    ///
+    /// [`draw()`](Self::draw) or [`build()`](Self::build) mut be called **before** this function.
```
dbr, line 65: What happens if `pass_filter` is called without build? I guess imgui aborts?

If so, I wonder if it'd be worth adding some extra object so you do

```rust
fn draw(&'a mut self) -> &'a TextFilterHandle {
    ...
}
fn build(&'a mut self) -> &'a TextFilterHandle {
```

And since the `pass_filter` methods etc would be on that `TextFilterHandle` struct, it would be impossible to misuse (since you can't get at the `pass_filter` method without calling one of those methods)

Not a vital thing - there are plenty of ways to make imgui abort via the bindings

dbr:
The CI failure in the Rust 1.54 checks is unrelated to PR

I think the `x11-dl v2.20.0` (used by `winit`) now uses edition 2021 which isn't supported by Rust 1.54. We can probably just bump the MSRV to 1.56

imgui/src/widget/text_filter.rs
tetenpapier: I don't exactly know too.. I have to try it but I don't have time lately.

imgui/src/widget/text_filter.rs
tetenpapier: Oh you're right. I just must create a String to add the `\0` at the end. I'll check `imgui::string::UiBuffer` too

imgui-examples/examples/test_window_impl.rs
```diff
@@ -117,6 +118,7 @@ impl Default for State {
             stacked_modals_item: 0,
             stacked_modals_color: [0.4, 0.7, 0.0, 0.5],
             app_log: Vec::new(),
+            filter: TextFilter::new(String::from("Test")),
```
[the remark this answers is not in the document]
tetenpapier: I've checked other implementations of the bindings without ever seeing anything identical.
When I read the provided methods, I thought it would be better to have a pointer generated once for ImGuiTextFilter (here `raw`, generated by `imgui_sys::ImGuiTextFilter_ImGuiTextFilter`).

However, your idea is more in the thought of the crate but I don't know if the created pointer will be destroyed at each frame ?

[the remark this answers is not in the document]
dbr: Marking as resolved - your current implementation is consistent with the C++ API, and there's no obvious reason to change it!

dbr:
Thanks! This seems good to me, merging

There's some room for optimizing away the internally allocated `String` but this can be done later if someone finds it necessary!

PR #659, opened 2022-08-17
Support winit v0.27.2

dbarkar:


dbr:
Hmm, it's odd the CI checks don't seem to have run (there should be an approval step but.. isn't appearing here)

However the diff seems good - thanks, shall merge!

PR #662, opened 2022-08-24
leave tree_node label None if not set

linw789:
This allows users to hide tree_node label. This change also added an
example of making tree_node as a button.


linw789:
Does anyone know how to remove highlight when mouse hovering over a tree node?

dbr:
Looks good, thanks!

> Does anyone know how to NOT highlight when mouse hovering over a tree node?

Can you just override the `HeaderHovered` style color? (e.g set the alpha to zero)

Otherwise, I'm not too sure - might be worth asking in the upstream repo

Issue #665, opened 2022-08-29
Decide on and document dev/release process


dbr:
As per #663

We currently don't have a documented approach to how we perform releases, and doing things like backporting changes to old releases

Key points would be:

1. Stable default branch
2. Develop branch with latest unreleased changes
3. Release branches
4. Checks needed before making release
5. Steps to actual publish imgui-rs (e.g release imgui-sys, then imgui; when to bump version numbers, etc)

dbr:
Thanks, that sounds good to me!

I'll make a PR shortly with a document writing that down. If everyone is happy with that doc then we can move on to making a 0.9 release and onwards :partying_face:

dbr:
PR with propsed doc at #672

parasyte:
FWIW, the process described in the document in the new PR is a good development strategy that I have used on several complex projects. Things do get a bit ugly when the stable branches go on for some time as `main` diverges more and more. It makes "just cherry-pick" infeasible after a while. It's great when that's all it takes, though!

MarijnS95:
@parasyte Indeed, I've been running into a similar scenario with a low-level bindings crate where we're cutting down heavily on quick/successive breaking releases. As a consequence I have to pick about 98% of the changes - which are non-breaking fixes or additions only - to stable branches, to the point where I'm considering to target/merge every PR to the stable branch instead, and manually rebase the three-or-so remaining breaking-change commits (via `main`) on top of the stable branch every once in a while.

Alas, it depends on the project and how many breaking changes versus additive development and bugfixes it sees.

airways:
For what it's worth, I like it, big improvement over the current situation!

PR #672, opened 2022-10-05
Proposed branching/release process

dbr:
As per my intepretation of @MarijnS95 's comment https://github.com/imgui-rs/imgui-rs/issues/665#issuecomment-1257758263

Goal being simple simple and easy to stick with. I'd like to expand the doc with exact step-by-step commands to make things trivial to follow along in future, but best to do that as part of releasing v0.9.0

@thomcc @sanbox-irl - any thoughts/opinions/desires-to-switch-back-to-SVN? :stuck_out_tongue:

### DIFF
--- a/docs/development-process.md
+++ b/docs/development-process.md
@@ -1,0 +1,32 @@
+# imgui-rs development process
+
+As summarised in [issue #665](https://github.com/imgui-rs/imgui-rs/issues/665)
+
+In summary:
+
+1. There is a `main` branch
+2. For each semver compatible release there is a stable branch (e.g `0.9-stable`)
+3. Each patch release becomes a tagged commit on this stable branch (e.g `v0.9.5` would come from a tagged commit on the `0.9-stable` branch)
+
+## General process
+
+Day to day development
+
+1. Work on `main` branch
+2. General PR's are submitted against the `main` branch
+
+When it is time to make a new release, we create a `x.y-stable` branch (e.g `0.9-stable`) from `main`
+
+1. Ensure `CHANGELOG` is up to date
+1. Bump `version` in the various `Cargo.toml`
+2. A stable branch is created, e.g `git switch -c 0.9-stable` and pushed to Github
+
+All further PR's are still done to `main`
+
+1. If they are applicable to previous release (e.g bugfixes or non-breaking changes), they are cherry-picked back to the applicable `stable` branch(es)
+
+## When to bump versions in Cargo.toml
+
+Only before publishing to crates.io.
+
+This makes users able use [patch.crates-io] without hand-editing versions throughout their dependency tree (typically impossible without forking/editing transitive dependencies, even if there are no breaking code changes otherwise).


docs/development-process.md
```diff
@@ -0,0 +1,32 @@
...
+
+When it is time to make a new release, we create a `x.y-stable` branch (e.g `0.9-stable`) from `main`
+
+1. Ensure `CHANGELOG` is up to date
+1. Bump `version` in the various `Cargo.toml`
+2. A stable branch is created, e.g `git switch -c 0.9-stable` and pushed to Github
+
+All further PR's are still done to `main`
+
+1. If they are applicable to previous release (e.g bugfixes or non-breaking changes), they are cherry-picked back to the applicable `stable` branch(es)
```
MarijnS95, line 26: How do you want this to happen? Should this be discussed/described in a PR beforehand? Then, should a maintainer perform the `cherry-pick` as soon as a PR is merged so that it is not forgotten about?

Or should this be a more dynamic process, where a maintainer (in an issue/discussion) at some point decides pick/backport a list of merged PRs and publish a patch release?

It depends on the kind and volume of changes/contributions a repository receives, keeping up with `cherry-pick`s might turn out to be a lot of work but also prevents the stable branch from lagging behind and turning into a gigantic pile of work when it comes to making a patch release.

---

Perhaps a more rigorous maintenance plan needs to be put in place (how often to allow breaking releases, how long to support how many older stable versions, whether to at least publish a patch release with accumulated fixes/additions right before doing the next breaking release, etc.), but that can also form over time pending how this release process pans out (again: depends on how many breaking changes vs purely additions/bugfixes this repo receives).

MarijnS95 (approved):
Looks good for a release process! This might need to simmer for a while to come up with a good _maintenance_ process though, coordinating when and how many breaking vs patch releases.

MarijnS95:
@dbr Where are we at on this? I'd like to get access to a new `imgui` release through crates.io :)

MarijnS95:
@dbr @thomcc @sanbox-irl another friendly nudge :wink:

MarijnS95:
@dbr @thomcc @sanbox-irl once again? What's the hold up for fleshing out the release process and finally bringing this update out to users?

docs/development-process.md
dbr: These are good points, but I suspect we should come back to this later in the interests of progress

For now I think the PR's would just go to the development branch, and if required be backported to stable by whoever merges them (e.g click merge, decide if it needs backported)

Once we are back to a less "abandonded-looking" development cycle, we should definitely discuss this further!

dbr:
Thanks for the gentle poking @MarijnS95 - sorry for the absense, no specific reason for the delay, just my attention has been dragged to work and house things!

Shall merge this as-is - shall tweak it when I actually go through the release process, and as mentioned in comment above we should definitely revisit some aspects of this later

I'll aim to get a release out in teh next few days - the only blocking thing remaining is.. I don't have access to the crates.io package (never noticed when I was added as a maintainer here, as sandbox has performed the last few releases!). I'll email him about this now

PR #673, opened 2022-10-05
Restoring "Include imgui/misc/freetype/ in crate package" from 0.8.x release

dbr:
In trying to straighten things out for #665 I noticed one of the bugfixes release in `0.8.something` wasn't in the main branch - specifically the change for #595

This reapplies this fix to main branch, allowing the "--features freetype" to work when using via crates.io

### DIFF
--- a/imgui-sys/Cargo.toml
+++ b/imgui-sys/Cargo.toml
@@ -10,8 +10,20 @@ license = "MIT/Apache-2.0"
 categories = ["gui", "external-ffi-bindings"]
 build = "build.rs"
 links = "imgui"
-# exclude .json, .lua from imgui dirs - they are intermediate artifacts from cimgui generator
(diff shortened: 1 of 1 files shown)


dbr:
Continuing the "everything ended up in a bit of a mess" story of 0.8's patch releases, this was fixed slightly differently in `main` in 9128066f01a36803c04609f288fc04dd2c04641b

However I think this PR is still worth merging as the glob patterns in 9128066f01a36803c04609f288fc04dd2c04641b result in an unnecessarily huge `.crate` (as it includes copies of all the imgui C++ example backends and docs etc), whereas the this PR only includes the necessary files

thomcc:
In general we do "fork off and do a patch release" we should be sure to merge the changes back to main in some fashion after that's done.

PR #676, opened 2022-10-27
Re-export `winit` and `glow` crates

lovelymono:
This makes it easier for users to use the correct version of these crates.

Signed-off-by: Lena Milizé <me@lvmn.org>


MarijnS95:
Seems like an okay change, but I rather prefer if users themselves determine their `winit`/`glow` version (based on more needs than just the `imgui` crate), and select an appropriate `imgui` version to match. This wouldn't be such an issue if `imgui` just released more often and more consistently, it's currently holding up the `winit` bump around the ecosystem for everyone using `imgui-winit-support` (see open issues and PRs about releases and the process around it).

dbr:
Agreed - I think this is worth merging as it'll make the simple case easier ("just want imgui to work"), and doesn't preclude more careful selection of versions where needed

Also certainly agreed about releasing more often :sweat_smile: Working on that now

PR #685, opened 2022-12-02
Update to imgui 1.89

dbr:
More straight forward than I had expected as the major changes are all backwards compatible

Few things to do:

1. [ ] Fix up docking feature
2. [ ] The 1.89.1 tag still had `IMGUI_VERSION` containing `1.89.1 WIP` - update in the next commit. Should probably use this instead of the exact tagged release to avoid the version string being confusing
3. [ ] Consider backwards-compatbility for the `ImageButton` change
4. [ ] Investigate what changes need to take advantage of the event-based input stuff (probably a separate PR)
5. [ ] Check over changelog for versions since previous 1.86 to see if there's any other significant changes for the bindings


dbr:
Superseded by #688 :partying_face:

Issue #689, opened 2022-12-25
Add impl `DataTypeKind` for size types (`usize`/`isize`)


Ararem:
# Description
When using a slider, you need to pass in a value that implements `DataTypeKind`. Currently, this is only the primitive numeric types, and I propose to extend it to include `usize` and `isize` as well.

# Proposal
We add an 
```rust
unsafe impl DataTypeKind for usize {
    const KIND: DataType = <PLATFORM DEPENDENT>;
}
```
We may need to add some internal casting to whichever integer type it matches. I assume this could be done by conditional compilation, or `size_of`: `const USIZE_WIDTH:usize = size_of::<usize>();`

# Alternatives
My current implementation is to cast to `u64` everywhere:
```rust
        // usize can't be used in a slider, so we have to cast to u64, use that, then case back
        type SliderType = u64; // Might fail on 128-bit systems where usize > u64, but eh
        let mut num_track_frames_compat: SliderType = *track_frames as SliderType;
        ui.slider_config(
            "Num Tracked Frames",
            1,
            Self::HARD_LIMIT_MAX_FRAMES_TO_TRACK as SliderType,
        )
        .flags(SliderFlags::LOGARITHMIC)
        .build(&mut num_track_frames_compat);
        *track_frames = num_track_frames_compat as usize;
        /*...*/

        *displayed_frames = min(*displayed_frames, *track_frames); // Don't allow it to go over num_track_frames
        let mut num_displayed_frames_compat: SliderType = *displayed_frames as SliderType; // Might fail on 128-bit systems, but eh
        ui.slider_config(
            "Num Displayed Frames",
            1,
            min(Self::HARD_LIMIT_MAX_FRAMES_TO_TRACK, *track_frames) as SliderType,
        )
        .flags(SliderFlags::LOGARITHMIC)
        .build(&mut num_displayed_frames_compat);
        *displayed_frames = num_displayed_frames_compat as usize;
        /*...*/
```

# Safety & Reliability
As far as I'm aware, usize changes depending on which platform it's being targeted at. On 32-bit systems, it's equivalent to u32, on 64-bit u64, etc. (Untested) solution:
Modify `src/internals.rs` to include these lines:
```rust

unsafe impl DataTypeKind for usize {
    #[cfg(target_pointer_width = "8")]
    const KIND: DataType = DataType::U8;

    #[cfg(target_pointer_width = "16")]
    const KIND: DataType = DataType::U16;

[...]

PR #693, opened 2023-01-04
Implement DataTypeKind for `usize` and `isize`

Ararem:
This allows direct use of the size types in `imgui-rs` functions, like `Slider`s. Previously, they had to be cast, passed in, then cast back. This change allows for direct use (no casts necessary)


Ararem:
Fixes #689

dbr:
LGTM, thanks!

Ararem:
@dbr Thanks for merging, do you know when this will be available on crates.io?

dbr:
Oh sorry for the lack of update - this is now released in v0.10

PR #701, opened 2023-01-14
Fix keyboard modifiers in Winit with the new ImGui version.

rodrigorc:
The recent update to DearImGui 1.89.2 was great, but I think it broke the handling of modifier keys.

The thing is that the values of `io.key_shift` and friends are updated in [`ImGui::UpdateKeyboardInputs()`](https://github.com/imgui-rs/imgui-rs/blob/main/imgui-sys/third-party/imgui-master/imgui/imgui.cpp#L8293) based on the values of the _pseudo keys_ such as `ImGuiMod_Ctrl`.

It looks like the [recommended way](https://github.com/imgui-rs/imgui-rs/blob/main/imgui-sys/third-party/imgui-master/imgui/imgui.cpp#L8289) to notify changes in the modifier keys is to just call `AddKeyEvent()`, as if it were a real key.

### DIFF
--- a/imgui-sdl2-support/src/lib.rs
+++ b/imgui-sdl2-support/src/lib.rs
@@ -143,10 +143,10 @@ fn handle_key(io: &mut Io, key: &Scancode, pressed: bool) {
 
 /// Handle changes in the key modifier states.
 fn handle_key_modifier(io: &mut Io, keymod: &Mod) {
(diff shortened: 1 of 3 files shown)


dbr:
Huh, quite right - no idea how I didn't spot this, I was sure I tested modifier keys were working! :shrug:

Thanks for this! Pushed a minor fix, shall merge when CI is happy!

PR #702, opened 2023-01-15
Add callbacks to the draw_list.

rodrigorc:
This PR adds the possibility to add callbacks to the draw_lists. Those were handled by the glow/glium backends, but we would have to add them using unsafe code and raw C callbacks.
```
    let draws = ui.get_window_draw_list();
    draws.add_callback(|| { do_something(); }).build();
```
Currently the callback must implement `FnOnce() + 'static`, that is the safest bet and enough for my needs.

Once the callback is run, I'm setting it to NULL, even though it is a `const` C pointer to avoid the risk of calling it twice, although I think that is not really needed, because an imgui draw-list can only be called once. But just in case...

Also, if the callback is added to the draw-list but then the draw-list is never run, the callback will leak, as DearImGui has no provision of a unused draw-list cleanup.


imgui/src/draw_list.rs
```diff
@@ -1186,3 +1191,46 @@ impl<'ui> ImageRounded<'ui> {
...
+
+impl<'ui, F: FnOnce() + 'static> Callback<'ui, F> {
+    /// Typically constructed by [`DrawListMut::add_callback`]
+    pub fn new(draw_list: &'ui DrawListMut<'_>, callback: F) -> Self {
+        Callback {
+            draw_list,
+            callback,
+        }
+    }
+    /// Call the callback when the window is drawn
```
dbr, line 1209: Minor thing but could possibly reword this slightly, maybe something like

> Adds the callback to the draw-list so it will be run when the window is drawn

Mainly the "Call the callback" part in original wording "half-implies" that calling `build` runs the callback

imgui/src/draw_list.rs
```diff
@@ -1186,3 +1191,46 @@ impl<'ui> ImageRounded<'ui> {
...
+    pub fn build(self) {
+        use std::os::raw::c_void;
+        // F is Sized, so *mut F must be a thin pointer.
+        let callback: *mut F = Box::into_raw(Box::new(self.callback));
+
+        // Note that if the callback is never run the box will leak.
```
dbr, line 1215: Seems like this should be mentioned in the main docstring as it's potentially significant?

dbr:
This looks good, thanks for the PR!

Added a couple of minor notes on the doc strings - I also wonder, would it be much work to add a simple example to `imgui-examples/` showing how this might be used in practice?

imgui/src/draw_list.rs
rodrigorc: Yes, that is right, fixed.

rodrigorc:
> would it be much work to add a simple example to `imgui-examples/` showing how this might be used in practice?

Ah, examples are tricky because they are usually used to do random OpenGL stuff that happens to be quite specific. Indeed, even DearImGui does not provide any callback example, that I know of, it just comments about using it for a render-to-texture pass and so on...

Anyway, I'm adding a commit with an example to the imgui-glow-renderer subdirectory that uses the callback to render a FramebufferObject, not so different from what I'm doing in my real project. I had to wrap the `AutoRenderer::gl` in an `Rc` because of the `'static` requirements of the callback. I cannot wrap the full `AutoRender` into an because that is mutably borrowed when calling `AutoRenderer::render()`.

Feel free to suggest any other way to change that and still get access to the GL context from the callback.

On a footnote, I'm quite sure that it is technically possible to get rid of that pesky `'static` in the callback. But that would probably require to wrap all the UI building in a scope similar to `std::thread::scope`, and it is probably not worth it.

rodrigorc:
Hey, @dbr could you check my latest commits to this PR?

dbr:
Oh sorry for missing this - I think the change to the renderer needed are fine

I fixed up conflict, shall merge when CI is happy (I think it might fail due to unrelated changes of MSRV from an upstream library - shall fix that first in `main` if so)

rodrigorc:
With the merge of the new `glow` version the new example in the PR broke. I've just pushed a fix.

@dbr: After the latest merges the git-log of this PR is a bit messy. I don't know the policy of this project, but if you like I can do a rebase/squash, before an eventual final merge?

dbr:
Ahh good, was worried I had botched something in the merge.

Yep a rebase sounds like a good idea!

rodrigorc:
Ok, I have just rebased this. I've split the `Rc<glow::Context>` in `AutoRenderer` change from the example because I happened to need that in a project of mine, so I can just cherry-pick it from this PR :cherries:.

dbr:
Perfect, thanks!

PR #708, opened 2023-01-21
Added experimental semver-check pipeline

### LINKED ISSUE #694: Setup cargo-semver-checks or similar in CI
Something like this could be used to flag PR's have breaking changes in them

https://github.com/obi1kenobi/cargo-semver-checks

Would mean the CI would never be green on intentionally breaking changes, but practically that is fine

Rob2309:
This PR is intended to resolve #694

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,23 @@ on:
     - cron: "0 0 * * 1"
 
 jobs:
+  semver-checks:
+    name: Check for Semver breaking changes
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        crate: [ imgui, imgui-sys ]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Check ${{ matrix.crate }}
+        uses: obi1kenobi/cargo-semver-checks-action@v1
+        with:
+          crate-name: ${{ matrix.crate }}
+
   rustfmt:
     name: Check rustfmt
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
       - name: Check ${{ matrix.crate }}
-        uses: obi1kenobi/cargo-semver-checks-action@v1
+        uses: obi1kenobi/cargo-semver-checks-action@v2
         with:
           crate-name: ${{ matrix.crate }}
 


dbr:
Excellent, thanks!

I think we should probably check ~all the crates (the renderers etc)

~Only other thing is we should cache (or at least reuse) the installation of `cargo-semver-checks` as the build takes a while and repeats for each of the `matrix.crate`~ - looks like this is in-progress with upstream https://github.com/obi1kenobi/cargo-semver-checks-action/pull/21 - so probably just ignore the rebuilding for now, and just update when that PR is released

Rob2309:
I noticed that the semver check currently checks against v0.9.0, which probably should not be the case right? That means we would need to bump the Cargo.toml version to the next version right after releasing one. Maybe this can be changed though

obi1kenobi:
I think this might be a bug in the v1 GitHub Action, which was written before `cargo-semver-checks` could by itself look up the correct baseline version to use based on what's on crates.io. The v1 GitHub Action's logic has some known issues, and we're excited to be removing it completely in the v2 of the action in the coming couple of weeks.

You shouldn't have to bump versions right after publishing. The intended behavior (tested in `cargo-semver-checks`, not the v1 Action) is:
- If the `Cargo.toml`'s version is not yet on crates.io, then that's assumed to be the about-to-be-published version. It is compared against the highest-numbered non-yanked non-prerelease version on crates.io that is less than the `Cargo.toml` version.
- If the `Cargo.toml`'s version is already on crates.io, then the local code is assumed to be a patch version on top of what `Cargo.toml` has. It is then compared against the `Cargo.toml` version as published on crates.io.

To get this behavior today, you could replace the GitHub Action with just `cargo install --locked cargo-semver-checks && cargo semver-checks check-release <any-flags-you-need>`. Then once the v2 Action is ready for prime-time, I'd recommend switching back to the action to get a speed boost thanks to things like prebuilt binaries and (future feature) baseline rustdoc caching.

Thanks for trying `cargo-semver-checks`, and please don't be shy about opening issues about points of friction you run into, even if minor!

Rob2309:
That sounds very promising :) Since there is no big pressure to introduce this, we might just want to wait for the action to be released.
Since you mentioned opening issues, should we still open an issue for the v1 action?

obi1kenobi:
> Since you mentioned opening issues, should we still open an issue for the v1 action?

No need — I just checked the latest `cargo-semver-checks` version and it correctly picks the published 0.10.0 version as a baseline, so the problem is guaranteed to go away in the v2 action. I think the v1 action probably just has a `<` instead of a `<=` somewhere :)

dbr:
Excellent :partying_face: 

> Since there is no big pressure to introduce this, we might just want to wait for the action to be released

Yep agreed, lets await the V2 action

obi1kenobi:
Oof, sorry about that! I'll check in with the person working on the action and see if we can get it to a release-able state soon. I definitely didn't intend to leave it in its current non-working state for this long... 😅

obi1kenobi:
We just released v2 of the action as well as a [new massively-faster cargo-semver-checks](https://twitter.com/PredragGruevski/status/1650493414197059587) version — [up to 2354x faster](https://twitter.com/PredragGruevski/status/1649473833177325581) measured on a large crate.
https://github.com/obi1kenobi/cargo-semver-checks-action/releases/tag/v2.0

Full docs here:
https://github.com/obi1kenobi/cargo-semver-checks-action#cargo-semver-checks-action

The `crate-name` argument is now called `package` (matching what `cargo` calls it) and accepts a comma-separated list as well, if you'd like to check both crates in one job. With the new optimizations, the "build rustdoc" portion of the runtime dominates the actual checking, so running both crates in one job might be more efficient if they depend on each other or have significant dependencies in common.

If you run into any issues migrating onto the new action, please give me a ping and I'd be happy to help!

sanbox-irl:
@Rob2309 do you have the bandwidth to update this fella with the new changes?

sanbox-irl:
I'm going to close this PR -- I suspect that with our change to the mono-repo this is far less of an issue, but on a bigger level, it's looking we don't have the bandwidth for this

PR #711, opened 2023-02-02
Update to glow 0.12

### LINKED ISSUE #686: Update glow beyond v0.10
Mostly to note this somewhere easier to find:

Currently we are using glow 0.10. To update to 0.11 is not straight-forward, explained nicely in the original glow-renderer PR:

https://github.com/imgui-rs/imgui-rs/pull/495#issuecomment-913056533

Don't fully understand this problem, but from the comments [on the linked upstream issue](https://github.com/grovesNL/glow/issues/187) it seems possible to workaround this with some `std::mem::transmute` calls, and a more robust option may end up in 0.12

Rob2309:
This PR intends to resolve #686.

Instead of using the `std::mem::transmute` workaround mentioned in several issues (e.g. https://github.com/grovesNL/glow/issues/187#issuecomment-962403142), I resorted to using a typesafe solution until a proper mechanism is put into place.

The way I recreate the GL objects from the raw `u32` handles is to directly create new `glow::Native*` objects, since their inner field is public. This at least ensures that any breaking layout changes will be detected, which is not necessarily the case in the `transmute()` workaround.

Of course, this makes compiling for wasm completely impossible right now, but as far as I can tell, this is not possible anyway.

### DIFF
--- a/imgui-glow-renderer/Cargo.toml
+++ b/imgui-glow-renderer/Cargo.toml
@@ -10,7 +10,7 @@ categories = ["gui", "rendering"]
 
 [dependencies]
 imgui = { version = "0.10.0", path = "../imgui" }
-glow = "0.10.0"
+glow = "0.12.0"
 memoffset = "0.6.4"
 
 [dev-dependencies]
--- a/imgui-glow-renderer/examples/utils/mod.rs
+++ b/imgui-glow-renderer/examples/utils/mod.rs
@@ -104,8 +104,8 @@ void main() {
 "#;
 
         let mut shaders = [
-            (glow::VERTEX_SHADER, VERTEX_SHADER_SOURCE, 0),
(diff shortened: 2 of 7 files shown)


Rob2309:
I just dug into the wasm types a bit and it should probably be possible to call get_parameter() instead of get_parameter_i32(). In WebGL get_parameter returns the corresponding WebGl* objects which can by cast via dyn_into(). While requiring special code only for wasm, this would probably work.

I could test this but I don't know yet how to compile an imgui app for wasm or whether its really possible atm.

Rob2309:
Update: with the current glow it actually seems impossible to use the values returned from get_parameter(...) as they return the actual GL objects but the glow bind_* calls expect a key into a map of GL objects.

dbr:
Great, thanks for this!

PR #712, opened 2023-02-06
Update glow to 0.12 and winit to 0.28.

rodrigorc:
Glow 0.11 could not be used because GL resources changed from plain integers into new-types with a private inner value:
```
- type Texture = c_uint;
+ pub struct Texture(NonZeroU32);
```
That breaks many usages from `imgui-rs`. But in Glow 0.12 the inner value is `pub` so all is ok again:
```
+ pub struct Texture(pub NonZeroU32);
```

And since I'm into it, I'm updating `winit`, that has just been updated. Feel free to dissmiss this change or request it to be a separated PR (sorry in advance, I don't know the policy about dependency udpates).

### DIFF
--- a/imgui-glow-renderer/Cargo.toml
+++ b/imgui-glow-renderer/Cargo.toml
@@ -10,7 +10,7 @@ categories = ["gui", "rendering"]
 
 [dependencies]
 imgui = { version = "0.10.0", path = "../imgui" }
-glow = "0.10.0"
+glow = "0.12.0"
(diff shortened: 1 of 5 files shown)


Rob2309:
It seems like this would be a duplicate of https://github.com/imgui-rs/imgui-rs/pull/711.

dbr:
Yep I think this is covered by #711, and there is some complication currently with the winit update described in #716

Regardless, thank you for the PR! :grin:

PR #714, opened 2023-03-03
Backport #710 (modifier key fix) to v0.10

dbr:


sanbox-irl:
nomating for close for this since we're well past v0.10

Issue #687, opened 2022-12-14
Add a way to store rendered "frames"


vityafx:
**Is your feature request related to a problem? Please describe.**
I want to support both, opengl and vulkan renderers of imgui. However, the opengl renderer is relying on having an object with a limited lifetime passed to the "render" function. At the same time, for Vulkan it is better to separate the "drawing" and "presentation" stages, due to the nature of vulkan. So I was sort of trying to support both APIs by separating the stages for opengl renderer too. However, this would require storing the `imgui::Ui` object with a limited lifetime which I can't guarantee. In other words, I'd have to own the data in my structure. With current imgui-rs API it is impossible to obtain the imgui draw state (vertex buffers and others) and own them. We only can get a reference to the current frame's data and guarantee its lifetime. 

Another problem is that now it is impossible to, let's say,  "draw" a bunch of frames of the imgui, then store these states, and later render those either to a frame buffer of an API of choice or dump to a file, or render without any API to a file, you would be able to do anything with it. The only option now is to render immediately at the same place where the `imgui::Ui` object is obtained.

**Describe the solution you'd like**
I'd like to have a way to obtain an owned copy of `imgui::DrawData`, namely vertex and index buffers, so that I am not limited by the lifetime of a single frame but can postpone the rendering and split it into stages and actually can do whatever I want with this data at any time later.

**Describe alternatives you've considered**
Can't think of any alternatives.

**Additional context**
[`imgui::DrawData`](https://docs.rs/imgui/latest/imgui/struct.DrawData.html) implements neither `Copy` nor `Clone`.

dbr:
Sorry for the slow reply!

This would be a good thing to support. Main issue is we can't trivially add `#[derive(Clone)]` to the DrawData as its private members contain pointers to `imgui::sys::ImDrawList` (and to `imgui::sys::ImGuiViewport` in the docking branch)

There was a similar request to (de)serialize the draw data in #606 which would work in this case - the solution there was to basically have a duplicate set of structs:
https://github.com/rmccrystal/imgui-rs-serialize/blob/a3e2288ee48a718153fd327041b7780fa5a32dae/src/draw_data.rs

We could provide these in `imgui-rs` itself (maybe as `DrawDataOwned`), but that has some downsides (like do renderers need to take both `imgui::DrawData` and `imgui::DrawDataOwned`?)

[...]

PR #721, opened 2023-03-29
Issue #687: Add OwnedDrawData for deep cloning DrawData

jbrd:
Introduce an `OwnedDrawData` struct that holds a cheaply-cloned `ImDrawData` structure that has been allocated and cloned on the heap. Designed for cases where you need to keep hold of the draw data so it can be evaluated at a later point in time (for instance, in multithreaded renderers where rendering the previous frame is overlapped with updating the next frame). It leverages `ImDrawList::CloneOutput` for cloning the underlying draw lists, and all allocations are made via ImGui, which has a nice consequence of them being tracked in the IO object's allocation metrics. The `OwnedDrawData` struct provides an accessor to the underlying `DrawData` struct, so that it can still be passed to renderers without modifying them.

As discussed in Issue #687

### DIFF
--- a/imgui/src/render/draw_data.rs
+++ b/imgui/src/render/draw_data.rs
@@ -1,3 +1,4 @@
+use std::mem::size_of;
 use std::slice;
 
 use crate::internal::{RawCast, RawWrapper};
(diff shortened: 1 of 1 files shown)


dbr:
Looks great, thanks!

jbrd:
Excellent, thank you!

PR #738, opened 2023-07-23
Add functions `begin_popup_context_*`.

rodrigorc:
Hi!

In ImGui there are these handy functions to do pop-up menus, usually by right-clicking the mouse. This PR adds basic wrappers around these.

The code is quite self-explanatory. The most notable thing missing is the ability to specify a flag other than `ImGuiPopupFlags_MouseButtonRight`. For me, I only open pop-ups when doing a `MouseButtonRight`, so this is good enough. But if you think it is wort it, we could ad an enum wrapper and a way to specify it, maybe a builder helper struct o a few `begin_popup_context_*_with_flags()` functions.

### DIFF
--- a/imgui/src/popups.rs
+++ b/imgui/src/popups.rs
@@ -252,7 +252,7 @@ impl Ui {
     ///
     /// This does not take a label, which means that multiple calls **in a row** will use the same label, which
     /// is based on the last node which had a label. Text and other non-interactive elements generally don't have
-    /// ids, so you'll need to use [begin_popup_context_with_label] for them.
+    /// ids, so you'll need to use [begin_popup_context_with_label](Self::begin_popup_context_with_label) for them.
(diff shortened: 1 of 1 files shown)


sanbox-irl:
Hey, thanks for this PR! This looks great -- I'll take a deeper look soon and likely merge!

sanbox-irl (changes requested):
Took a look -- I think yup, your instinct is correct -- we'll need a version that can take an optional mouse button.

I'll push a version and let me know how it looks?

sanbox-irl:
okay @rodrigorc I added a few extra methods to handle setting a label or not. I agree that handling the non-RMB case should be deferred till we have a user who wants that, since it sounds...awfully niche and annoying.

Let me know what you think of these additions and we can merge it!

rodrigorc:
I see the changes all fine, thanks.

PR #745, opened 2023-09-10
gamepad support for winit using gilrs

Calcoph:
Have to use WinitPlatform::handle_controller_event() instead of WinitPlatform::handle_event()

Adds the dependency [gilrs](https://gitlab.com/gilrs-project/gilrs) to `imgui-winit-support`

### DIFF
--- a/imgui-winit-support/Cargo.toml
+++ b/imgui-winit-support/Cargo.toml
@@ -12,3 +12,4 @@ categories = ["gui"]
 [dependencies]
 imgui = { version = "0.11.0", path = "../imgui" }
 winit = { version = "0.27.2", default-features = false }
+gilrs = { version = "0.10.2" }
--- a/imgui-winit-support/src/lib.rs
+++ b/imgui-winit-support/src/lib.rs
@@ -73,8 +73,10 @@
 //! })
 //! ```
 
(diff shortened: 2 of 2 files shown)


sanbox-irl:
hi @Calcoph thanks for the hard work!

However, i feel uncertain about this PR and think it might be best to close it for now. That's because this actually can be entirely handled by users directly rather than baked into this crate -- IMO that's a big win. In fact, i really see your work as having two distinct halves -- an `imgui-gilrs` crate, which handles imgui + gilrs type glue, and your hook into winit. I'd love to support such an `imgui-gilrs` crate with a shoutout on the README, but this repository has made many errors by *overly* directly supporting many third party dependencies. This hasn't been a great idea, so I'm loathe to take on another dependency whose updates we'd need to follow.

Calcoph:
I like your idea of a separate `imgui-gilrs` crate, since it doesn't really need any of the private `imgui-winit-support` code. In order to not put extra pressure onto imgui mantainers, I've decided to make that crate separately. It's [here](https://github.com/Calcoph/imgui-gilrs) if anyone is interested.

sanbox-irl:
Thank you! I'll add a line in the README for this project

PR #746, opened 2023-09-12
unsafety in text fix

sanbox-irl:
Hiya! Long time no see.

There's some unsafety in our UiBuffer in particular situations -- an additionally makes clippy stop yelling at us.

The buffer is fairly simple -- it's just a giant list of bytes which we write user provided strings and then append a null byte to afterwards. We hand Dear Imgui a pointer to a position within that buffer (the start of the string the user gave), which it reads naively until it hits the null byte we gave.

This works fine, but since we use a Vec, multiple pushes can cause the backing buffer to get reallocated. We in a few places, notably the Tree structures, held onto invalid pointers past pushes.

This PR fixes those unsafety issues. There is perhaps a better way to architect this to use the borrow checker more easily to our advantage, but I haven't done that exploration work here.

### DIFF
--- a/imgui-examples/examples/custom_textures.rs
+++ b/imgui-examples/examples/custom_textures.rs
@@ -189,8 +189,8 @@ impl Lenna {
         decoder.read_image(&mut image)?;
         let raw = RawImage2d {
             data: Cow::Owned(image),
-            width: width as u32,
-            height: height as u32,
+            width,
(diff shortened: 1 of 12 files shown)


sanbox-irl:
Goign to go ahead and merge this -- we should also update MSRV in the future

PR #754, opened 2023-11-18
Updated winit, glutin and glium dependencies

Rob2309:
This PR updates winit to 0.29, glutin to 0.31 and glium to 0.33.
Currently, the crates.io release of glium is not compatible with the other dependencies. Until this is fixed, this PR uses a patch to use the git version of glium instead, which is compatible already. Thus, until crates.io is updates, this PR probably should not be merged.

The new key handling of winit seems very unintuitive to me, so someone checking the updated key handling code would be appreciated.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,6 @@ members = [
 
 package.rust-version = "1.70"
 resolver = "2"
+
+[patch.crates-io]
+glium = { git="https://github.com/glium/glium" }
--- a/imgui-examples/Cargo.toml
+++ b/imgui-examples/Cargo.toml
@@ -10,7 +10,7 @@ publish = false
 
 [dev-dependencies]
 copypasta = "0.8"
-glium = { version = "0.32.1", default-features = true }
+glium = { version = "0.33.0", default-features = true }
 image = "0.23"
 imgui = { path = "../imgui", features = ["tables-api"] }
 imgui-glium-renderer = { path = "../imgui-glium-renderer" }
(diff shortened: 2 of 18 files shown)


sanbox-irl:
Thanks for this `Thus, until crates.io is updates, this PR probably should not be merged.` unfortunately, yes, that's why we haven't updated. I am not very pleased with that!

sanbox-irl:
Actually looks like `glium` finally published `0.33` a few weeks ago!

sanbox-irl:
Actually, that release didn't update their deps enough for us (I believe -- it's awfully difficult to track these). I suspect it might still be worth merging though, since at least then we could point users to using as a git dep instead of from crates.io

Sewer56:
It's not super urgent in any manner, but I should note in any case that I'm unable to run any of the examples under Wayland without this patch. Neither `glow` or `glium`. After this patch, both work without any issue whatsoever.

damyanp:
Just a note that it seems that this commit breaks compilation on Windows - several of the examples fail to build.  eg:

```
error[E0308]: mismatched types
   --> imgui-examples\examples\support\mod.rs:36:29
    |
36  |         .set_window_builder(builder)
    |          ------------------ ^^^^^^^ expected `winit::window::WindowBuilder`, found `WindowBuilder`
    |          |
    |          arguments to this method are incorrect
    |
    = note: `WindowBuilder` and `winit::window::WindowBuilder` have similar names, but are actually distinct types
note: `WindowBuilder` is defined in crate `winit`
```

I've not had a chance to dig much deeper I'm afraid, but I thought I'd mention it as I didn't see any issues about this already.

PR #755, opened 2023-11-23
Ability to fetch current frame.

Nhlest:
Maybe i am missing something but i could not find a workaround for this without a method that looks something like this. 
Example use-case: While using bevy engine's ECS model and storing the imgui Context as a non-sendable resource in ECS's world i am completely unable to construct imgui's UI from different systems since only one scope can own mutable reference to the current frame's Ui. This method allows me to store imgui's context as a resource and retreive current frame on demand.

Also i found set_current_context useful for when i need to call imgui functions from a shared library (something to do with shared libraries having its own static memory for imgui context pointer). [Explanation here](https://github.com/ocornut/imgui/blob/master/imgui.cpp#L1130)

### DIFF
--- a/imgui/src/context.rs
+++ b/imgui/src/context.rs
@@ -114,6 +114,12 @@ impl Context {
         clear_current_context();
         SuspendedContext(self)
     }
+    /// Sets current global imgui context, i.e. in a dynamically linked library.
+    pub fn set_current_context(&self) {
+        unsafe {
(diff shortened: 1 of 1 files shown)


sanbox-irl:
Apologies for the wait. I believe this isn't necessary -- see `Context::activate`. That should be able to get you the Context when you need it in a globally available way

PR #756, opened 2023-12-09
fix: use correct alpha blending function

### LINKED ISSUE #733: imgui-glium-renderer has wrong alpha blending settings
The winit `Blend::alpha_blending()` used in the glium renderer backend produces an incorrect blending function for our purposes. 

Winit provides (and imgui-glium-renderer uses):
```rust
alpha: BlendingFunction::Addition {
    source: LinearBlendingFactor::SourceAlpha,
    destination: LinearBlendingFactor::OneMinusSourceAlpha
},
```
where the correct function for "overlay" alpha blending should be 
```rust
alpha: BlendingFunction::Addition {
    source: LinearBlendingFactor::One,
    destination: LinearBlendingFactor::OneMinusSourceAlpha,
},
```

**Screenshots**:

Without the change - even though the window background is fully opaque, the drawn imgui windows overwrite the alpha with an incorrect value.
image

With the change, the window opacity is preserved as expected.
image

If this is not intentional (which i presume it isn't) *and is not deemed to be a mistake in winit upstream* then I will submit a PR later when I get the chance.

Syndelis:
Fixes #733

## About this PR

This PR changes the default alpha blending function just as described by @Lateasusual in #733.

I'm unsure of why this hasn't been done yet, but there appears to be a consensus on the mentioned issue that this is the way to go.

### DIFF
--- a/imgui-glium-renderer/src/lib.rs
+++ b/imgui-glium-renderer/src/lib.rs
@@ -9,8 +9,8 @@ use glium::uniforms::{
     MagnifySamplerFilter, MinifySamplerFilter, Sampler, SamplerBehavior, SamplerWrapFunction,
 };
 use glium::{
-    program, uniform, vertex, Blend, DrawError, DrawParameters, IndexBuffer, Program, Rect,
-    Surface, Texture2d, VertexBuffer,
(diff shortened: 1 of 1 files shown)


Diegovsky:
+1

Diegovsky:
@sanbox-irl is this able to be merged?

PR #763, opened 2024-01-31
Remove noisy key event debug print

mkrasnitski:
When #754 was merged it left a debug print in `imgui-winit-support` by accident, which produces a lot of noise. My guess is it was used to test that key events are received when the window focus changes. Either way it should be removed.


dbr:
Thanks for the PR - the stray print was removed in #774 along with some dependency updates!

PR #764, opened 2024-02-10
Update glium dependency

asd988:
Examples didn't compile, looks like it was just enough updating glium.


matt-kimball:
This should be merged.  I needed to make the same change locally to `imgui-glium-renderer` to successfully use it.

dbr:
Thanks for the PR - this and some other deps were updated as part of #774 !

PR #770, opened 2024-03-28
Remove debug message

sarchar:
This debug message is printed on every key event, flooding the terminal.


dbr:
Thanks for the PR - the stray print was removed in #774 along with some dependency updates!

PR #772, opened 2024-04-19
Remove stray println

colinmarc:
Looks like this snuck in in #754.

### DIFF
--- a/imgui-winit-support/src/lib.rs
+++ b/imgui-winit-support/src/lib.rs
@@ -483,8 +483,6 @@ impl WinitPlatform {
                 // https://github.com/ocornut/imgui/issues/5047
                 handle_key_modifier(io, &key, pressed);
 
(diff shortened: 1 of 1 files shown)


colinmarc:
By the way, it would be great to get a new release targeting `winit = "0.29"`. I don't think there's any need to pin the minor revision (`0.29.2` currently).

dbr:
Thanks for the PR - the stray print was removed in #774, along with variuos dependencies updates

> I don't think there's any need to pin the minor revision (0.29.2 currently)

I think the Cargo.toml version syntax is a bit confusing in this case - `winit = "0.29.2"` specifies `0.29.2` as a minimum version, but `0.29.3` etc can still be picked up. To actually pin it to `.2` it would need to say `winit = "=0.29.2"`

That aside, a new release is definitely overdue! Shall try and get a new release of imgui-rs out soon (hopefully today!)

PR #774, opened 2024-04-24
Finally fixed dependency hell

Rob2309:
Since by now it seems like every required imgui dependency plays nicely with the winit update, this PR should finally fix all compile errors in this project.


dbr:
Good news - thanks for the PR!

Issue #777, opened 2024-05-02
Crashes on 1.78 with imgui-opengl-renderer


jestarray:
```
unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_nounwind_fmt::runtime
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/panicking.rs:110:18
   2: core::panicking::panic_nounwind_fmt
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/panicking.rs:123:9
   3: core::panicking::panic_nounwind
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/panicking.rs:156:5
   4: core::slice::raw::from_raw_parts::precondition_check
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/intrinsics.rs:2799:21
   5: core::slice::raw::from_raw_parts
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/slice/raw.rs:98:9
   6: imgui::render::draw_data::DrawData::cmd_lists
             at /home/a/.cargo/registry/src/index.crates.io-6f17d22bba15001f/imgui-0.11.0/src/render/draw_data.rs:60:9
   7: imgui::render::draw_data::DrawData::draw_lists
             at /home/a/.cargo/registry/src/index.crates.io-6f17d22bba15001f/imgui-0.11.0/src/render/draw_data.rs:48:23
   8: imgui_opengl_renderer::Renderer::render
             at ./rust-imgui-opengl-renderer/src/lib.rs:211:24
```             
This is where it crashes:
https://github.com/imgui-rs/imgui-rs/blob/cd97e171fe046c063a6e731d6819a887e48d985d/imgui/src/render/draw_data.rs#L59

I am using this library to hook into imgui which calls `draw_lists()` here: 
https://github.com/michaelfairley/rust-imgui-opengl-renderer/blob/master/src/lib.rs#L211

- rust 1.78
 - imgui-rs 0.11.0
Arch Linux

How to reproduce:
[...]

PR #779, opened 2024-05-14
fix: check memory boundaries before using slice::from_raw_parts

juliusl:
In rustc 1.78  `ub_checks::assert_unsafe_precondition` was added to `slice::from_raw_parts`
```rs
        ub_checks::assert_unsafe_precondition!(
            check_language_ub,
            "slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`",
            (
                data: *mut () = data as *mut (),
                size: usize = size_of::<T>(),
                align: usize = align_of::<T>(),
                len: usize = len,
            ) =>
            ub_checks::is_aligned_and_not_null(data, align)
                && ub_checks::is_valid_allocation_size(size, len)
        );
```

This causes the rendering code in `draw_data.rs` to panic during runtime. This fix adds checks before creating the slice and if they fail returns an empty slice. 

Validated on WSL2/Ubuntu 22.04

Addresses #777

### DIFF
--- a/imgui/src/render/draw_data.rs
+++ b/imgui/src/render/draw_data.rs
@@ -60,8 +60,8 @@ impl DrawData {
             return &[];
         }
         slice::from_raw_parts(
-            self.cmd_lists as *const *const DrawList, 
-            self.cmd_lists_count as usize
+            self.cmd_lists as *const *const DrawList,
+            self.cmd_lists_count as usize,
         )
     }
     /// Converts all buffers from indexed to non-indexed, in case you cannot render indexed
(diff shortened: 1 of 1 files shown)


sanbox-irl:
Thank you for the PR @juliusl !

jestarray:
@sanbox-irl is it possible to issue `0.12.1` with this fix in? Because anyone who starts a new project and just adds
````
imgui = { version = "0.12" }
```
 will just crash instantly and there's no way around it other than setting it to use a git commit
 I've been using this rev for about a year or so now
 ```
 imgui = { git = "https://github.com/imgui-rs/imgui-rs.git", rev = "8140b3f", features = [
    "tables-api",
] }
 ```

PR #783, opened 2024-06-06
Glow texture example vec capacity fix

SethPyle376:
Small thing I noticed going through the example code. This vec should be initialized with capacity for 3 bytes per pixel instead of 1.


sanbox-irl:
good find!

PR #786, opened 2024-07-07
block __darwin_off_t, all of stdio from bindgen

ArcaneNibble:
The imgui-sys crate was previously leaking platform-specific information into the generated code. This replaces it with an enum with no variants which can only be handled via indirection (i.e. it isn't possible to create an instance of a FILE)


sanbox-irl:
hi @ArcaneNibble can you rebase this onto the current `main` branch?

sanbox-irl:
closing as stale -- let me know if you can return to it!

PR #793, opened 2024-09-27
split the repository up

sanbox-irl:
- **it's going, but it doesn't actually work**
- **simplified repository down**
- **removed resources folder, as we don't use it anymore**
- **moved markdown to md, because it's bothered me for years**


sanbox-irl:
Remaining work to do:

1. Add README's and add some CI to test the platform and renderer crates
2. Update this repository's README (and CI if necessary)
3. Start actually fixing imgui's problem, which have been somewhat in arrested development as these changes have migrated.

This PR is a large change to the repository, in two ways, one concrete and one more abstract:

1. The concrete way this works is that it will be significantly easier for us to update platform and renderer bugs. These are *overwhelmingly* the source of the bugs that we get, which makes sense -- those crates actually are more complicated than imgui.

2. Morally, the main imgui crate is this organizations primary responsibility. Making sure that we have a good starting helper crate system is important, but not as important as keeping imgui up to date.

PR #805, opened 2024-10-12
add Copy and Clone traits for TableColumnSetup and TableColumnFlags

Sk44rt:
creating TableColumnSetup struct in every frame does not looks like a good idea as for performance
traits Copy and Clone should at least allow to initialize TableColumnSetup once and reuse it every frame


sanbox-irl:
These are bitflags, so making them every frame will be pretty cheap. That being said, no reason to not implement Copy/Clone IMO!

Thank you for the PR

PR #806, opened 2024-10-12
Return the dockspace ID from `dockspace_over_main_viewport`

sagacity:
When creating a dockspace over a main viewport, you need the resulting id in order to use the `DockManager*` APIs (for instance to programmatically layout a bunch of windows).


sanbox-irl:
Thank you for the PR!

Issue #808, opened 2024-10-16
Cannot compile due to conflicting implementations of Clone & Copy for TableColumnFlags


LB767:
**Describe the bug**
I tried to use the git repo directly in my project (because of some fixes and features 0.12 doesn't have) but apparently cannot get it to compile.

**To Reproduce**
I'm using the following in my cargo.toml
```
imgui = { git = "https://github.com/imgui-rs/imgui-rs.git", features = ["docking", "tables-api"] }
imgui-glow-renderer = { git = "https://github.com/imgui-rs/imgui-glow-renderer.git" }
imgui-winit-support = { git = "https://github.com/imgui-rs/imgui-winit-support.git" }
```
And I try to use some table-api functionality.

**Expected behavior**
It should compile

**Screenshots**
The compiler error for the Copy trait:
```
error[E0119]: conflicting implementations of trait `std::marker::Copy` for type `TableColumnFlags`
   --> C:\Users\lbf\.cargo\git\checkouts\imgui-rs-61ccdc9370f94e57\91fdf45\imgui\src\tables.rs:161:1
    |
161 | / bitflags! {
162 | |     /// Flags for [TableColumnSetup] and [table_setup_column_with].
163 | |     #[repr(transparent)]
164 | |     #[derive(Copy, Clone, Default)]
    | |              ---- first implementation here
...   |
219 | |     }
220 | | }
    | |_^ conflicting implementation for `TableColumnFlags`
    |
    = note: this error originates in the derive macro `Copy` which comes from the expansion of the macro `bitflags` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0204]: the trait `std::marker::Copy` cannot be implemented for this type
   --> C:\Users\lbf\.cargo\git\checkouts\imgui-rs-61ccdc9370f94e57\91fdf45\imgui\src\tables.rs:752:10
    |
752 | #[derive(Copy, Clone, Debug, Default)]
    |          ^^^^
...
757 |     pub flags: TableColumnFlags,
    |     --------------------------- this field does not implement `std::marker::Copy`
    |
    = note: this error originates in the derive macro `Copy` (in Nightly builds, run with -Z macro-backtrace for more info)
```

**Please describe your environment**
 - imgui-rs latest git version as of 16.10.24
 - Windows 11

LB767:
Ah thank you, I guess it was just very unlucky timing from my part :)

Sk44rt:
[...]

PR #809, opened 2024-10-16
remove Copy and Clone traits from TableColumnFlags

Sk44rt:
my bad, i didn't checked for errors before making PR #805 
and I only realized it after I saw issue #808 
so let's merge it before new release
I'm embarrassed about it (╥﹏╥)


sanbox-irl:
How did it pass the CI *before* it got merged? How weird, will need to investigate

Sk44rt:
> How did it pass the CI _before_ it got merged? How weird, will need to investigate

i think CI does not use `--all-features` flag for `cargo check`

Issue #810, opened 2024-10-20
Is there a reason why there are no bindings to `AddFocusEvent` ?


jestarray:
https://github.com/imgui-rs/imgui-rs/blob/8140b3f0a8662927432a139e19767618f0e328f2/imgui-sys/src/bindings.rs#L3834

sanbox-irl:
Not particularly -- I'd take a PR!

PR #811, opened 2024-10-20
add support for focus event

jestarray:
closes https://github.com/imgui-rs/imgui-rs/issues/810

### DIFF
--- a/imgui/src/io.rs
+++ b/imgui/src/io.rs
@@ -457,9 +457,10 @@ impl Io {
             sys::ImGuiIO_AddKeyEvent(self.raw_mut(), key as u32, down);
         }
     }
+
     /// Queue a gain/loss of focus for the application (generally based on OS/platform focus of your window)
     /// Note: [`io.config_debug_ignore_focus_loss`] will ignore this event from firing
-    #[doc(alias = "AddFocusEvent")
+    #[doc(alias = "AddFocusEvent")]
     pub fn add_focus_event(&mut self, focused: bool) {
         unsafe {
             sys::ImGuiIO_AddFocusEvent(self.raw_mut(), focused);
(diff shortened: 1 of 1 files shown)


imgui/src/io.rs
```diff
@@ -458,6 +458,12 @@ impl Io {
         }
     }
 
+    pub fn add_focus_event(&mut self, focused: bool) {
```
sanbox-irl, line 461: Can you add documentation above this?

I would start by ripping out the documentation from `ImGuiIo_AddFocusEvent` itself in the C++ (I'll post that below):

```
// Queue a gain/loss of focus for the application (generally based on OS/platform focus of your window)
```

I'd also add a note re: `ConfigDebugIgnoreFocusLoss`, which basically makes imgui ignore this event firing.

Lastly, I'd add a doc alias `#[doc(alias = "AddFocusEvent")]` for the C++ api to bind to us directly.

sanbox-irl (approved):
Sorry this took a bit! Mostly this is just a request for more docs. I understand why you didn't add some (the other methods don't have any) but i'd like to improve the quality of the library overtime

imgui/src/io.rs
```diff
@@ -457,7 +457,9 @@ impl Io {
             sys::ImGuiIO_AddKeyEvent(self.raw_mut(), key as u32, down);
         }
     }
-
+    // Queue a gain/loss of focus for the application (generally based on OS/platform focus of your window)
+    // Note: [`io.config_debug_ignore_focus_loss`] will ignore this event from firing
```
sanbox-irl, line 461: Needs to be `///` instead of just `//`

sanbox-irl (changes requested):
last one!

PR #818, opened 2024-12-27
delete use of the `chlorine` crate

Lokathor:
Hello!

It appears that you're currently still using the `chlorine` crate. While I do love the dedication, since those definitions were added to `core::ffi` in 1.64, and since your readme has an msrv listed of 1.70, I've gone ahead and deleted your use of `chlorine` and moved you over to just using `core::ffi`. I hope everything has been set right, it passed a local `cargo test`.


sanbox-irl:
thank you lokathor!!

Issue #820, opened 2025-01-10
Exiting the app when supporting multiple viewports fails a IM_ASSERT_USER_ERROR


paulo-santana:
**Describe the bug**
The API for setting up the multiple viewports support stores the PlatformViewportBackend object inside the context.IO.BackendPlatformUserData, which causes the app to fail when the context is dropped, because Dear Imgui checks if i's null before. 

Even though it's just a leak check before the app exiting, it still would be nice to have the app to exit gracefully. Besides, the user data was intended for the user of the lib to use, so I think we should avoid using it for the app internals and serve it in the public API instead of locking it inside the crate.

**To Reproduce**
Run [imgui-winit-glow-renderer-viewports](https://github.com/imgui-rs/imgui-winit-glow-renderer-viewports)'s example with
```sh
cargo run --example viewports_basic
```
then close the app.

**Expected behavior**
A clean exit without errors

**Screenshots**
![image](https://github.com/user-attachments/assets/9f34c4b2-c81b-43b4-be15-6a39c8f4caff)
![image](https://github.com/user-attachments/assets/628d3e96-39a2-4fa4-aacf-7c1ce7ac7cae)
![image](https://github.com/user-attachments/assets/2e8ab350-d9be-4ff7-837e-e2429a27361b)

**Please describe your environment**
 - imgui-rs version 0.12.0 with docking feature
 - arch btw

**Other information**
I know the docking and viewport are experimental both in Dear Imgui and in imgui-rs, but I wanted to at least have the problem tracked, so that's why I'm opening the issue. If i find a good way to fix it I'll gladly open a PR. I'm initially thinking of just saving the backend impl somewhere else

Also maybe consider another approach of providing an API for users to manage viewports. I'm not that well versed in rust yet, but i've been thinking it could take closures defined in the platform, so it would have access to all of it's objects directly and wouldn't need to queue commands. But I don't know if the compiler agrees with me

paulo-santana:
@sanbox-irl I tried using a global inside the crate to store the object, you can view the diff here https://github.com/paulo-santana/imgui-rs/pull/2. If you and the other maintainers agree with the changes I'll open the pull request.

sanbox-irl:
@paulo-santana definitely the right direction -- can you open a PR? I suspect it should actually be a thread-local (i'm not confident the Send/Sync impls are actually valid), and perhaps we can even get away with putting it somewhere else. That being said, if you open a PR, I'll give more complete thoughts

PR #821, opened 2025-01-12
move platform and renderer backends out of Dear Imgui

paulo-santana:
Store the objects containing the callback methods for multi viewport
support in two crate globals that are managed by imgui-rs, instead of using
BackendPlatformUserdata and BackendRendererUserData

fixes https://github.com/imgui-rs/imgui-rs/issues/820

### DIFF
--- a/imgui/src/docking_utils.rs
+++ b/imgui/src/docking_utils.rs
@@ -1,9 +1,17 @@
 use std::{
+    cell::UnsafeCell,
     ffi::{c_void, CStr},
     os::raw::{c_char, c_int},
+    sync::{Mutex, MutexGuard},
 };
 
-use crate::{Io, PlatformIo, Viewport};
+use crate::{PlatformIo, Viewport};
+
+pub(crate) static PLATFORM_VIEWPORT_BACKEND: Mutex<Option<crate::PlatformViewportContext>> =
+    Mutex::new(None);
+
+pub(crate) static RENDERER_VIEWPORT_BACKEND: Mutex<Option<crate::RendererViewportContext>> =
+    Mutex::new(None);
 
 /// Trait holding functions needed when the platform integration supports viewports.
 ///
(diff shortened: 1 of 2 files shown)


imgui/src/docking_utils.rs
```diff
@@ -1,9 +1,17 @@
...
+    cell::UnsafeCell,
     ffi::{c_void, CStr},
     os::raw::{c_char, c_int},
+    sync::{Mutex, MutexGuard},
 };
 
-use crate::{Io, PlatformIo, Viewport};
+use crate::{PlatformIo, Viewport};
+
+pub(crate) static PLATFORM_VIEWPORT_BACKEND: Mutex<Option<crate::PlatformViewportContext>> =
```
sbenedetto, line 10: What is the reason for using a Mutex? Is this meant to be shared among threads?

sanbox-irl: Yeah, we probably don't need a mutex -- imgui is generally single-threaded. I suspect we can move it into a thread local. That also lets us remove the unsound send/sync imps

paulo-santana: > What is the reason for using a Mutex?

No reason at all except that's what I had to do to make the code compile. I'll look into making it thread local as sanbox suggested

paulo-santana: Here's a first tackle at it https://github.com/paulo-santana/imgui-rs/pull/3

took me hours to give up trying to return a reference to the backends directly. The lifetimes wouldn't allow it

paulo-santana: Probably wouldn't need the type to be Optional if we come back with the dummy backends, i'll try that next

paulo-santana: I'll merge it and update the PR, if you have any suggestions on how to handle it differently you can just throw it

Also I added back the dummy backends so there's no need to use an Option, which makes the code a little bit cleaner. If you do prefer the Option, i think it should be done in another PR. It's not directly related to the issue, i think it's more of a design choice.

sanbox-irl (approved):
lgtm! thank you for your hard work!

sanbox-irl:
Note: CI failed, but this is because we'll need to bump MSRV which we've needed to do for awhile. I am going to merge this and then immediately bump MSRV on main.

PR #838, opened 2025-10-19
feat: add `external` feature to disable internal C++ build

psyche314:
when enabled, this feature skips compiling `libcimgui.a` within imgui-sys, allowing the library to be provided externally instead.

this commit addresses the issue where ImGui has already been built externally, but the project still compiles a new instance of it.


sanbox-irl:
Thanks for the work!

PR #840, opened 2025-12-22
fix: compile imgui-sys using same defines as cimgui

aliu:
cimgui generates its bindings with `IMGUI_DISABLE_OBSOLETE_FUNCTIONS=1` set, however imgui-sys is not building with the same compile definition. This results in some mismatches between the generated rust bindings and the linked library.

See https://github.com/cimgui/cimgui/blob/d94ad1b16224a01592ec60f21b284f0883ea9a56/generator/generator.lua#L169-L171


aliu:
FYI this change now enables this assertion: https://github.com/ocornut/imgui/issues/5548. I believe it is the only observable difference in behavior resulting from setting `IMGUI_DISABLE_OBSOLETE_FUNCTIONS` when compiling the current vendored version of ImGui.

PR #841, opened 2025-12-30
fix: fix inconsistent assertion

YichiZhang0613:
There are some inconsistent assertions.
For this assertion, the predicate requires channel_index < self.channels_count while the message indicates channel_index cannot be higher than self.channels_count, which means "<=". So the message should be changed into "Provided channel index ({}) is greater than or equal to channel count".
```rust
        assert!(
            channel_index < self.channels_count,
            "Channel cannot be set! Provided channel index ({}) is higher than channel count ({}).",
            channel_index,
            self.channels_count
        );
```
For this assertion, the predicate requires width >= 0 while the message requires width > 0. As it makes no sense when it equals 0, so the predicate should be changed into `width > 0`. The same issue elsewhere should be fixed as well.
```rust
assert!(width >= 0, "font texture width must be positive");
```


YichiZhang0613:
@sanbox-irl Hi, sorry for the interruption. I see you've been active reviewer recently. Could you please review this PR when you have time? It's been open for four weeks and I'd appreciate any feedback. Thank you very much.

sanbox-irl:
@YichiZhang0613 Thanks for the reminder -- it looks right to me. I looked into imgui's source and it looks like this is what we want. Arguably, we should just take a NonZeroU32, but sometimes encoding too much into the type system is a huge pain

PR #843, opened 2026-06-20
Fix a typo in FontConfig::apply_to_raw_config which ignores rasterizer_density

OlegOAndreev:
A typo prevents RasterizerDensity from being set, which leaves the fonts blurry on high-dpi displays.


sanbox-irl:
Thank you!

